PR #3, opened 2019-05-09
Format code according to default style

djc:
I don't know how attached you are to your particular style, but, FWIW, I found the style sufficiently different from what I am used to that it was a bit distracting (especially your alignment of `:` characters).


est31:
The `:` are placed with space before no space after whenever they are followed by a type or type constraint, and placed with space before and after whenever they are used for assignments in struct initializers. The usage is fundamentally different and I'm not really happy about both how the language chose the same syntax for both nor about how the official format guidelines chose the same format for both. I format the `:` like I format a `=` when it's used like one. Also I prefer using tabs :).

Generally I'd prefer to keep it in my personal style that I use for other projects as well.

PR #4, opened 2019-05-09
Add API to allow signing certificates with another certificate's key

djc:


est31:
The first commit is alright you can put it into a separate PR if you want it merged right away.

As for the second commit: I'm a bit surprised webpki accepts the certificates as CA certificates because actually the standard requires the CA flag to be enabled: https://tools.ietf.org/html/rfc5280#section-4.2.1.9

>  Conforming CAs MUST include this extension in all CA certificates
   that contain public keys used to validate digital signatures on
   certificates

As well as subject key identifier to be present: https://tools.ietf.org/html/rfc5280#section-4.2.1.2

> To facilitate certification path construction, this extension MUST
   appear in all conforming CA certificates, that is, all certificates
   including the basic constraints extension (Section 4.2.1.9) where the
   value of cA is TRUE.

Could you add a flag like `can_sign` or something to the CertificateParams and output those values if the flag is true?

djc:
@briansmith would it make sense to add checks to webpki for this stuff? I was thinking building a trust anchor could fail if these are not satisfied.

briansmith:
> @briansmith would it make sense to add checks to webpki for this stuff? I was thinking building a trust anchor could fail if these are not satisfied.

Is the certificate v1 or v3?

djc:
It's v3.

briansmith:
Let's discuss it in a webpki issue; please file one if you are interested. There are pros/cons to making such a change.

djc:
Opened https://github.com/briansmith/webpki/issues/99.

djc:
I've worked on an initial attempt to write the `basicConstraints` extension, but it currently generates a `BadDER` error from `webpki`. Any suggestions for what I'm doing wrong? (I'd be happy to dive in myself, but you probably know much more about all the DER stuff.)

Also not sure yet how to generate the right DER for `subjectKeyIdentifier` and `authorityKeyIdentifier`, do you have some hints about how to do this? As the identifier I can probably just use a hash of the public key as suggested in the RFC, though I'd be inclined to use SHA256 instead of SHA1 (and it seems like as an identifier this shouldn't cause any issues). I'm not sure there are special concerns about dealing with the public key "as a BIT STRING". Also not sure about what it means for the encoding to have three OPTIONAL fields or how to encode that (I suppose `OCTET STRING` would require `write_bytes()` and then maybe `write_null()` for the optional fields I don't want?).

est31:
For the BadDER errors they were quite a pain during development of the library because they don't show the actual reason so you can't dive into the code base to find out. I think I did something like `sed -i -e 's/Err(Error::BadDER)/panic!("Bad DER")/' src/*.rs` and then checked the location info of the panic. Running this on the webpki codebase gives you a compile error but should get you started.

For webpki, I don't really mind it accepting keys *without* the extension. I guess it makes life easier for people to add self-generated certificates to the store. I just wanted to make sure that we generate more or less compliant keys or at least give people the option to.

> I'd be inclined to use SHA256 instead of SHA1 (and it seems like as an identifier this shouldn't cause any issues).

Yeah that RFC is outdated. SHA-256 is totally fine as would be just setting it to a constant.

> I'm not sure there are special concerns about dealing with the public key "as a BIT STRING".

I've learned (the hard way :p) that libraries like webpki generally do check for things like the type. The subjectPublicKey is serialized as BIT STRING too, you can draw inspiration from [here](https://github.com/est31/rcgen/blob/d6182512753db5419ebbc37d122fe8e7ab6910f2/src/lib.rs#L273).

> Also not sure about what it means for the  encoding to have three OPTIONAL fields or how to encode that (I suppose OCTET STRING would require write_bytes() and then maybe write_null() for the optional fields I don't want?).

As there is a tag  (the number inside the `[]`) I think you don't need to worry about any `OPTIONAL` fields you left out. Not sure what to do when there is no tag never encountered it. You can write the tag using the `write_tagged` function.

djc:
Okay, basicConstraints and subjectKeyIdentifier are now both included.

src/lib.rs
```diff
@@ -177,6 +171,46 @@ pub struct CertificateParams {
 	pub serial_number :Option<u64>,
 	pub subject_alt_names :Vec<String>,
 	pub distinguished_name :DistinguishedName,
+	pub is_ca :IsCa,
+}
+
+/// Whether the certificate is allowed to sign other certificates
+pub enum IsCa {
+    /// The certificate can only sign itself
```
est31: Could you use tabs here?

src/lib.rs
```diff
@@ -177,6 +171,46 @@ pub struct CertificateParams {
...
+    /// The certificate may be used to sign other certificates
+    Ca(BasicConstraints),
+}
+
+/// The path length constraint (only relevant for CA certificates)
+///
+/// Sets an optional upper limit on the length of the intermediate certificate chain
+/// length allowed for this CA certificate (not including the end entity certificate).
+pub enum BasicConstraints {
+    /// No constraint
```
est31: Tabs needed here, too.

src/lib.rs
```diff
@@ -290,17 +324,44 @@ impl Certificate {
...
+							let digest = digest::digest(&digest::SHA256, self.key_pair.public_key().as_ref());
+							writer.next().write_bytes(&digest.as_ref());
+						});
+						// Write basic_constraints
+						writer.next().write_sequence(|writer| {
+							let oid = ObjectIdentifier::from_slice(OID_BASIC_CONSTRAINTS);
+							writer.next().write_oid(&oid);
+							let bytes = yasna::construct_der(|writer| {
+								writer.write_sequence(|writer| {
+									writer.next().write_bool(true);
```
est31: Maybe add a terse comment about what this bool means? Like `// cA flag` or something.

src/lib.rs
djc: Done.

src/lib.rs
djc: Done.

src/lib.rs
djc: Done.

djc:
Addressed your nits, thanks for the review and sorry about getting the tabs wrong!

est31:
@djc you are still using spaces

djc:
🤦‍♂

est31:
Thanks for the PR!

djc:
BTW are you planning a new release soon? Would be nice to have this stuff on crates.io.

est31:
@djc definitely, I plan to do it this evening.

est31:
@djc done!

https://github.com/est31/rcgen/releases/tag/v0.3.0

https://github.com/est31/rcgen/blob/v0.3.0/CHANGELOG.md#release-030---may-18-2019

https://docs.rs/rcgen/0.3.0/rcgen/

djc:
Uh, why does the change log mention @Ralith instead of me?😜

est31:
Oh sorry dude. I fucked up LMAO. Will fix this in an instant.

est31:
commit 3a20c8bf49632a071cffcf75ab685cd43f353ee4 : sorry again :)

djc:
No problem, thanks for fixing it quickly

PR #5, opened 2019-06-02
Add support for writing certificate signing requests

djc:
Hey, would you be interested in merging something like this? Seems like a feature that mostly fits, at least, although it maybe widens the scope a bit.

On the other hand, I'm not sure how best to test this for now (although I suppose the openssl library exposes API to actually handle CSRs). So far this is based on just comparing openssl-generated CSRs with the JS ASN.1 thingy you point to in your README.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,9 @@ pub fn generate_simple_self_signed(subject_alt_names :impl Into<Vec<String>>) ->
 // Example certs usable as reference:
 // Uses ECDSA: https://crt.sh/?asn1=607203242
 
+/// pkcs-9-at-extensionRequest in RFC 2985
+const OID_PKCS_9_AT_EXTENSION_REQUEST :&[u64] = &[1, 2, 840, 113549, 1, 9, 14];
+
 /// id-at-countryName in RFC 5820
 const OID_COUNTRY_NAME :&[u64] = &[2, 5, 4, 6];
 /// id-at-organizationName in RFC 5820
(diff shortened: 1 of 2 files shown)


est31:
@djc thanks for this PR. It's definitely in scope. I've wanted to add this feature eventually myself. It's really cool to have CSR generation because now maybe [acme-client](https://github.com/onur/acme-client) could switch to rcgen.

As for testing I think openssl can be used for it. I suggest creating a `X509Req` and calling its [verify](https://docs.rs/openssl/0.10.23/openssl/x509/struct.X509Req.html#method.verify) function.

djc:
Added a test as per your suggestion.

djc:
(This does introduce the slight naming oddity that a `Certificate` object no longer always represents an actual certificate, because in the CSR case it's merely the `CertificateParams` combined with the certificate's key pair without the means to sign it into a usable certificate. Not sure if that's something you want to address as part of this PR?)

src/lib.rs
```diff
@@ -94,6 +94,9 @@ pub fn generate_simple_self_signed(subject_alt_names :impl Into<Vec<String>>) ->
 // Example certs usable as reference:
 // Uses ECDSA: https://crt.sh/?asn1=607203242
 
+/// pkcs-9-at-extensionRequest in RFC 2985
+const OID_PKCS_9_AT_EXTENSION_REQUEST: &[u64] = &[1, 2, 840, 113549, 1, 9, 14];
```
est31: code format nit: should be `OID_PKCS_9_AT_EXTENSION_REQUEST :&[u64]`

est31:
> naming oddity that a Certificate object no longer always represents an actual certificate

Yeah I think the API details are out of scope for the PR.

djc:
Also I think at least some versions of ACME require support for custom extensions, which doesn't add yet. But ACME is the motivating use case for me as well, so I can look at supporting that next.

src/lib.rs
djc: Fixed.

djc:
Thanks for the quick review!

est31:
@djc if you want a new release, just ask.

djc:
Will do. I don't think it's needed just yet.

djc:
Do you know if the acme-client maintainer is amenable to replacing openssl with a *ring*-based stack?

est31:
@djc I haven't had any communication with them if that's what you wonder about. I only thought it'd be a nice project.

djc:
The project seems a bit dead. I'm talking to the maintainer of an alternative here: https://github.com/breard-r/acmed/issues/2.

est31:
Oh yeah it kinda is. Getting acmed to use rcgen sounds like a good idea, too.

PR #9, opened 2019-06-10
Support user supplied keypairs

fzgregor:
- renamed DynKeyPair to KeyPair - might not be dynamic anymore
- made KeyPair public added From implementation
- moved serialized key pair into KeyPair struct
- introduced Certificate::from_params_and_keypair

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,7 +43,7 @@ use pem::Pem;
 use ring::digest;
 use ring::signature::{EcdsaKeyPair, Ed25519KeyPair};
 use ring::rand::SystemRandom;
-use ring::signature::KeyPair;
+use ring::signature::KeyPair as RingKeyPair;
 use untrusted::Input;
 use ring::signature::{self, EcdsaSigningAlgorithm, EdDSAParameters};
 use yasna::DERWriter;
@@ -56,8 +56,7 @@ use bit_vec::BitVec;
 /// A self signed certificate together with signing keys
 pub struct Certificate {
 	params :CertificateParams,
-	key_pair :DynKeyPair,
-	key_pair_serialized :Vec<u8>,
+	key_pair :KeyPair,
 }
 
 
@@ -183,6 +182,8 @@ pub struct CertificateParams {
 	pub distinguished_name :DistinguishedName,
 	pub is_ca :IsCa,
 	pub custom_extensions :Vec<CustomExtension>,
+	/// The certificate's key pair, a new random key pair will be generated if this is `None`
+	pub key_pair :Option<KeyPair>,
 	// To make the struct non-exhaustive
 	_hidden :(),
 }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -304,16 +303,20 @@ fn dt_to_generalized(dt :&DateTime<Utc>) -> GeneralizedTime {
 }
 
 impl Certificate {
-	/// Generates a new self-signed certificate from the given parameters
-	pub fn from_params(params :CertificateParams) -> Self {
-		let (key_pair, key_pair_serialized) = DynKeyPair::generate(&params.alg);
+	/// Generates a new certificate from the given parameters and key pair
+	pub fn from_params_and_keypair(params :CertificateParams, key_pair :KeyPair) -> Self {
```
est31: Instead of this function, could you add a `key_pair :Option<KeyPair>` field to `CertificateParams` instead?

src/lib.rs
```diff
@@ -577,46 +583,65 @@ enum SignAlgo {
...
+	EcKp(EcdsaKeyPair, Vec<u8>),
+	/// A Ed25519 key pair
+	EdKp(Ed25519KeyPair, Vec<u8>),
+}
+
+impl From<&[u8]> for KeyPair {
+	fn from(pkcs8 :&[u8]) -> KeyPair {
+		use std::iter::FromIterator;
+
+		if let Ok(edkp) = Ed25519KeyPair::from_pkcs8(Input::from(pkcs8)) {
```
est31: Could you add a `let input = Input::from(pkcs8);` to the front and then use it in each function? This makes the lines shorter. Possibly you'll have to clone it, most likely you won't.

src/lib.rs
fzgregor: Sure thing. Like this https://github.com/est31/rcgen/commit/33ea827ce1dcbee50d88dea2adfd8b4c548dbc2c#diff-b4aea3e418ccdb71239b96952d9cddb6L307?

src/lib.rs
fzgregor: Did that. I also move the pkc8 vector cloning in front: https://github.com/est31/rcgen/commit/33ea827ce1dcbee50d88dea2adfd8b4c548dbc2c#diff-b4aea3e418ccdb71239b96952d9cddb6R591

src/lib.rs
est31: Yup, that's good.

fzgregor:
Please note that I also moved the `serialize_*` functions into `KeyPair` and made the `generate()` function public. As `KeyPair` is public now this feels more correct to me. I'm not sure how you think about this. Given your comment at https://github.com/est31/rcgen/issues/7#issuecomment-500831262 you might feel different about this.

fzgregor:
Oh wow, you're fast.

est31:
@fzgregor making the `generate` function public hurts nobody and the code is needed for certificate generation. In my comment you linked I was only worried that I didn't want to expand scope of this crate infinitely :). Or rather, that I'm not sure yet whether I do want to expand it :).

fzgregor:
Glad to hear that.

PR #12, opened 2019-06-12
Support external ca-certificates

fzgregor:
This is an early, untested state not meant to be merged right now!

Supporting external ca-certificates by allowing the user to generate `CertificateParms` from a certificate in DER or PEM format. Extracting of data from user certificate is restricted to the bare minimum (distinguished names, signature algorithm, key) according to https://github.com/est31/rcgen/issues/7#issuecomment-500831262.

Open questions:
- Should the code prohibit the usage of `from_ca_cert` `CertificateParams` to generate new `Certificates`?
- Should there be more error variants, e.g. `NoCaCertificate`, `CertificateAndKeyMismatch`?
- How to iterate over `DnType` variants? (see https://github.com/est31/rcgen/compare/master...fzgregor:master#diff-b4aea3e418ccdb71239b96952d9cddb6R172) Is it necessary?


src/lib.rs
```diff
@@ -132,23 +140,41 @@ const OID_PE_ACME :&[u64] = &[1, 3, 6, 1, 5, 5, 7, 1, 31];
 #[allow(missing_docs)]
 /// The attribute type of a distinguished name entry
 pub enum DnType {
-	CountryName,
-	OrganizationName,
 	CommonName,
+	LocalityName,
```
est31: Could you put the change that adds these variants into a separate PR?

src/lib.rs
```diff
@@ -132,23 +140,41 @@ const OID_PE_ACME :&[u64] = &[1, 3, 6, 1, 5, 5, 7, 1, 31];
...
+			DnType::CountryName => OID_COUNTRY_NAME,
+			DnType::StreetName => OID_STREET_NAME,
+			DnType::CustomDnType(ref oid) => oid.as_slice(),
 			DnType::_Nonexhaustive => unimplemented!(),
 		};
 		ObjectIdentifier::from_slice(sl)
 	}
+
+	fn from_oid(slice: &[u64]) -> Self {
+		// TODO use specific types if they exist?
```
est31: It should be easy to make a match in the reverse direction. Could you do that? The compiler has smart logic to make such matches fast, at least I think I remember `nox` talking about something like that.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
 	}
 }
 
+impl CertificateParams {
+	/// Parses the ca certificate from the ASCII PEM format
+	/// 
+	/// See `from_ca_cert_der` for more details.
+	#[cfg(all(feature = "pem", feature = "x509-parser"))]
+	pub fn from_ca_cert_pem(pem_str :&str, key_pair: KeyPair) -> Result<Self, RcgenError> {
```
est31: style nit: please put the space before the colon.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+		let certificate = pem::parse(pem_str)
+			.or(Err(RcgenError::CouldNotParseCertificate))?;
+		let certificate_der :&[_] = &certificate.contents;
```
est31: Is this `&[_]` needed? Have you tried removing it? If it's not needed, it should be removed.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+
+	/// Parses the ca certificate from the DER format
+	/// 
+	/// This function is only of use if you have an existing ca certificate with which
+	/// you want to sign a certificate newly generated by `rcgen` using the
+	/// `serialize_der_with_signer()` or `serialize_pem_with_signer()` functions.
+	/// 
+	/// Will not check if certificate is a ca certificate!
+	#[cfg(feature = "x509-parser")]
+	pub fn from_ca_cert_der(ca_cert: &[u8], key_pair: KeyPair) -> Result<Self, RcgenError> {
```
est31: style nit: please put the space before the colon.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+		let (_remainder, x509) = x509_parser::parse_x509_der(ca_cert)
+			.or(Err(RcgenError::CouldNotParseCertificate))?;
+
+		// WDYT?
+		// if !x509.tbs_certificate.is_ca() {
```
est31: The other code doesn't do the check either. We make it clear enough that the parsing logic right now is so rudimentary that it only makes sense for signing other certs, so I think it's not needed.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+		// 	return Err(RcgenError::NoCaCertificate));
+		// }
+
+		let alg = SignatureAlgorithm::from_oid(x509.signature_algorithm.algorithm.iter().as_slice())?;
+
+		let mut dn = DistinguishedName::new();
+		for rdn in x509.tbs_certificate.subject.rdn_seq.iter() {
+			let attr = if rdn.set.len() > 1 {
```
est31: Could you add a comment that the `x509-parser` ensures that len >= 1? The spec requires that the length is > 0 but in theory the parser could have ignored that.

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+				// panic!("rcgen does not support distinguished names with more than one attribute");
```
est31: Could this be a proper comment?

src/lib.rs
```diff
@@ -216,6 +242,65 @@ impl Default for CertificateParams {
...
+
+			let dn_type = DnType::from_oid(attr.attr_type.iter().as_slice());
+			let dn_value = String::from_utf8(value.into())
+				.or(Err(RcgenError::CouldNotParseCertificate))?;
+			dn.push(dn_type, dn_value);
+		}
+
+		Ok(
+			CertificateParams {
+				alg: alg,
```
est31: style nit: please put space before and after the colon.

src/lib.rs
```diff
@@ -738,6 +831,27 @@ pub struct SignatureAlgorithm {
 	write_null_params :bool,
 }
 
+impl SignatureAlgorithm {
+	fn iter() -> std::slice::Iter<'static, &'static SignatureAlgorithm> {
+		static ALGORITHMS: [&'static SignatureAlgorithm; 4] = [
```
est31: style nit: please put the space before the colon.

Also, `'static` isn't needed in constants any more since... [1.17.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1170-2017-04-27). Also, a slice would be nicer than an array because then you don't have to update the length when you expand the list.

est31:
> Should there be more error variants, e.g. NoCaCertificate, CertificateAndKeyMismatch?

`CertificateAndKeyMismatch` would definitely be a good idea but I think it's best done in a separate PR.

src/lib.rs
fzgregor: Awesome!

src/lib.rs
fzgregor: Done

src/lib.rs
fzgregor: It's an assert now

src/lib.rs
fzgregor: Removed

src/lib.rs
fzgregor: Done

src/lib.rs
fzgregor: Done

src/lib.rs
fzgregor: Done

src/lib.rs
fzgregor: Done

fzgregor:
TBD: testing

fzgregor:
Blocked by https://github.com/rusticata/x509-parser/pull/5

est31:
> Blocked by rusticata/x509-parser#5

Yeah `UTCTime` is crap. It only supports dates between 1950 and 2050. `GeneralizedTime` on the other hand supports dates between `0 CE` and `10 000 CE`. That's much better.

est31:
This needs a rebase. Also, could you add a line ` - cargo test --verbose` to travis.yml?

fzgregor:
> This needs a rebase. 

I'll take care of that once `x509-parser` or WDYT - should `rcgen`'s `Cargo.toml` link directly to the git repository + hash? The fix landed in `master`...

> Also, could you add a line ` - cargo test --verbose` to travis.yml?

You mean a `cargo test` with `--features x509-parser`?

est31:
> should rcgen's Cargo.toml link directly to the git repository + hash?

After this PR, I'd like to make a release. So after you have confirmed that the test passes for the git repo, you can ask the maintainer of x509-parser for a new release.

> You mean a cargo test with --features x509-parser?

Erm yeah of course I meant that :).

fzgregor:
Had to introduce ordered preserving for DistinguishedName. Ordering mismatch, due to HashMap in DN, was failing webpki test non-deterministically.

src/lib.rs
```diff
@@ -214,6 +257,61 @@ impl Default for CertificateParams {
...
+
+			let dn_type = DnType::from_oid(attr.attr_type.iter().as_slice());
+			let dn_value = String::from_utf8(value.into())
+				.or(Err(RcgenError::CouldNotParseCertificate))?;
+			dn.push(dn_type, dn_value);
+		}
+
+		Ok(
+			CertificateParams {
+				alg : alg,
```
est31: Nit: You can just do `alg,` here.

fzgregor: Done

PR #20, opened 2019-06-25
Expose KeyPair interface

fzgregor:
Offering an interface to obtain a `KeyPair`'s public key in DER and PEM and access `is_compatible()`

Avoiding code duplication by moving public key serialization into `KeyPair`.

Addresses https://github.com/est31/rcgen/issues/15 and https://github.com/est31/rcgen/issues/18

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -455,11 +455,7 @@ impl Certificate {
 				}
 			});
 			// Write subjectPublicKeyInfo
-			writer.next().write_sequence(|writer| {
-				self.params.alg.write_oids_sign_alg(writer.next());
-				let pk = self.key_pair.public_key();
-				writer.next().write_bitvec_bytes(&pk, pk.len() * 8);
-			});
+			self.key_pair.serialize_public_key_der(writer.next());
 			// Write extensions
 			writer.next().write_tagged(Tag::context(0), |writer| {
 				writer.write_sequence(|writer| {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -859,11 +862,29 @@ impl KeyPair {
...
+	/// The returned string can be interpreted with `openssl pkey --inform PEM -pubout -pubin -text`
+	#[cfg(feature = "pem")]
+	pub fn public_key_pem(&self) -> String {
+		let p = Pem {
+			tag : "PUBLIC KEY".to_string(),
+			contents : self.public_key_der(),
+		};
+		pem::encode(&p)
+	}
+	/// Serializes the key in PKCS#8 format in DER
```
est31: Could you explicitly point out that this includes the private key? The current statement is ambiguous. Either "the private key" or "the key pair" is fine.

src/lib.rs
```diff
@@ -859,11 +862,29 @@ impl KeyPair {
...
 	pub fn serialize_der(&self) -> Vec<u8> {
 		self.serialized_der.clone()
 	}
-	/// Serializes the private key in PEM format
+	/// Serializes the key in PKCS#8 format in PEM
```
est31: Same ambiguity issue here.

src/lib.rs
```diff
@@ -883,6 +904,37 @@ pub struct SignatureAlgorithm {
 	write_null_params :bool,
 }
 
+impl fmt::Debug for SignatureAlgorithm {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
est31: Colon placement ` :`.

src/lib.rs
```diff
@@ -883,6 +904,37 @@ pub struct SignatureAlgorithm {
...
+		} else if self == &PKCS_ED25519 {
+			write!(f, "PKCS_ED25519")
+		} else {
+			write!(f, "Unknown")
+		}
+    }
+}
+
+impl PartialEq for SignatureAlgorithm {
+    fn eq(&self, other: &Self) -> bool {
```
est31: colon placement

src/lib.rs
fzgregor: very good point

PR #22, opened 2019-07-02
make KeyPair::public_key public

bddap:
Makes public key easily accessible.

This way the user does not need to serialize to der, then deserialize to get the raw pk.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -833,7 +833,7 @@ impl KeyPair {
 			writer.next().write_bitvec_bytes(&pk, pk.len() * 8);
 		})
 	}
-	/// Get the public key from this key pair
+	/// Get the public key of this key pair
 	pub fn public_key(&self) -> &[u8] {
 		match &self.kind {
 			KeyPairKind::Ec(kp) => kp.public_key().as_ref(),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -833,7 +833,8 @@ impl KeyPair {
 			writer.next().write_bitvec_bytes(&pk, pk.len() * 8);
 		})
 	}
-	fn public_key(&self) -> &[u8] {
+	/// Get the public key from this key pair
```
est31: could you replace "from" with "of"

bddap: done

est31:
Hmm I wonder what format the raw key is in. Do you know how to call it?

bddap:
[rcgen::KeyPair::public_key](https://github.com/est31/rcgen/blob/master/src/lib.rs#L836-L842) wraps [ring::signature::KeyPair::public_key](https://briansmith.org/rustdoc/ring/signature/trait.KeyPair.html#tymethod.public_key).

The format is raw bytes. I assume the public key is in network endian, but ring doesn't specify.

est31:
Oh ok.

bddap:
@est31 I've been doing some testing.

```rust
// This test passes
let kp = KeyPair::generate(&rcgen::PKCS_ECDSA_P256_SHA256).unwrap();
assert_eq!(kp.public_key()[0], 4);
assert_eq!(kp.public_key().len(), 65);
```

The first byte of the public key is always 4, which I didn't expect. Leads me to believe there is a tag at the start of the public key.

It's not worrisome behavior, just unexpected. Does this make sense to you?

PR #31, opened 2020-03-27
Fix multiple RFC 5280 violations, Key Identifier improvements

nthuemmel:
This PR:

* Fixes a wrong extension format of the `subject key identifier` extension, which led to various X.509 client implementations (WolfSSL, golang) to reject the generated CA certificates
* Changes how key identifiers are calculated
* Fixes minor CA non-conformance issues (basic constraints were not marked critical, `authority key identifier` extension was missing)

Detailed descriptions are part of the commit messages.


nthuemmel:
@est31 Thanks for the reply! Allowing the user to choose a different algorithm for key identifiers is reasonable, and I think the `CertificateParams` is the right place to put it in. We only need to be careful to choose the correct parameter - the Authority Key Identifier needs to use the parameter of the issuer cert, while the Subject Key Identifier needs to use the parameter of the new cert params.

As for the unconditional inclusion of the Authority Key Identifier, you are probably right - I may have misread the RFC there. I have found no authoritative sources that imply the strict necessity of the extension during a quick search. I would change it to be optional, controlled by the `CertificateParams`. Would that be acceptable? I assume you would like to have it disabled by default, correct?

est31:
> I would change it to be optional, controlled by the CertificateParams. Would that be acceptable? I assume you would like to have it disabled by default, correct?

Yes, that'd be great! If we find sources that give us an interpretation towards one direction in the future, the default can still be changed.

PR #41, opened 2020-11-16
Add support for creating certificates from CSRs

djc:
Depends on https://github.com/rusticata/x509-parser/pull/58 and (transitively) https://github.com/rusticata/oid-registry/pull/1, but if you have any early feedback that would be great!


djc:
The upstream PRs have been merged (but not yet released).

est31:
Thanks for this PR! I'll review it more closely later but as a remark for the start, I'm not sure if the `from_name` function and the TryFrom impl should be added. This makes the public rcgen API use types from the `x509-parser` crate which means that semver breaking updates cause us to have them too. In general though, it's good to have CSR support in rcgen.

est31:
Also: could you add tests? :)

djc:
About the semver compatibility issues:

- `DistinguishedName::from_name()` is actually private, so I don't think that should be a concern?
- The main entry point `CertificateSigningRequest::from_x509()` itself exposes an x509-parser type in the public API. If it would take DER or PEM bytes, that means either (a) rcgen should do signature verification or (b) the CSR has to be parsed twice if the users wants to also do signature verification. What do you prefer, or do you see another option?

Will add some tests.

est31:
> from_name() is actually private

good point, case rested.

> either (a) rcgen should do signature verification or (b) the CSR has to be parsed twice if the users wants to also do signature verification. What do you prefer, or do you see another option?

Is it hard to do (a)? rcgen's main task is to create certificates, where you can't do much wrong, but doing signature verification wrong doesn't seem *that* risky, as an attacker having a certificate for a public key they don't own doesn't give them *that* many more capabilities. But the same time, it's not the goal of rcgen to verify things, but to create them.

djc:
I changed it to verify the signature using the signature verification feature from x509-parser (which I wrote for their CSR support -- this seemed more convenient than reproducing the verification logic in rcgen. I've also moved the `TryFrom` impl to a private `try_from()` method on `SanType`. Let me know if this is okay, then I'll write up some tests next.

est31:
@djc sounds like a plan! :+1:

djc:
Updated to use x509-parser 0.9.

est31:
Wonderful! Will take a look at it later today.

src/lib.rs
```diff
@@ -156,6 +156,19 @@ pub enum SanType {
 }
 
 impl SanType {
+	#[cfg(feature = "x509-parser")]
+	fn try_from(name: &x509_parser::extensions::GeneralName<'_>) -> Result<Self, RcgenError> {
```
est31, line 160: Could you rename this to `try_from_general` or something? `try_from` is very generic, makes mainly sense if part of the TryFrom trait impl. Also: colon placement.

src/lib.rs
```diff
@@ -403,6 +416,31 @@ impl DistinguishedName {
 			iter :self.order.iter()
 		}
 	}
+
+	#[cfg(feature = "x509-parser")]
+	fn from_name(name: &x509_parser::x509::X509Name) -> Result<Self, RcgenError> {
```
est31, line 421: colon placement

src/lib.rs
```diff
@@ -424,6 +462,101 @@ impl <'a> Iterator for DistinguishedNameIterator<'a> {
...
+
+/// Data for a certificate signing request
+#[allow(missing_docs)]
+pub struct CertificateSigningRequest {
+	pub params :CertificateParams,
+	pub public_key :PublicKey,
+}
+
+impl CertificateSigningRequest {
+	/// Parse certificate signing request from the ASCII PEM format
```
est31, line 490:
```suggestion
	/// Parse a certificate signing request from the ASCII PEM format
```

src/lib.rs
```diff
@@ -424,6 +462,101 @@ impl <'a> Iterator for DistinguishedNameIterator<'a> {
...
+	///
+	/// See `from_der()` for more details.
+	#[cfg(all(feature = "pem", feature = "x509-parser"))]
+	pub fn from_pem(pem_str :&str) -> Result<Self, RcgenError> {
+		let csr = pem::parse(pem_str)
+			.or(Err(RcgenError::CouldNotParseCertificationRequest))?;
+		Self::from_der(&csr.contents)
+	}
+
+	/// Parse certificate signing request from DER-encoded bytes
```
est31, line 500:
```suggestion
	/// Parse a certificate signing request from DER-encoded bytes
```

src/lib.rs
```diff
@@ -424,6 +462,101 @@ impl <'a> Iterator for DistinguishedNameIterator<'a> {
...
+	///
+	/// Currently, this only supports the `Subject Alternative Name` extension.
```
est31, line 502: Maybe say here that for other extensions, an error is returned.

src/lib.rs
```diff
@@ -424,6 +462,101 @@ impl <'a> Iterator for DistinguishedNameIterator<'a> {
...
+			public_key: PublicKey { alg, raw },
+		})
+	}
+	/// Serializes the requested certificate, signed with another certificate's key, in binary DER format
+	pub fn serialize_der_with_signer(&self, ca :&Certificate) -> Result<Vec<u8>, RcgenError> {
+		self.params.serialize_der_with_signer(&self.public_key, ca)
+	}
+	/// Serializes the requested certificate, signed with another certificate's key, to the ASCII PEM format
+	///
+	/// *This function is only available if rcgen is built with the "pem" feature*
```
est31, line 549: Thanks for adding this! :heart:

est31 (changes requested):
Ok except for some nits.

djc:
Addressed all the nits.

est31 (approved):
Thanks! It's a bit sad that the nom update brought bit-vec into our lockfile which has way too many dependencies for the simple task it's doing, but that's more of a long term problem. Merging.

djc:
Yeah, x509-parser should maybe default to depending on nom with default-features = false... Thanks for merging! Do you plan on publishing a new release soon?

est31:
@djc waiting for travis CI to pass, then I'll release. After that I'll switch to github actions because I don't have patience :).

djc:
Yeah, I was surprised you don't have Actions here, yet!

est31:
0.8.6 is out!

djc:
Thanks!

PR #42, opened 2020-11-18
Disable botan dependency and tests on macOS

djc:
Building Botan on macOS apparently is not supported upstream:

"Currently, systems other than Windows and POSIX (such as VMS,
MacOS 9, OS/390, OS/400, …) are not supported by the build system,
primarily due to lack of access."

https://botan.randombit.net/handbook/building.html


est31 (approved):
Thanks!

PR #43, opened 2020-12-01
Fix mixed signing algorithms & Botan dependency

nthuemmel:
This PR fixes two bugs:

* Certificates which were signed by a CA using a different signing algorithm than the certificate itself were not usable, because the wrong signing algorithm identifier was serialized for the issuer's signature and in the tbsCertificate. This has been fixed, and tests added.
* Botan was always pulled in as a dependency. It is now only pulled in as a dev-dependency only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,5 +34,5 @@ features = ["x509-parser"]
 openssl = "0.10"
 webpki = "0.21"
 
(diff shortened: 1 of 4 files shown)


est31:
Oh no how could I miss that change for botan. I guess after this I'll immediately do another release.

est31 (approved):
Thanks!

PR #45, opened 2020-12-03
Strip nanos from DateTime<Utc>

### LINKED ISSUE #44: DateTime nanoseconds not being stripped
It appears that `write_dt_utc_or_generalized()` is not stripping nanoseconds for `chrono::DateTime<Utc>`.

Simple example is as follows: 
```
// create certificate_params
...
let current_time = Utc::now();
// set not_before and not_after
...
// generate certificate
let cert = Certificate::from_params(certificate_params)?;

```
`libs.rs`:
```
fn write_dt_utc_or_generalized(writer: DERWriter, dt: &DateTime<Utc>) -> Result<(), RcgenError> {
    // RFC 5280 requires CAs to write certificate validity dates
    // below 2050 as UTCTime, and anything starting from 2050
    // as GeneralizedTime [1]. The RFC doesn't say anything
    // about dates before 1950, but as UTCTime can't represent
    // them, we have to use GeneralizedTime if we want to or not.
    // [1]: https://tools.ietf.org/html/rfc5280#section-4.1.2.5
    if (1950..2050).contains(&dt.year()) {
        let ut = UTCTime::from_datetime::<Utc>(dt); // fails here
        writer.write_utctime(&ut);
    } else {
        let gt = dt_to_generalized(dt)?;
        writer.write_generalized_time(&gt);
    }
    Ok(())
}
```

trevor-crypto:
When creating `CertificateParams` and using `DateTime<Utc>`, specifically
`chrono::Utc::now()`, the nanoseconds do not get stripped off before
attempting to call `UTCTime::from_datetime` (yasna), causing an error due to the ASN.1
standard

Closes #44

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1037,12 +1037,14 @@ pub fn date_time_ymd(year :i32, month :u32, day :u32) -> DateTime<Utc> {
 	DateTime::<Utc>::from_utc(naive_dt, Utc)
 }
 
-fn dt_strip_nanos(dt :&DateTime<Utc>) -> Result<DateTime<Utc>, RcgenError> {
+fn dt_strip_nanos(dt :&DateTime<Utc>, allow_leap: bool) -> Result<DateTime<Utc>, RcgenError> {
 	// Set nanoseconds to zero (or to one leap second if there is a leap second)
 	// This is needed because the GeneralizedTime serializer would otherwise
 	// output fractional values which RFC 5280 explicitly forbode [1].
+	// UTCTime cannot express fractional seconds or leap seconds
+	// therefore, it needs to be stripped of nanoseconds fully.
 	// [1]: https://tools.ietf.org/html/rfc5280#section-4.1.2.5.2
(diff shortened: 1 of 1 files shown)


est31 (changes requested):
Could you add a test? Also, maybe explain in the comment why stripping is not just useful for `GeneralizedTime` but also `UtcTime`.

trevor-crypto:
> Could you add a test? Also, maybe explain in the comment why stripping is not just useful for `GeneralizedTime` but also `UtcTime`.

Sure thing

trevor-crypto:
Writing the tests made me realize that `UTCTime` does not support leap seconds like `GeneralizedTime` does -- I noted this in the comments.

est31 (approved):
Thanks!

est31:
@trevor-crypto I've merged your PR to the master branch. If you want me to make a release, please say say so.

trevor-crypto:
> @trevor-crypto I've merged your PR to the master branch. If you want me to make a release, please say say so.

@est31 That would be great. Thank you!

PR #46, opened 2020-12-19
Small readme example code fix

Nukesor:
Hey :)

I just noticed, that there's an unnecessary `.unwrap()` in the example code of your Readme.
Nothing big.
https://docs.rs/rcgen/0.8.9/rcgen/struct.Certificate.html#method.serialize_private_key_pem

Thanks for providing this library!


est31 (approved):
Thanks!

PR #48, opened 2021-01-19
Impl Zeroize for KeyPair and Certificate

didier-wenzek:
Here is a proposed implementation for https://github.com/est31/rcgen/issues/47.

This proposition is to have a new optional feature, named `zeroize`. With that option enabled, a dependency to `zeroize` is added and the `Zeroize` trait is implemented for `KeyPair`, and all structures that include such a key pair (`Certificate`, `Certificate`, `CertificateSigningRequest` and `CertificateParams`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,6 +347,7 @@ dependencies = [
  "webpki",
  "x509-parser",
  "yasna",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


est31 (changes requested):
Can you please use tabs for indentation? Otherwise looks great.

didier-wenzek:
Updated using tabs for indentation.

est31 (approved):
Thanks!

didier-wenzek:
Thank you!

PR #49, opened 2021-02-24
fix docs link

andrewbanchich:


est31 (approved):
Well it's not really wrong, but after checking some random popular repos (eg proc-macro-hack) it seems to be the common way to add a docs badge. Thanks.

PR #50, opened 2021-02-24
Allow KeyPair::from_pem to accept any AsRef<[u8]>

andrewbanchich:


est31:
It's a bit non-obvious to folks, at least to me, that `str` implements `AsRef<[u8]>`. So when you read the function you don't really know what it needs to be passed. All it saves is an invocation of the `str::from_utf8` function, right?

Is this PR motivated by a concrete use case? Does your code obtain data in a u8-slice-like form that's not a `str`? Have you, when using `from_pem`, had to convert to a str?

andrewbanchich:
> It's a bit non-obvious to folks, at least to me, that `str` implements `AsRef<[u8]>`.

I could add an example to the docs which uses a `&str` and other types which implement that.

I added it because `&str` must be valid UTF-8, but a PEM doesn't have to be, which means users of `rcgen` may have a valid PEM which they can't use. It also is more ergonomic because if they already have bytes, they don't need to convert it to a `&str` when the underlying `pem` crate doesn't care.

est31:
Alright after having been in the situation of wanting this myself for 3a41c1107dd700d69ce536a0a1590740f54efe0b , I think this change makes sense. Could you maybe revert the whitespace changes? The old state is as intended.

est31:
Also a rebase would be helpful.

PR #51, opened 2021-02-24
cargo fmt

andrewbanchich:
Using rustfmt would help contributions adhere to a style. Some people (including myself) also have rustfmt run on save, which made a simple change I PRed earlier have an enormous diff, which I had to disable auto formatting for.


est31:
Thanks but rustfmt currently can't be configured to adhere to the style I want. I'd have to fork it and add the options myself.

PR #54, opened 2021-04-12
Implement Hash for PublicKey and some more minor improvements

zurborg:
I want to check for duplicate public keys in CSRs. My idea is a `HashMap<&PublicKey, CertificateSigningRequest>`, but `PublicKey` needs to implement `Hash`.

The problem is `ring` does not implement much. I think my `Hash` impl for `SignAlgo` is sufficient, basically a hash over the raw data of a public key could also do fine. For that reason, I made the `PublicKeyData` trait public, in order to access the raw data via `PublicKeyData::raw_bytes` (but without any documentation yet).

And there are some other non-breaking improvements like `impl Default for DistinguishedName`, `impl FromStr for CidrSubnet` and some other clippy suggestions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use rcgen::generate_simple_self_signed;
 # fn main () {
 // Generate a certificate that's valid for "localhost" and "hello.world.example"
 let subject_alt_names = vec!["hello.world.example".to_string(),
-    "localhost".to_string()];
+	"localhost".to_string()];
 
 let cert = generate_simple_self_signed(subject_alt_names).unwrap();
 println!("{}", cert.serialize_pem().unwrap());
@@ -78,7 +78,7 @@ extern crate rcgen;
 use rcgen::generate_simple_self_signed;
 # fn main () {
 let subject_alt_names :&[_] = &["hello.world.example".to_string(),
-    "localhost".to_string()];
+	"localhost".to_string()];
 
 let cert = generate_simple_self_signed(subject_alt_names).unwrap();
 // The certificate is now valid for localhost and the domain "hello.world.example"
@@ -243,6 +243,28 @@ macro_rules! mask {
 }
 
 impl CidrSubnet {
+	/// Obtains the CidrSubnet from the well-known
+	/// addr/prefix notation.
+	/// ```
+	/// # use std::str::FromStr;
+	/// # use rcgen::CidrSubnet;
+	/// // The "192.0.2.0/24" example from
+	/// // https://tools.ietf.org/html/rfc5280#page-42
+	/// let subnet = CidrSubnet::from_str("192.0.2.0/24").unwrap();
+	/// assert_eq!(subnet, CidrSubnet::V4([0xC0, 0x00, 0x02, 0x00], [0xFF, 0xFF, 0xFF, 0x00]));
+	/// ```
+	#[allow(clippy::should_implement_trait, clippy::result_unit_err)]
+	pub fn from_str(s :&str) -> Result<Self, ()> {
+		let mut iter = s.split('/');
+		if let (Some(addr_s), Some(prefix_s)) = (iter.next(), iter.next()) {
+			let addr = IpAddr::from_str(addr_s).map_err(|_| ())?;
+			let prefix = u8::from_str(prefix_s).map_err(|_| ())?;
+			Ok(Self::from_addr_prefix(addr, prefix))
+		} else {
+			Err(())
+		}
+	}
+
 	/// Obtains the CidrSubnet from an ip address
 	/// as well as the specified prefix number.
 	///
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1538,6 +1556,15 @@ impl PartialEq for SignatureAlgorithm {
 
 impl Eq for SignatureAlgorithm {}
 
+impl Hash for SignatureAlgorithm {
```
est31, line 1574: Why not derive it?

src/lib.rs
```diff
@@ -1215,6 +1217,22 @@ enum SignAlgo {
...
 }
 
+impl Hash for SignAlgo {
+	fn hash<H: Hasher>(&self, state: &mut H) {
+		match self {
+			SignAlgo::EcDsa(_) => {
+				"SignAlgo::EcDsa".hash(state);
+			}
+			SignAlgo::EdDsa(_) => {
+				"SignAlgo::EdDsa".hash(state);
```
est31, line 1241: This strips away all the useful data inside the SignAlgo, especially which *kind* of EdDsa algo it is. Instead I think you should convert the interior pointer to a static into an integer and hash that value. Alternatively one can extend the enum with another field and use that field for hashing.

src/lib.rs
```diff
@@ -1508,7 +1508,8 @@ impl PublicKeyData for KeyPair {
 	}
 }
 
-trait PublicKeyData {
+#[allow(missing_docs)]
+pub trait PublicKeyData {
```
est31, line 1523: Why is making this public required?

src/lib.rs
```diff
@@ -589,7 +595,7 @@ impl Default for CertificateParams {
 		// not_before and not_after set to reasonably long dates
 		let not_before = date_time_ymd(1975, 01, 01);
 		let not_after = date_time_ymd(4096, 01, 01);
-		let mut distinguished_name = DistinguishedName::new();
+		let mut distinguished_name = DistinguishedName::default();
```
est31, line 603: Please don't use Default here, I'm not a fan of it. I like new far more.

src/lib.rs
```diff
@@ -243,26 +243,6 @@ macro_rules! mask {
...
-	/// addr/prefix notation.
-	/// ```
-	/// # use std::str::FromStr;
-	/// # use rcgen::CidrSubnet;
-	/// // The "192.0.2.0/24" example from
-	/// // https://tools.ietf.org/html/rfc5280#page-42
-	/// let subnet = CidrSubnet::from_str("192.0.2.0/24").unwrap();
-	/// assert_eq!(subnet, CidrSubnet::V4([0xC0, 0x00, 0x02, 0x00], [0xFF, 0xFF, 0xFF, 0x00]));
-	/// ```
-	pub fn from_str(s :&str) -> Result<Self, ()> {
```
est31, line 255: This is an API breaking change. I also wonder how implementing this trait benefits users. In fact usage is harder now as you have to import the `FromStr` trait. It's also way more hidden in the rustdoc. So please don't do it.

src/lib.rs
```diff
@@ -598,8 +598,8 @@ pub struct CertificateParams {
 impl Default for CertificateParams {
 	fn default() -> Self {
 		// not_before and not_after set to reasonably long dates
-		let not_before = date_time_ymd(1975, 01, 01);
-		let not_after = date_time_ymd(4096, 01, 01);
+		let not_before = date_time_ymd(1975, 1, 1);
```
est31, line 601: 01 is in accordance with how dates are commonly written. I like it better that way.

src/lib.rs
```diff
@@ -306,7 +306,7 @@ impl FromStr for CidrSubnet {
 	/// assert_eq!(subnet, CidrSubnet::V4([0xC0, 0x00, 0x02, 0x00], [0xFF, 0xFF, 0xFF, 0x00]));
 	/// ```
 	fn from_str(s :&str) -> Result<Self, Self::Err> {
-		let mut iter = s.split("/");
+		let mut iter = s.split('/');
```
est31, line 309: :+1:

src/lib.rs
```diff
@@ -78,7 +78,7 @@ extern crate rcgen;
 use rcgen::generate_simple_self_signed;
 # fn main () {
 let subject_alt_names :&[_] = &["hello.world.example".to_string(),
-	"localhost".to_string()];
+    "localhost".to_string()];
```
est31, line 81: Why does this whitespace change exist? What does clippy complain about it?

est31 (changes requested):
Thanks for the PR. The core idea of the PR is good, I'm not that sure about some of the minor improvements.

src/lib.rs
zurborg: Clippy complains about deriving while `PartialEq` is implemented explicity: https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

src/lib.rs
zurborg: For my purpose this is sufficient, but the only way to retain `EcdsaSigningAlgorithm` or `EdDSAParameters` is via the `Debug` trait. See https://github.com/zurborg/rcgen/commit/14232c8d24f71cc2328f6879fc5e3583e763073d

src/lib.rs
zurborg: I don't know why the methods kept to be private. All methods in the trait are immutable.

src/lib.rs
zurborg: I'm confused because `01` is an octal number, even when `01 == 1`

src/lib.rs
zurborg: See here: https://rust-lang.github.io/rust-clippy/master/index.html#tabs_in_doc_comments

src/lib.rs
zurborg: I think using traits benefits almost everyone. The auto-generated documentation for this crate will still mention `from_str`.

src/lib.rs
est31: Thanks. rcgen uses tabs for indentation so the lint is wrong. Please don't do it.

src/lib.rs
est31: It might be sufficient for your purpose, but it would be quite unexpected for people if the hash were the same for two distinct signature algorithms.

src/lib.rs
est31: Good point, but Rust has no octal literals.

src/lib.rs
est31: Understandable. Can you still add a comment on why there is no derive here?

src/lib.rs
zurborg: Why? You are using `Default::default()` at some other places, too. I think the `Default` implementation here should be coherent with concluded calls of `default()`.

src/lib.rs
zurborg: Oh, you're right. In C `01` is actutal an octal literal. A clippy hint solves it, too.

est31:
I'm closing this PR as it's becoming a mess (e.g. it has a merge commit). I've added two commits from the diff 35756f57a2aa1b81fee3c8de387434e0ad30ebf3 and 52858c91f130ac06b52724ebf6b1256dd6a89e12 to the master branch. In principle, a Hash implementation for PublicKey is welcome, but this is my project. I don't want to have to explain why I prefer `default` or leading zeroes. PRs that focus on adding a Hash implementation, esp one that doesn't use Debug but does it properly and performantly without string involvement, would be nice.

PR #55, opened 2021-04-13
Implement Hash for SignatureAlgorithm and PublicKey

zurborg:
This is another attempt for implementing Hash for `PublicKey`. I analyzed `SignatureAlgorithm::eq` and the solely compared field is `oids_sign_alg`. To my understanding of how collections rely on a `Hash` implementation, hashing any other field in `SignatureAlgorithm` violates this contract: `k1 == k2 -> hash(k1) == hash(k2)`. It looks kinda odd to me. Simple test cases are included, and all are passing.

Besides, `SignAlgo` can implement `Hash` with `std::ptr::hash()`, but this is apparently no longer needed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,7 @@ use std::convert::TryFrom;
 use std::error::Error;
 use std::net::IpAddr;
 use std::str::FromStr;
+use std::hash::{Hash, Hasher};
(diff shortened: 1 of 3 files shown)


est31 (approved):
Thanks!

PR #57, opened 2021-05-04
Implement `KeyPair::from_remote`

### LINKED ISSUE #56: Ability to generate a certificate from (remote) raw private key
Currently there is no way to generate a full X.509 certificate from a raw private and public key.
Additionally support for a "remote" raw private key would be amazing too, with remote meaning that no direct access is available, but a callback to signing is.

At the moment my ad-hoc solution is to copy some parts of rcgen.

I would propose to add an `enum` of sorts to `rcgen::CertificateParams::key_pair` that has the ability to take in a `Box` to a trait similar to [`rustls::sign::SigningKey`](https://docs.rs/rustls/0.19.1/rustls/sign/trait.SigningKey.html) and the public key in it's raw form.

Intending to make a PR myself if it's decided that this falls into the scope of rcgen.

My use case:
I'm currently writing a library that enables cross-platform access to the OS certificate vault and store. This includes generation, storage and some level of editing.

The goal is to be able to generate a key pair inside the OS vault and use that for example in rustls. When done that way, there is no access to the private key directly. This works great with rustls because it offers [`SigningKey`](https://docs.rs/rustls/0.19.1/rustls/sign/trait.SigningKey.html), which can call the OS vault signing function to sign a message without direct access to the private key.

The missing part is X.509 certificate generation. All three OS's (Linux, MacOS and Windows) generate only raw keys. Which is a great option for the future: https://github.com/ctz/rustls/issues/423.
But right now and in some other use cases, for example using it on a server with a HSM for example, you need to generate a X.509 certificate from the already generated raw keys in addition to the private key not being directly accessible.

At the moment the usual solution is to use tools or OS API's to generate certificates, it would be great to allow that to be replaced by rcgen.

daxpedda:
Implements X.509 certificate generation from a remote private key.

Fixes #56.


src/lib.rs
```diff
@@ -1215,11 +1215,15 @@ impl Certificate {
 		Ok(pem::encode(&p))
 	}
 	/// Serializes the private key in PKCS#8 format
+	///
+	/// Panics if called with on a remote key pair.
```
est31:
```suggestion
	/// Panics if called on a remote key pair.
```

src/lib.rs
```diff
@@ -1215,11 +1215,15 @@ impl Certificate {
...
 	pub fn serialize_private_key_der(&self) -> Vec<u8> {
 		self.key_pair.serialize_der()
 	}
 	/// Serializes the private key in PEM format
 	///
+	/// Panics if called with on a remote key pair.
```
est31:
```suggestion
	/// Panics if called on a remote key pair.
```

src/lib.rs
```diff
@@ -1274,6 +1290,29 @@ impl KeyPair {
 		let private_key_der :&[_] = &private_key.contents;
 		Ok(private_key_der.try_into()?)
 	}
+
+	/// Parses the key pair from a raw public key and a remote private key
```
est31, line 1294:
```suggestion
	/// Obtains the key pair from a raw public key and a remote private key
```

src/lib.rs
```diff
@@ -1274,6 +1290,29 @@ impl KeyPair {
...
+			serialized_der : Vec::new(),
+		})
+	}
+}
+
+/// A private key that is not directly accessible, but can be used to sign messages
+///
+/// This is used to enable generating certificates from a remote and raw private key.
+pub trait RemoteKeyPair {
+	/// Return's the public key of this key pair
```
est31, line 1308:
```suggestion
	/// Returns the public key of this key pair
```

src/lib.rs
```diff
@@ -1476,7 +1520,13 @@ impl KeyPair {
 		pem::encode(&p)
 	}
 	/// Serializes the key pair (including the private key) in PKCS#8 format in DER
+	///
+	/// Panics if called with on a remote key pair.
```
est31:
```suggestion
	/// Panics if called on a remote key pair.
```

src/lib.rs
```diff
@@ -1274,6 +1290,29 @@ impl KeyPair {
...
+		})
+	}
+}
+
+/// A private key that is not directly accessible, but can be used to sign messages
+///
+/// This is used to enable generating certificates from a remote and raw private key.
+pub trait RemoteKeyPair {
+	/// Return's the public key of this key pair
+	fn public_key(&self) -> &[u8];
```
daxpedda, line 1309: I could store the public key within `KeyPair::Remote` directly instead.

tests/webpki.rs
```diff
@@ -232,6 +232,48 @@ fn test_webpki_separate_ca_with_other_signing_alg() {
 				&webpki::ED25519, &webpki::ECDSA_P256_SHA256, sign_msg_ed25519);
 }
 
+#[test]
+fn from_remote() {
+	struct Remote(EcdsaKeyPair);
+
+	impl RemoteKeyPair for Remote {
+    	fn public_key(&self) -> &[u8] {
```
est31, line 240: Indentation problems, please only use tabs

tests/webpki.rs
```diff
@@ -232,6 +232,48 @@ fn test_webpki_separate_ca_with_other_signing_alg() {
...
+			self.0.public_key().as_ref()
+		}
+
+    	fn sign(&self, msg :&[u8]) -> Result<Vec<u8>, rcgen::RcgenError> {
```
est31, line 244: Indentation problem, please only use tabs

tests/webpki.rs
```diff
@@ -232,6 +232,48 @@ fn test_webpki_separate_ca_with_other_signing_alg() {
...
+			let system_random = SystemRandom::new();
+			self.0.sign(&system_random, msg)
+				.map(|s| s.as_ref().to_owned())
+				.map_err(rcgen::RcgenError::from)
+		}
+
+    	fn algorithm(&self) -> &'static rcgen::SignatureAlgorithm {
```
est31, line 251: Indentation problem, please only use tabs

src/lib.rs
```diff
@@ -1274,6 +1290,29 @@ impl KeyPair {
...
+
+	/// Signs `msg` using the selected algorithm
+	fn sign(&self, msg :&[u8]) -> Result<Vec<u8>, RcgenError>;
+
+	/// Reveals which algorithm will be used when you call `sign()`
+	fn algorithm(&self) -> &'static SignatureAlgorithm;
```
daxpedda, line 1315: This is already stored in `KeyPair` anyway, I'm not sure if it's useful to have it in the trait at all.

est31 (changes requested):

src/lib.rs
daxpedda: Sorry about that, I wasn't actually done going through it again!

daxpedda:
Sorry about the mess, I will squash in a moment.

tests/webpki.rs
daxpedda: Done.

est31 (approved):
Thanks!

PR #58, opened 2021-05-19
Adding keyUsage extension.

jaredwolff:
Added:
* Use of bit_vec for keyUsage generation
* Generation of keyUsage using enum mapping
* Basic test to confirm functionality


src/lib.rs
```diff
@@ -940,6 +990,28 @@ impl NameConstraints {
 	}
 }
 
+/// One of the purposes contained in the [key usage](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)
```
est31, line 993:
```suggestion
/// One of the purposes contained in the [key usage](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3) extension
```

src/lib.rs
```diff
@@ -765,6 +772,49 @@ impl CertificateParams {
...
+										bits.grow( (index + 1) - bits.len(), false);
+									}
+
+									bits.set(index, true);
+								}
+
+								let der = yasna::encode_der(&bits);
+
+								// Write them
+								writer.next().write_bytes(&der);
```
est31, line 813: Before we're adding an entire dependency, what about trying `write_bitvec_bytes` instead? It shouldn't be too complicated to manually write a bitvec. It seems it can't be more than 9 bits anyways, no?

est31 (changes requested):

src/lib.rs
```diff
@@ -1770,4 +1842,49 @@ mod tests {
 			assert!(dt_to_generalized(&dt).is_ok());
 		}
 	}
+
+	#[test]
+	fn test_with_key_usages() {
```
est31, line 1847: Can you put this test into `tests/generic.rs`?

Also it would be great if there were at least one test with webpki.

est31: Wait no it's ok, I think this entire module should be moved to the tests directory but I'll do that in a later change. It can stay in here.

src/lib.rs
jaredwolff: Fixed this

src/lib.rs
jaredwolff: Logic or'ing some bits is easy. Getting the format correctly has been proving to be troublesome. I don't have another solution for this unfortunately.

jaredwolff: Gave this another whirl. I think I got it. Force pushed the changes.

src/lib.rs
jaredwolff: I added another test to test coverage with bytes > 8.

src/lib.rs
```diff
@@ -765,6 +770,63 @@ impl CertificateParams {
...
+									}
+								}
+
+								let mut nb = 1;
+								if msb > 7 {
+									nb = 2;
+								}
+
+								let mut bits = bits.to_le_bytes();
+								bits[0] = bits[0].reverse_bits();
```
est31, line 814: Can't this be done with `bits.to_be_bytes` instead?

est31 (changes requested):
Ok except for nit!

src/lib.rs
jaredwolff: What's the reasoning for using BE over LE? Most architectures are LE, no?

est31: Mhh right the reverse_bits is actually needed because it is about the *bit* order and not the *byte* order. However, if you swap order of the operations it is shorter. The code `bits.reverse_bits().to_be_bytes()` is equivalent and only requires one line. It's not harder to understand it I think. I'll do it in a later change.

est31 (approved):
Thanks!

src/lib.rs
jaredwolff: Ahh yea good point. Thanks @est31

PR #61, opened 2021-08-10
Use public key value as a default serial number

jpastuszek:
Firefox will complain if the same serial number (defaults to 42) is used with different certificate. This happens if you regenerate the certificate. This patch uses value of the public key (first 8 bytes) as the value for the serial number. This way every certificate will get unique serial number.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -722,7 +722,10 @@ impl CertificateParams {
 				writer.write_u8(2);
 			});
 			// Write serialNumber
-			let serial = self.serial_number.unwrap_or(42);
+			let serial = self.serial_number.unwrap_or_else(|| {
+				let bytes: [u8; 8] = pub_key.raw_bytes()[0..8].try_into().unwrap();
+				u64::from_le_bytes(bytes)
+			});
 			writer.next().write_u64(serial);
 			// Write signature
 			ca.params.alg.write_alg_ident(writer.next());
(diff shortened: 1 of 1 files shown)


est31:
Oh that's unfortunate. I'm not sure if the first 8 bytes of public keys are different enough from each other. Maybe use the first 8 bytes of the sha256 of the public key instead? Also, could you please use tabs for indentation? This project uses tabs.

est31:
Lastly, I'd love a comment explaining the situation.

jpastuszek:
OK, I have fixed the tabs.

jpastuszek:
I have written simple program (https://github.com/jpastuszek/post-me) that sets up a web server and prints QR code with local address of that server. You scan it with your phone and point a browser to that address. You can then put data in the form and post it to the server - it will be printed to standard output.
So in order to make it encrypted I wanted to add TLS. With self-signed certs I would get encryption but no protection for man-in-the-middle attacks which should be acceptable for local networks. 

I want the program to be run with zero setup and with no state stored on the device so it generates new certificate at start.

The problem is that if you accept a self signed certificate in Firefox (I did not test with Chrome) it will remember it. So next time you start the post-me program Firefox fill complain that it already has a certificate remembered for this given domain (local network IP in this case) and serial number, but it has different public key (as it was freshly generated) and won't let me in (no way to go past this). So the serial number needs to be unique for each generated certificate for it to work.

I think that this is a general problem as the serial number should be always unique; sure we can't guarantee that with just 64bits by choosing a random number but as a default it should be much better than using one number (42) as you only get 1/2^64 chance of getting duplicate on second run vs 100%.

jpastuszek:
I have added the hashing of the whole public key to ensure good randomness.

est31 (approved):
Thanks!

PR #70, opened 2022-01-04
Implement TryForm<Vec<u8>> for KeyPair struct

prx0:
Hello!

I've implemented TryForm<Vec<u8>> trait for KeyPair to be able to call try_from on a Vec<u8>, avoid to create a new Vec<u8> from a slice of u8.
I also refactor the logic the get the kind and algorithm used for the serialized DER for TryForm trait.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1448,6 +1448,28 @@ impl KeyPair {
 			serialized_der : pkcs8_vec,
 		})
 	}
+
+    fn from_raw(pkcs8: &[u8]) -> Result<(KeyPairKind, &'static SignatureAlgorithm), RcgenError> {
+        Ed25519KeyPair::from_pkcs8_maybe_unchecked(pkcs8)
+            .map(|edkp| (KeyPairKind::Ed(edkp), &PKCS_ED25519))
+            .or_else(|_| {
+                EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P256_SHA256_ASN1_SIGNING, pkcs8)
+                    .map(|eckp| (KeyPairKind::Ec(eckp), &PKCS_ECDSA_P256_SHA256))
+            })
+            .or_else(|_| {
+                EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P384_SHA384_ASN1_SIGNING, pkcs8)
+                    .map(|eckp| (KeyPairKind::Ec(eckp), &PKCS_ECDSA_P384_SHA384))
+            })
+            .or_else(|_| {
+                RsaKeyPair::from_pkcs8(pkcs8).map(|rsakp| {
+                    (
+                        KeyPairKind::Rsa(rsakp, &signature::RSA_PKCS1_SHA256),
+                        &PKCS_RSA_SHA256,
+                    )
+                })
+            })
+            .map_err(|_| RcgenError::CouldNotParseKeyPair)
+    }
 }
 
 /// A private key that is not directly accessible, but can be used to sign messages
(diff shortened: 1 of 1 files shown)


est31 (changes requested):
The `TryFrom` impl for the Vec should just call as_slice and then rely on the impl for slices. And I don't like the refactor.

prx0:
Do you prefer something like that ? 

```rust
impl TryFrom<&[u8]> for KeyPair {
        type Error = RcgenError;
        fn try_from(pkcs8 :&[u8]) -> Result<KeyPair, RcgenError> {
                // using many if else or just map, as you want
                let (kind, alg) = if let Ok(edkp) = Ed25519KeyPair::from_pkcs8_maybe_unchecked(pkcs8) {
                        (KeyPairKind::Ed(edkp), &PKCS_ED25519)
                } else if let Ok(eckp) = EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P256_SHA256_ASN1_SIGNING, pkcs8) {
                        (KeyPairKind::Ec(eckp), &PKCS_ECDSA_P256_SHA256)
                } else if let Ok(eckp) = EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P384_SHA384_ASN1_SIGNING, pkcs8) {
                        (KeyPairKind::Ec(eckp), &PKCS_ECDSA_P384_SHA384)
                } else if let Ok(rsakp) = RsaKeyPair::from_pkcs8(pkcs8) {
                        (KeyPairKind::Rsa(rsakp, &signature::RSA_PKCS1_SHA256), &PKCS_RSA_SHA256)
                } else {
                        return Err(RcgenError::CouldNotParseKeyPair);
                };

                Ok(KeyPair {
                        kind,
                        alg,
                        serialized_der : pkcs8.to_vec(), // copies the slice into a new vec
                })
        }
}

impl TryFrom<Vec<u8>> for KeyPair {
        type Error = RcgenError;
        fn try_from(pkcs8: Vec<u8>) -> Result<KeyPair, RcgenError> {
            KeyPair::try_from(pkcs8.as_slice())
        }
}
```

But the [to_vec](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.to_vec) method copies the slice into a new vector in memory. I tried to avoid to create a new vector.

est31:
> But the to_vec method copies the slice into a new vector in memory. I tried to avoid to create a new vector.

Oh right good point. Then it should be the slice version that calls the Vec version. All I want it so save on duplicate code :).

Stargateur:
> > But the to_vec method copies the slice into a new vector in memory. I tried to avoid to create a new vector.
> 
> Oh right good point. Then it should be the slice version that calls the Vec version. All I want it so save on duplicate code :).

Problem is if `slice` version call `Vec<u8>` this will allocate a vector even if the verification fail, it's not duplicate code cause it's have a light behavior difference. The refactor prevent mistake at compile time so it's should be ok.

src/lib.rs
```diff
@@ -1556,28 +1578,27 @@ impl From<pem::PemError> for RcgenError {
...
-
-		Ok(KeyPair {
-			kind,
-			alg,
-			serialized_der : pkcs8_vec,
-		})
-	}
+    type Error = RcgenError;
+    fn try_from(pkcs8: &[u8]) -> Result<KeyPair, RcgenError> {
+        let (kind, alg) = KeyPair::from_raw(pkcs8).expect("unable to detect the kind and algorithm");
```
Stargateur, line 1583: `let (kind, alg) = KeyPair::from_raw(pkcs8)?;`

src/lib.rs
```diff
@@ -1556,28 +1578,27 @@ impl From<pem::PemError> for RcgenError {
...
+            alg,
+            serialized_der: pkcs8.to_vec(),
+        })
+    }
+}
+
+impl TryFrom<Vec<u8>> for KeyPair {
+    type Error = RcgenError;
+    fn try_from(pkcs8: Vec<u8>) -> Result<KeyPair, RcgenError> {
+        let (kind, alg) = KeyPair::from_raw(pkcs8.as_slice()).expect("unable to detect kind and algorithm");
```
Stargateur, line 1595: `let (kind, alg) = KeyPair::from_raw(pkcs8)?;`

Stargateur (changes requested):

src/lib.rs
prx0: Thx ! Fixed by https://github.com/est31/rcgen/pull/70/commits/680317503027feccd87e277b1c3f2c2f628ba568

src/lib.rs
prx0: Thx ! Fixed by https://github.com/est31/rcgen/pull/70/commits/680317503027feccd87e277b1c3f2c2f628ba568

est31:
> Problem is if slice version call Vec<u8> this will allocate a vector even if the verification fail, it's not duplicate code cause it's have a light behavior difference. 

Good point, I think a separate method is good. I still prefer the old if let chain as it is more compact. Also, the PR uses spaces for indentation while this project uses tabs. Would it be possible to adjust to this project's code style?

prx0:
> I still prefer the old if let chain as it is more compact. Also, the PR uses spaces for indentation while this project uses tabs. Would it be possible to adjust to this project's code style?

No problem, I changed the code to use tabs instead of spaces and it use the if let chain as you requested.

est31 (approved):
Thanks!

Issue #65, opened 2021-10-25
Dealing with unsoundness in chrono


djc:
Hey, have you decided what you want to do with the chrono issues?

https://rustsec.org/advisories/RUSTSEC-2020-0159

Maybe it makes sense for rcgen to just adopt time 0.3 instead?

makr11st:
@est31 Hi, is there an update on #66 it seems the PRs in deps are now merged.

est31:
The time crate does not support leap seconds, see https://github.com/time-rs/time/issues/193 . This can be an issue for certificate generation libraries as single seconds [can become important](https://community.letsencrypt.org/t/2021-06-08-certificate-lifetime-incident-valid-for-an-extra-one-second/153426) in the security domain. So I'm a bit reluctant, because rcgen does not use the offending function that makes chrono unsafe in the database, but it does rely on leap second support. The PR #66 has to change some of the unit tests that check for leap seconds.

Ultimately, if there is no movement on the time crate, I think I'll just merge it as is and accept the regression.

est31:
Alright, when the new ring version will be released, I'll switch to time: https://github.com/briansmith/ring/issues/1416#issuecomment-989381344

makr11st:
No worries and no pushing, just was wondering if there were additional dependencies you were waiting for, it makes sense to implement proper leap seconds handling. I saw the leap second discussion in the other repos, but it doesn't seem like a straightforward problem to solve.
Thanks for replying.

est31:
Hmmm no changes for ring. There seems to be progress on the chrono front though: https://github.com/chronotope/chrono/pull/632#issuecomment-1002788352

est31:
Also: https://github.com/chronotope/chrono/pull/639

makr11st:
Thanks for the merge and release.

PR #71, opened 2022-01-16
feat: support non utf8 strings

### LINKED ISSUE #59: Loading CA cert and using for signing doesn't match original certificate
I'm trying to develop a P2P application where clients can register with a server, which will sign their certificates, so that they can later talk to each other without the involvement of the server.

I've been trying to load the CA cert using `CertificateParams::from_ca_cert_pem` and sign the client certs with `Certificate::serialize_der_with_signer`.  However the "issuer" attached to the resulting signed cert doesn't match the original cert (specifically the "subject" is 1 byte different).

I don't really know enough about TLS to figure out why this is happening, so let me know if I can provide any additional details.

My (testing) code looks like this:

```rust
    let ca_key = rcgen::KeyPair::from_der(debug_certs::CA_KEY_PK8).unwrap();
    let params = rcgen::CertificateParams::from_ca_cert_pem(
        debug_certs::CA_CERT,
        ca_key,
    )
    .unwrap();
    let ca_cert = rcgen::Certificate::from_params(params).unwrap();
    let gen_cert = rcgen::generate_simple_self_signed(vec![hostname]).unwrap();
    gen_cert.serialize_der_with_signer(&ca_cert).unwrap();
```

The issuer "subject" lines I get during certificate verification look like this (as byte arrays, note byte index 9)

Certificate "Issuer":
```
[49, 11, 48, 9, 6, 3, 85, 4, 6, 12, 2, 85, 83, 49, 17, 48, 15, 6, 3, 85, 4, 8, 12, 8, 86, 105, 114, 103, 105, 110, 105, 97, 49, 33, 48, 31, 6, 3, 85, 4, 10, 12, 24, 73, 110, 116, 101, 114, 110, 101, 116, 32, 87, 105, 100, 103, 105, 116, 115, 32, 80, 116, 121, 32, 76, 116, 100]
```
Trust Anchor "Subject":
```
[49, 11, 48, 9, 6, 3, 85, 4, 6, 19, 2, 85, 83, 49, 17, 48, 15, 6, 3, 85, 4, 8, 12, 8, 86, 105, 114, 103, 105, 110, 105, 97, 49, 33, 48, 31, 6, 3, 85, 4, 10, 12, 24, 73, 110, 116, 101, 114, 110, 101, 116, 32, 87, 105, 100, 103, 105, 116, 115, 32, 80, 116, 121, 32, 76, 116, 100]
```

[debug-game-cert.pem.txt](https://github.com/est31/rcgen/files/6644906/debug-game-cert.pem.txt)
[example_signed_cert.pem.txt](https://github.com/est31/rcgen/files/6644909/example_signed_cert.pem.txt)

omjadas:
fixes #59
Allows non UTF-8 values to be used for distinguished name entries

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@ use ring::signature::KeyPair as RingKeyPair;
 use ring::signature::{self, EcdsaSigningAlgorithm, EdDSAParameters};
 use yasna::DERWriter;
 use yasna::models::{GeneralizedTime, UTCTime};
+use yasna::tags::{TAG_BMPSTRING, TAG_TELETEXSTRING, TAG_UNIVERSALSTRING};
 use chrono::{DateTime, Timelike, Datelike};
 use chrono::{NaiveDate, Utc};
 use std::collections::HashMap;
@@ -370,6 +371,31 @@ impl DnType {
 	}
 }
 
+/// A distinguished name entry
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+#[non_exhaustive]
+pub enum DnValue {
+	/// A string of characters from the T.61 character set
+	TeletexString(Vec<u8>),
+	/// An ASCII string containing only A-Z, a-z, 0-9, '()+,-./:=? and <SPACE>
+	PrintableString(String),
+	/// A string encoded using UTF-32
+	UniversalString(Vec<u8>),
+	/// A string encoded using UTF-8
+	Utf8String(String),
+	/// A string encoded using UCS-2
+	BmpString(Vec<u8>),
+}
+
+impl<T> From<T> for DnValue
+where
+	T :Into<String>
+{
+	fn from(t :T) -> Self {
+		DnValue::Utf8String(t.into())
+	}
+}
+
 #[derive(Debug, PartialEq, Eq, Clone)]
 /**
 Distinguished name used e.g. for the issuer and subject fields of a certificate
@@ -382,7 +408,7 @@ See also the RFC 5280 sections on the [issuer](https://tools.ietf.org/html/rfc52
 and [subject](https://tools.ietf.org/html/rfc5280#section-4.1.2.6) fields.
 */
 pub struct DistinguishedName {
-	entries :HashMap<DnType, String>,
+	entries :HashMap<DnType, DnValue>,
 	order :Vec<DnType>,
 }
 
@@ -395,11 +421,8 @@ impl DistinguishedName {
 		}
 	}
 	/// Obtains the attribute value for the given attribute type
-	pub fn get(&self, ty :&DnType) -> Option<&str> {
-		self.entries.get(ty).map(|s| {
-			let s :&str = s;
-			s
-		})
+	pub fn get(&self, ty :&DnType) -> Option<&DnValue> {
+		self.entries.get(ty)
 	}
 	/// Removes the attribute with the specified DnType
 	///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -414,11 +427,11 @@ impl DistinguishedName {
 		removed
 	}
 	/// Inserts or updates an attribute that consists of type and name
-	pub fn push(&mut self, ty :DnType, s :impl Into<String>) {
+	pub fn push(&mut self, ty :DnType, s :DnValue) {
```
est31, line 430: I'd prefer to keep the old API for the push side, as 99% of use cases would probably be UTF-8 strings. This doesn't just reduce the diff of this PR but also makes updating easier for the users. Usage of the API would also remain as easy as it used to be. Can you rename this function to `push_ext` or something and say something along the lines that you should use `push` unless there is a specific need to use something other than `Utf8Strings`?

The `get` side is fine, as I think it's less used.

est31 (changes requested):
Thanks for the PR! This is an API breaking change, so requires a minor semver bump for a future release. Not sure yet if I want to wait with merging or merge it right now.

src/lib.rs
omjadas: Would it be better to have a single method that takes an `impl Into<DnValue>` and provide the following implementation?

```rust
impl<T> From<T> for DnValue
where
	T :Into<String>
{
	fn from(t :T) -> Self {
		DnValue::Utf8String(t.into())
	}
}
```

omjadas:
@est31 I have updated the PR per your suggestions. Let me know if you would prefer `push` to be split into separate methods and I can make the change.

src/lib.rs
est31: > Would it be better to have a single method that takes an impl Into<DnValue> and provide the following implementation?

That's a great idea! I'm a bit worried though that the impl is not immediately seen. A code example in the rustdoc that uses a str literal would solve that I think.

est31:
Also I'd like to see a test case for #59. Ideally it'd not use a certificate from the wild due to copyright but one you generated yourself (maybe even using rcgen). As for the breaking change, one of the possible resolutions of #65 seems to require one already, and the ring author said they wanted to make a breaking change themselves soon.

omjadas:
@est31 I have added the requested example and tests. Let me know if there is anything else you would like changed/added.

src/lib.rs
```diff
@@ -414,7 +437,16 @@ impl DistinguishedName {
 		removed
 	}
 	/// Inserts or updates an attribute that consists of type and name
-	pub fn push(&mut self, ty :DnType, s :impl Into<String>) {
+	///
+    /// ```
```
est31, line 441: nit: formatting

src/lib.rs
```diff
@@ -414,7 +437,16 @@ impl DistinguishedName {
...
+	/// # use rcgen::{DistinguishedName, DnType, DnValue};
+	/// let mut dn = DistinguishedName::new();
+	/// dn.push(DnType::OrganizationName, "Crab widgits SE");
+    /// dn.push(DnType::CommonName, DnValue::PrintableString("Master Cert".to_string()));
```
est31, line 445: nit: formatting

est31 (changes requested):
OK except for nits.

src/lib.rs
omjadas: fixed

src/lib.rs
omjadas: fixed

PR #72, opened 2022-01-21
chore: use ia5_string for Rfc822Name and DnsName

omjadas:
According to [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280#page-38) DNS Name and RFC822 Name are IA5 strings.

```
GeneralName ::= CHOICE {
     otherName                       [0]     OtherName,
     rfc822Name                      [1]     IA5String,
     dNSName                         [2]     IA5String,
     x400Address                     [3]     ORAddress,
     directoryName                   [4]     Name,
     ediPartyName                    [5]     EDIPartyName,
     uniformResourceIdentifier       [6]     IA5String,
     iPAddress                       [7]     OCTET STRING,
     registeredID                    [8]     OBJECT IDENTIFIER }
```

`write_ia5_string` panics if the provided string is not ASCII, perhaps we should add our own `is_ascii` check and return an error instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -719,7 +719,7 @@ impl CertificateParams {
 					writer.next().write_tagged_implicit(Tag::context(san.tag()), |writer| {
 						match san {
 							SanType::Rfc822Name(name) |
-							SanType::DnsName(name) => writer.write_utf8_string(name),
+							SanType::DnsName(name) => writer.write_ia5_string(name),
 							SanType::IpAddress(IpAddr::V4(addr)) => writer.write_bytes(&addr.octets()),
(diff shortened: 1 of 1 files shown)


est31:
Just to know the motivation, have you ran into software not accepting rcgen generated certificates due to this issue? Or did you just compare the spec to what rcgen is doing?

omjadas:
I noticed this when reading through the spec and comparing to rcgen. From some quick research, it looks like Go rejects certificates with malformed SANs. There are probably other implementations that would also reject them.

https://github.com/golang/go/blob/b7fa0f941f05fdf8420f15cd5ebe2f209da172e0/src/crypto/x509/parser.go#L379

PR #76, opened 2022-02-09
Check for 1.53 support

### LINKED ISSUE #75: Can we lower the MSRV?
0.9 raised the MSRV to 1.56 by virtue of setting the edition to 2021. IMO it would be nice to avoid that for a while longer; for Quinn, for example, we have a policy of sticking to 6-month old versions for the MSRV.

djc:
Fixes #75.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [macOS-latest, ubuntu-latest]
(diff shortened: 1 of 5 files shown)


djc:
Would appreciate a release after this is merged.

est31 (changes requested):
Can you add a changelog entry?

djc:
Done.

est31 (approved):
Thanks!

PR #66, opened 2021-11-01
Switch from `chrono` to `time`

connec:
⚠️ This currently includes `yasna.rs` as a git dependency, because there's no release with https://github.com/qnighy/yasna.rs/pull/61 yet. I wanted to open the PR anyway to indicate the changes, and perhaps motivate a release for `yasna.rs`.

---

Since https://github.com/qnighy/yasna.rs/pull/61 is merged, I wanted to make the corresponding updates here as well, to hopefully resolve #65.

The changes are very mechanically once again, and again the main limitation is that `time` cannot represent leap seconds. This affects some of the logic and removes a test case.

I've checked the output from `main.rs` and the examples with `openssl x509 -in certs/cert.pem -noout -text` and it's as I'd expect. I've also tried it with a library using QUIC w/ self-signed certs and that also works as expected.

PR #77, opened 2022-02-21
Update x509-parser to 0.13

matze:
Despite #66, chrono is still pulled in transitively via x509-parser 0.12. x509-parser 0.13 [switched to time as well](https://github.com/rusticata/x509-parser/blob/master/CHANGELOG.md#0130), so using rcgen with an updated version won't cause `cargo audit` issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,45 @@
 version = 3
 
 [[package]]
+name = "asn1-rs"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


est31:
@matze do you want a crates.io release?

matze:
That would be splendid! This is the last piece missing in our corporate project that requires us to make an exception for `cargo audit`.

est31:
[0.9.2](https://github.com/est31/rcgen/releases/tag/v0.9.2) is out.

Issue #83, opened 2022-04-13
Raw bytes expected instead for RemoteKeyPair#public_key


nelson-ks:
It is stated in the doc that the public_key should returned in DER format
https://docs.rs/rcgen/0.9.2/rcgen/trait.RemoteKeyPair.html#tymethod.public_key

However it should be the raw bytes of the public key as this is what provided from ring
https://docs.rs/rcgen/0.9.2/src/rcgen/lib.rs.html#1781
I can't find in the docs of thering library but here it says the output from ring is in raw format
https://docs.rs/rcgen/0.9.2/rcgen/struct.KeyPair.html#method.public_key_raw

est31:
I guess this was an oversight of 261aded17b4979b0d56ca1dafade1889837d3eb3 ... It's good that I made the attempt to document the format at all though, as it shows that people are interested in it. PRs to fix welcome!

nelson-ks:
Ah I see. 
I thought it expects some DER identifier when it mentions DER format.

I have make a PR. #84  Please feel free to comment.
And thank you for your work. It is very helpful

est31:
Fixed by #84

PR #84, opened 2022-04-13
Update RemoteKeyPair doc

nelson-ks:
Use "raw binary format" instead of "DER format" for clarity as mentioned in #83


est31:
Can you link to the `KeyPair::public_key_raw` function and say it's the same format as exported there? "raw" on its own is a bit ambiguous.

est31:
intra-doc-links work like this: https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html

nelson-ks:
No problems. Yeah this is better.

est31 (approved):
Thanks!

PR #86, opened 2022-07-15
Explicitly display CA:FALSE

doraneko94:
When generating a certificate, I think it's common practice to specify `Basic Constraints` as `CA:FALSE` if it's not a CA.
Therefore, I've modified `lib.rs` so that `Basic Constraints` are written to the certificate even if `SelfSignedOnly` is specified for `isCA` of `CertificateParams`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -911,22 +911,38 @@ impl CertificateParams {
 								});
 							}
 						}
-						if let IsCa::Ca(ref constraint) = self.is_ca {
-							// Write subject_key_identifier
-							Self::write_extension(writer.next(), OID_SUBJECT_KEY_IDENTIFIER, false, |writer| {
-								let key_identifier = self.key_identifier(pub_key);
-								writer.write_bytes(key_identifier.as_ref());
-							});
-							// Write basic_constraints
-							Self::write_extension(writer.next(), OID_BASIC_CONSTRAINTS, true, |writer| {
-								writer.write_sequence(|writer| {
-									writer.next().write_bool(true); // cA flag
-									if let BasicConstraints::Constrained(path_len_constraint) = constraint {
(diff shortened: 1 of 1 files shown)


est31:
[RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280) wants us to have the Subject Key Identifier extension with a `SHOULD` requirement:

```
   Where an end entity
   has obtained multiple certificates, especially from multiple CAs, the
   subject key identifier provides a means to quickly identify the set
   of certificates containing a particular public key.  To assist
   applications in identifying the appropriate end entity certificate,
   this extension SHOULD be included in all end entity certificates.
```

And the Basic Constraints extension with a `MAY` requirement:

```
   This extension MAY appear as a critical or
   non-critical extension in end entity certificates.
```

Generally though I want this library to generate certificates that are as small as possible so I'm not sure I want this. I'm open to adding an option for it though. Also now I'm wondering if `SelfSignedOnly` is a good name because it is also meant for end entity certificates signed by a proper CA.

doraneko94:
I think the name `SelfSignedOnly ` is the best if there is a clear reason.
So how about adding a new `IsCA` element to give `CA:FALSE`, 
or to give `CA:FALSE` if `IsCA::Ca(BasicConstraint::Constrained(0))` is specified?
If the former method is adopted, what do you think is the best name?

est31:
Maybe one could have `IsCA::NoCa` and `IsCa::ExplicitNoCa`, where the current behaviour for `SelfSignedOnly` becomes `NoCa`, and `ExplicitNoCa` adds the extension and `CA:FALSE`?

doraneko94:
I agree with you. Please check new PR.

est31 (approved):
Thanks!

PR #87, opened 2022-07-15
Add KeyPair::serialized_der method

jean-airoldie:
Allows the user to access the DER encoded keypair's bytes without
cloning.

My specific use case is that I'm implementing my own `KeyPair<'_>` type that I use for signing, and without this method I would also need a `OwnedKeyPair` version.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1768,6 +1768,19 @@ impl KeyPair {
 
 		self.serialized_der.clone()
 	}
+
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks!

est31:
> If you merge this, I would love for a new release if possible.

https://github.com/est31/rcgen/releases/tag/v0.9.3

PR #93, opened 2022-10-06
Parse an IP address into SanType from a x509 GeneralName.

iamjpotts:
Support parsing certificates that have an IP address as a subject alternate name.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,6 +49,8 @@ use std::fmt;
 use std::convert::TryFrom;
 use std::error::Error;
 use std::net::IpAddr;
+#[cfg(feature = "x509-parser")]
+use std::net::{Ipv4Addr, Ipv6Addr};
 use std::str::FromStr;
 use std::hash::{Hash, Hasher};
 
@@ -159,6 +161,19 @@ pub enum SanType {
 	IpAddress(IpAddr),
 }
 
+#[cfg(feature = "x509-parser")]
+fn ip_addr_from_octets(octets: &[u8]) -> Result<IpAddr, RcgenError> {
+	if let Ok(ipv6_octets) = <&[u8; 16]>::try_from(octets) {
+		Ok(Ipv6Addr::from(*ipv6_octets).into())
+	}
+	else if let Ok(ipv4_octets) = <&[u8; 4]>::try_from(octets) {
+		Ok(Ipv4Addr::from(*ipv4_octets).into())
+	}
+	else {
+		Err(RcgenError::InvalidIpAddressOctetLength(octets.len()))
+	}
+}
+
 impl SanType {
 	#[cfg(feature = "x509-parser")]
 	fn try_from_general(name :&x509_parser::extensions::GeneralName<'_>) -> Result<Self, RcgenError> {
@@ -171,6 +186,9 @@ impl SanType {
 			}
 			x509_parser::extensions::GeneralName::URI(name) => {
 				SanType::URI((*name).into())
+			}
+			x509_parser::extensions::GeneralName::IPAddress(octets) => {
+				SanType::IpAddress(ip_addr_from_octets(octets)?)
 			}
 			_ => return Err(RcgenError::InvalidNameType),
 		})
@@ -1585,6 +1603,8 @@ pub enum RcgenError {
 	#[cfg(feature = "x509-parser")]
 	/// Invalid subject alternative name type
 	InvalidNameType,
+	/// An IP address was provided as a byte array, but the byte array was an invalid length.
+	InvalidIpAddressOctetLength(usize),
 	/// There is no support for generating
 	/// keys for the given algorithm
 	KeyGenerationUnavailable,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -159,6 +161,47 @@ pub enum SanType {
...
+	let ip = match len {
+		4 => Ipv4Addr::new(
+			octets[0],
+			octets[1],
+			octets[2],
+			octets[3]
+		)
+			.into(),
+		16 => {
+			let o: [u8; 16] = [
```
est31, line 177: Can you use `= octets.try_into().unwrap()`? This should do it I think. The unwrap is safe because the size is known.

src/lib.rs
```diff
@@ -159,6 +161,47 @@ pub enum SanType {
 	IpAddress(IpAddr),
 }
 
+#[cfg(feature = "x509-parser")]
+fn ip_addr_from_octets(octets: &[u8]) -> Result<IpAddr, RcgenError> {
+	let len = octets.len();
+
+	let ip = match len {
+		4 => Ipv4Addr::new(
```
est31, line 169: same here

src/lib.rs
```diff
@@ -2249,4 +2298,111 @@ mod tests {
...
+			let expected = IpAddr::from([10, 20, 30, 40]);
+
+			assert_eq!(expected, actual)
+		}
+
+		#[test]
+		fn ipv6() {
+			let actual = ip_addr_from_octets(
+				&[
+					1,
```
est31, line 2322: This formatting is not great at all. Can it be on one line? Also if you have the same slice then it might make sense to just put it into a variable? Lastly, if it's increasing numbers you could probably do it like in the `too_many` test, via collecting a range.

src/lib.rs
```diff
@@ -2249,4 +2298,111 @@ mod tests {
...
+					14,
+					15,
+					16
+				]
+			)
+				.unwrap();
+
+			let expected = IpAddr::from(
+				[
+					1,
```
est31, line 2344: same here

est31 (changes requested):

src/lib.rs
```diff
@@ -1621,6 +1669,7 @@ impl fmt::Display for RcgenError {
 			CouldNotParseKeyPair => write!(f, "Could not parse key pair")?,
 			#[cfg(feature = "x509-parser")]
 			InvalidNameType => write!(f, "Invalid subject alternative name type")?,
+			InvalidIpAddressOctetLength(actual) => write!(f, "Invalid IP address octet length of {} bytes", actual)?,
```
est31, line 1672: this can use the `{actual}` syntax.

src/lib.rs
iamjpotts: Doesn't compile:

`the trait From<&[u8]> is not implemented for Ipv4Addr`

iamjpotts:
I was surprised that `IpAddr` doesn't have any method that accepts a slice, array reference, or vector to return `Result<IpAddr>`

What are your thoughts on creating an `imp` module with `pub(crate)` visibility and putting the IP address conversion code and tests in it?

Seems like it clutters up `lib.rs` but I didn't want to introduce too much change at once.

src/lib.rs
iamjpotts: There's a `nightly` api that may help with this in the future:

https://doc.rust-lang.org/std/primitive.slice.html#method.split_array_ref

src/lib.rs
iamjpotts: Creating the input with a range and collecting it to a vector won't compile, because `IpAddr::from` requires a fixed-length array.

I can deduplicate the octets though and make it one line.

src/lib.rs
est31: Yeah you'd have to first convert the slice into an array reference, and then convert that array reference into an array, and *then* you would have to convert that into an `Ipv4Addr`. No nightly features needed, everything stable since over a year ago.

iamjpotts: lol, that stretched my rust legs, but I figured out how to do it and updated the PR.

est31 (changes requested):
This is almost good now!

src/lib.rs
iamjpotts: Done.

iamjpotts:
> This is almost good now!

I see its marked `changes requested` but I didn't see any more comments.

Any further feedback on the last update?

src/lib.rs
```diff
@@ -159,6 +161,25 @@ pub enum SanType {
 	IpAddress(IpAddr),
 }
 
+#[cfg(feature = "x509-parser")]
+fn ip_addr_from_octets(octets: &[u8]) -> Result<IpAddr, RcgenError> {
+	match <&[u8; 16]>::try_from(octets) {
```
est31, line 166: Can you put back the old matching on the length of octets? That needs an unwrap but it seems nicer to me than this code.

est31:
> I see its marked changes requested but I didn't see any more comments.

Oh sorry, not sure what happened, it seems to have gotten lost. I've re-added the comment.

src/lib.rs
iamjpotts: I can do that but some people take issue with `unwrap` in library (non-test, non-example) code.

Another way besides `match` you may like better would be `if let`, for example

```
	if let Ok(ipv6_octets) = <&[u8; 16]>::try_from(octets) {
		Ok(Ipv6Addr::from(*ipv6_octets).into())
	}
	else if let Ok(ipv4_octets) = <&[u8; 4]>::try_from(octets) {
		Ok(Ipv4Addr::from(*ipv4_octets).into())
	}
	else {
		Err(RcgenError::InvalidIpAddressOctetLength(octets.len()))
	}
```

Thoughts on the `if let` variant?

est31: I don't mind unwraps in this location, but if you prefer if let that's also fine. Both code looks nice, but a nested match just isn't beautiful.

iamjpotts: `if let` pushed

est31 (approved):
Thanks!

PR #94, opened 2022-10-11
Fix typeo in rsa-irc

iamjpotts:


est31:
Merged as part of #93

Issue #95, opened 2022-10-18
CA certificate parsed as NoCa


andrenth:
Hello

I'm using rcgen to create a CA certificate and then parse its parameters back. The issue is that I create the certificate with `params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);`, but when parsing it, I get `IsCa::NoCa` in the `is_ca` field:

```rust
use std::fs;

use rcgen::{
    BasicConstraints, Certificate, CertificateParams, DistinguishedName, DnType,
    ExtendedKeyUsagePurpose, IsCa, KeyPair, KeyUsagePurpose, PKCS_ECDSA_P384_SHA384,
};
use time::{Duration, OffsetDateTime};

fn create_ca_cert() -> Result<(), Box<dyn std::error::Error>> {
    let now = OffsetDateTime::now_utc();
    let mut params = CertificateParams::default();
    let organization = "MyOrg".to_owned();

    params.alg = &PKCS_ECDSA_P384_SHA384;
    params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);

    params.not_before = now;
    params.not_after = now + Duration::days(3650);

    params.key_usages.push(KeyUsagePurpose::KeyCertSign);
    params.key_usages.push(KeyUsagePurpose::DigitalSignature);

    params
        .extended_key_usages
        .push(ExtendedKeyUsagePurpose::ClientAuth);
    params
        .extended_key_usages
        .push(ExtendedKeyUsagePurpose::ServerAuth);

    let mut distinguished_name = DistinguishedName::new();
    distinguished_name.push(DnType::OrganizationName, &organization);
    distinguished_name.push(DnType::OrganizationalUnitName, "MyOrgUnit");
    params.distinguished_name = distinguished_name;

    let certificate = Certificate::from_params(params)?;
    fs::write("ca.pem", certificate.serialize_pem()?)?;
    fs::write("ca.key", certificate.serialize_private_key_pem())?;

    Ok(())
}

fn parse_ca_cert() -> Result<IsCa, Box<dyn std::error::Error>> {
    let crt = fs::read_to_string("ca.pem")?;
    let key = fs::read_to_string("ca.key")?;

    let key_pair = KeyPair::from_pem(&key)?;
    let params = CertificateParams::from_ca_cert_pem(&crt, key_pair)?;
    Ok(params.is_ca)
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    create_ca_cert()?;
[...]

PR #96, opened 2022-10-20
Parse more certificate parameters

andrenth:
Parse most fields of `CertificateParams` from `X509Certificate`. Addresses #95.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -729,7 +729,7 @@ impl CertificateParams {
 		let dn = DistinguishedName::from_name(&x509.tbs_certificate.subject)?;
 		let is_ca = Self::convert_x509_is_ca(&x509)?;
 		let validity = x509.validity();
-		let serial = Self::convert_x509_serial(&x509);
+		let serial = Self::convert_x509_serial(&x509)?;
 		let subject_alt_names = Self::convert_x509_subject_alternative_name(&x509)?;
 		let key_usages = Self::convert_x509_key_usages(&x509)?;
 		let extended_key_usages = Self::convert_x509_extended_key_usages(&x509)?;
@@ -770,19 +770,19 @@ impl CertificateParams {
 		Ok(is_ca)
 	}
 	#[cfg(feature = "x509-parser")]
-	fn convert_x509_serial(x509 :&x509_parser::certificate::X509Certificate<'_>) -> u64 {
+	fn convert_x509_serial(x509 :&x509_parser::certificate::X509Certificate<'_>) -> Result<u64, RcgenError> {
 		const SIZE: usize = std::mem::size_of::<u64>();
 
 		let raw = x509.raw_serial();
 		let len = raw.len();
 
-		if len >= SIZE {
-			return u64::from_be_bytes(raw[(len - SIZE)..].try_into().unwrap())
+		if len > SIZE {
+			return Err(RcgenError::CouldNotParseCertificate);
 		}
 
 		let mut bytes = [0u8; SIZE];
 		bytes[(SIZE - len)..].copy_from_slice(raw);
-		u64::from_be_bytes(bytes)
+		Ok(u64::from_be_bytes(bytes))
 	}
 	#[cfg(feature = "x509-parser")]
 	fn convert_x509_subject_alternative_name(x509 :&x509_parser::certificate::X509Certificate<'_>) -> Result<Vec<SanType>, RcgenError> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -727,14 +727,218 @@ impl CertificateParams {
 		let alg = SignatureAlgorithm::from_oid(&alg_oid.collect::<Vec<_>>())?;
 
 		let dn = DistinguishedName::from_name(&x509.tbs_certificate.subject)?;
+		let is_ca = Self::convert_x509_is_ca(&x509)?;
+		let validity = x509.validity();
+		let serial = Self::convert_x509_serial(&x509);
+		let subject_alt_names = Self::convert_x509_subject_alternative_name(&x509)?;
+		let key_usages = Self::convert_x509_key_usages(&x509)?;
+		let extended_key_usages = Self::convert_x509_extended_key_usages(&x509)?;
+		let name_constraints = Self::convert_x509_name_constraints(&x509)?;
+
 		Ok(
 			CertificateParams {
 				alg,
+				is_ca,
+				subject_alt_names,
+				key_usages,
+				extended_key_usages,
+				name_constraints,
 				distinguished_name : dn,
 				key_pair : Some(key_pair),
+				serial_number: Some(serial),
+				not_before : validity.not_before.to_datetime(),
+				not_after : validity.not_after.to_datetime(),
 				.. Default::default()
 			}
 		)
+	}
+	#[cfg(feature = "x509-parser")]
+	fn convert_x509_is_ca(x509 :&x509_parser::certificate::X509Certificate<'_>) -> Result<IsCa, RcgenError> {
+		use x509_parser::extensions::BasicConstraints as B;
+
+		let basic_constraints = x509.basic_constraints()
+			.or(Err(RcgenError::CouldNotParseCertificate))?.map(|ext| ext.value);
+
+		let is_ca = match basic_constraints {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -727,15 +727,219 @@ impl CertificateParams {
...
 	}
+	#[cfg(feature = "x509-parser")]
+	fn convert_x509_is_ca(x509 :&x509_parser::certificate::X509Certificate<'_>) -> Result<IsCa, RcgenError> {
+		use x509_parser::extensions::BasicConstraints as B;
+
+		let basic_constraints = x509.basic_constraints()
+			.or(Err(RcgenError::CouldNotParseCertificate))?.map(|ext| ext.value);
+
+		let is_ca = match basic_constraints {
+			Some(B { ca: true, path_len_constraint: Some(n) }) => IsCa::Ca(BasicConstraints::Constrained(*n as u8)),
```
est31, line 763: This should return some error if n does not fit an u8 I think.

src/lib.rs
```diff
@@ -727,15 +727,219 @@ impl CertificateParams {
...
+	}
+	#[cfg(feature = "x509-parser")]
+	fn convert_x509_serial(x509 :&x509_parser::certificate::X509Certificate<'_>) -> u64 {
+		const SIZE: usize = std::mem::size_of::<u64>();
+
+		let raw = x509.raw_serial();
+		let len = raw.len();
+
+		if len >= SIZE {
+			return u64::from_be_bytes(raw[(len - SIZE)..].try_into().unwrap())
```
est31, line 779: same here, I think an error (or ignoring the serial if it doesnt fit) is better than truncation. In other places we ignore stuff unknown to us, and that is fine, but there is a difference between "i don't know" and "i know what this is but can't fit the data so I truncate it".

src/lib.rs
```diff
@@ -727,15 +727,219 @@ impl CertificateParams {
...
+
+		let mut result = Vec::new();
+		for subtree in subtrees {
+			let subtree = match &subtree.base {
+				GeneralName::RFC822Name(s) => GeneralSubtree::Rfc822Name(s.to_string()),
+				GeneralName::DNSName(s) => GeneralSubtree::DnsName(s.to_string()),
+				GeneralName::DirectoryName(n) => GeneralSubtree::DirectoryName(DistinguishedName::from_name(&n)?),
+				GeneralName::IPAddress(bytes) if bytes.len() == 8 => {
+					let mut addr = [0u8; 4];
+					addr.copy_from_slice(&bytes[..4]);
```
est31, line 925: I think it would be better to use the [`TryFrom` impl](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html#impl-TryFrom%3C%26%27a%20%5BT%5D%3E-for-%26%27a%20%5BT%3B%20N%5D) here together with an unwrap. Same goes for the other conversions in this function.

est31 (changes requested):

src/lib.rs
andrenth: Addressed in c1b0b40. I was unsure if just returning `Unconstrained` instead of an error would be better, but went with the error route.

est31 (approved):

src/lib.rs
andrenth: I've changed it to return an error in f29f839.

src/lib.rs
andrenth: Done in 15bf1f7

src/lib.rs
andrenth: This is now failing tests sometimes because the serial number is too large. According to RFC 5280, a serial number of up to 20 bytes must be handled, but that is currently impossible in rcgen due to the use of `u64` for this field.

Do you think it's ok to change the field type to [BigUint](https://docs.rs/num-bigint/0.4.3/num_bigint/struct.BigUint.html) (or maybe just keep the raw bytes, as `Vec<u8>`, possibly in a new field)?. Otherwise I don't see an option other than not parsing large serial numbers, but that's kinda unsatisfying.

est31: This is a tough question, I would like to keep this library simple but I also understand that we have to support 20 bytes or more. Maybe this could be deferred for later?

andrenth: On the one hand I understand not wanting to introduce a dependency, but on the other I guess it would be easier to break backward compatibility now rather than later.

What do you think about just adding a `raw_serial_number: Option<Vec<u8>>` field to `CertificateParams` for now?

est31: Do you have a specific use case for the serial number stuff? Otherwise I don't think we need to address this right now.

andrenth: I would like to use an UUID that identifies the certificate as its serial number for my use case. It's more of a "would be nice to have" thing though, not strictly necessary, as I can use some other field for that.

est31: Hmm this makes sense. Having a separate field feels like a hack to me. I think it should be the main serial number field, but we should use a new type wrapper for the big number so that we don't depend on that crate's API. Ideally something that contains the `Vec<u8>` directly without needing a crate, then offering conversion methods to/from integers. But this is a large design space. That's why I said I wanted to have a separate PR for it, as the other changes here are fine.

src/lib.rs
```diff
@@ -727,15 +727,216 @@ impl CertificateParams {
...
+
+		if let Some(sans) = sans {
+			let mut subject_alt_names = Vec::with_capacity(sans.len());
+			for san in sans {
+				use x509_parser::extensions::GeneralName;
+				let name = match san {
+					GeneralName::RFC822Name(s) => SanType::Rfc822Name(s.to_string()),
+					GeneralName::DNSName(s) => SanType::DnsName(s.to_string()),
+					GeneralName::URI(s) => SanType::URI(s.to_string()),
+					GeneralName::IPAddress(bytes) if bytes.len() == 4 => {
```
iamjpotts, line 801: There's an example in `fn ip_addr_from_octets` of how to get a byte array from a byte slice without using `unwrap` - though I think you could use that function here and have only one `match` clause for `GeneralName::IPAddress`

src/lib.rs
```diff
@@ -727,15 +727,216 @@ impl CertificateParams {
...
+						let bytes: [u8; 4] = (*bytes).try_into().unwrap();
+						SanType::IpAddress(IpAddr::from(bytes))
+					}
+					GeneralName::IPAddress(bytes) if bytes.len() == 16 => {
+						let bytes: [u8; 16] = (*bytes).try_into().unwrap();
+						SanType::IpAddress(IpAddr::from(bytes))
+					}
+					_ => continue,
```
iamjpotts, line 809: Why wouldn't we want the compiler to flag that an implementation is needed here if a new entry for `GeneralName` is added in the future?

src/lib.rs
```diff
@@ -727,15 +727,216 @@ impl CertificateParams {
...
+
+		let mut result = Vec::new();
+		for subtree in subtrees {
+			let subtree = match &subtree.base {
+				GeneralName::RFC822Name(s) => GeneralSubtree::Rfc822Name(s.to_string()),
+				GeneralName::DNSName(s) => GeneralSubtree::DnsName(s.to_string()),
+				GeneralName::DirectoryName(n) => GeneralSubtree::DirectoryName(DistinguishedName::from_name(&n)?),
+				GeneralName::IPAddress(bytes) if bytes.len() == 8 => {
+					let addr: [u8; 4] = bytes[..4].try_into().unwrap();
+					let mask: [u8; 4] = bytes[4..].try_into().unwrap();
```
iamjpotts, line 926: Should be possible to write this in a way that avoids `unwrap`

iamjpotts (commented):
How concerned are we about having tests? Something like `cfssl` could be used to generate example certificates to parse.

src/lib.rs
est31: `unwrap` is fine here: if it gets triggered, then something bad happened to the `== 8` above, e.g. in a refactor, which would indicate a bug. It likely gets removed by the optimizer anyways. I prefer something with unwrap over something longer or something that uses unsafe.

est31:
Tests would definitely not hurt, they are not required for merging of this PR though.

src/lib.rs
andrenth: I've removed serial number parsing in 5a40247. Hopefully I'll be able to work on a separate PR for that soon.

src/lib.rs
iamjpotts: See my proposed implementation in #101

PR #99, opened 2022-10-30
Add example of generating a new CA and a new server cert signed by it

### LINKED ISSUE #79: Add example to demonstrate how to create a CA and a certificate signed with that CA.
Add example to demonstrate how to create a CA and a certificate signed with that CA.

Basically I want to be able to generate the ca.crt/ca.key and redis.crt/redis.key files that this script generates https://github.com/redis/redis/blob/unstable/utils/gen-test-certs.sh#L5+L6
But I cant figure out how to from looking at the API, it seems like it should be possible though?

iamjpotts:
Resolves #79

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,11 @@ default = ["pem"]
 features = ["x509-parser"]
 
 [dev-dependencies]
+native-tls = "0.2"
 openssl = "0.10"
+pipe = { version = "0.4", features = ["bidirectional"] }
 x509-parser = { version = "0.15", features = ["verify"] }
+rustls = "0.21"
 rustls-webpki = { version = "0.101.0", features = ["std"] }
 rand = "0.8"
 rsa = "0.9"
@@ -48,3 +51,11 @@ botan = { version = "0.10", features = ["vendored"] }
 # ignores profile overrides for non leaf packages)
 [profile.dev.package.num-bigint-dig]
 opt-level = 3
+
+[[example]]
+name = "auto-gen-ca-and-server-nativetls"
+required-features = ["x509-parser"]
+
+[[example]]
+name = "auto-gen-ca-and-server-tls"
+required-features = ["x509-parser"]
--- a/examples/auto-gen-ca-and-server-tls.rs
+++ b/examples/auto-gen-ca-and-server-tls.rs
@@ -1,0 +1,217 @@
+//! Auto-generate a CA certificate and a server certificate signed
+//! by that CA, and start a TLS server and client using them.
+//!
+//! Run with:
+//!
+//!     $ cargo run --example auto-gen-ca-and-server-tls --features=x509-parser
+//!
+//! This doesn't start a network service (not even on localhost).
+//! Instead, it creates an in-memory TLS server and an in-memory
+//! TLS client in two separate threads in the same process.
+
+use std::error::Error;
+use std::io::{BufRead, BufReader, Write};
+use std::sync::Arc;
+use std::thread;
+use rustls;
+use rustls::{PrivateKey, RootCertStore, ServerName};
+use rustls::server::Acceptor;
+
+use rcgen::{BasicConstraints, CertificateSigningRequest, DnType, IsCa, SanType, KeyUsagePurpose, DistinguishedName, CertificateParams, Certificate, RcgenError};
+
+const SAN: &str = "example-server";
+
+fn read_line<A: BufRead>(reader: &mut A) -> Result<String, std::io::Error> {
+	let mut buffer = String::new();
+
+	reader.read_line(&mut buffer)?;
+
+	Ok(buffer.trim_end().into())
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let ca = gen_cert_for_ca()?;
+	let rustls_ca = rustls::Certificate(ca.serialize_der().unwrap());
+
+    println!("CA private key:\n{}", ca.serialize_private_key_pem());
+    println!();
+    println!("CA certificate:\n{}", ca.serialize_pem()?);
+
+    let host_cert = gen_cert_for_server(&ca)?;
+	let host_cert_rustls = rustls::Certificate(host_cert.signed_certificate_der);
+	let host_pk = PrivateKey(host_cert.private_key_der);
+
+    println!("Preparing to verify with tls");
+
+    let leaf_then_ca = vec![host_cert_rustls, rustls_ca.clone()];
+
+    println!();
+
+	// Does not work with either buffered or unbuffered bipipe, even when we use the same message
+	// pattern as with the nativetls example, where client sends then reads, and server reads and
+	// then sends. Client blocks sending a message; server blocks reading a message.
+    let (mut p_client, mut p_server) = pipe::bipipe_buffered();
+
+    let t_client = thread::spawn(move || -> Result<(), String> {
+        println!("Client creating");
+
+		let mut root_store = RootCertStore::empty();
+
+		root_store.add(&rustls_ca)
+			.unwrap();
+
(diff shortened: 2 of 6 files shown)


iamjpotts:
Build failure related to parsing a revoked certificate might be fixed by https://github.com/rustls/rcgen/pull/142.

Cargo.toml
```diff
@@ -33,7 +33,9 @@ default = ["pem"]
 features = ["x509-parser"]
 
 [dev-dependencies]
+native-tls = "0.2"
```
djc, line 36: If we do this with rustls we can don't even need the pipe crate, and we can avoid relying on C libraries.

(Honestly, I'm not sure it makes sense to include a whole TLS setup in the example -- that seems to be more like a testing thing?)

iamjpotts: In my case the 'solution to a problem' was 'how can i create temporary servers for testing with temporary TLS/SSL certificates, and connect clients without suppressing TLS/SSL validations' - a common testing scenario.

iamjpotts:
I've added a commit to replace usage of `nativetls` with `rustls` but I'm stuck trying to fix a bug.

`write_all` hangs in both the client thread and server thread.

I'm sure I'm missing something obvious, lol.

iamjpotts:
Here's the current output:

```
Joining client
Client creating
Server creating
Server accepting
Server reading tls
Client connecting
Client sending
Server accepted
Server created
Server sending
```

Client gets stuck at `Client sending` (`write_all`) and server gets stuck at `Server sending` (also a `write_all`).

I'm sure this is a lack of understanding on my part of one of the crates I'm trying to use. Help would be appreciated.

iamjpotts:
It's getting stuck in the handshake step (`is_handshaking()` is `true` for both server and client connections).

iamjpotts:
Current output after additional commits:

```Preparing to verify with tls

Joining client
Client creating
Server creating
Server accepting
Server reading tls
Client connecting
Client handshaking? true
Client handshaking? true
Client flushing
Server accepted
Server created
Server handshaking? true
Server handshaking? true
Server flushing
```

Gets stuck in same places.

cpu:
> It's getting stuck in the handshake step (is_handshaking() is true for both server and client connections).

I haven't had a chance to review the code you have in progress, but you might be able to crib from the `rustls/rustls/tests/common/mod.rs` code used by the `api.rs` integration tests, in particular the [`do_handshake` helper](https://github.com/rustls/rustls/blob/0f5deca3454528e7b29b47889d1505577b06ea7d/rustls/tests/common/mod.rs#L435-L447)

iamjpotts:
> > It's getting stuck in the handshake step (is_handshaking() is true for both server and client connections).
> 
> I haven't had a chance to review the code you have in progress, but you might be able to crib from the `rustls/rustls/tests/common/mod.rs` code used by the `api.rs` integration tests, in particular the [`do_handshake` helper](https://github.com/rustls/rustls/blob/0f5deca3454528e7b29b47889d1505577b06ea7d/rustls/tests/common/mod.rs#L435-L447)

Thanks for the suggestion. I've looked thru that code (as well as `transfer` and `make_pair_for_arc_configs`) and it requires shared knowledge of both the client and server connections. I'm trying to setup an example where the only shared knowledge is a bidirectional pipe and CA certificate, and the creation of the rustls client is completely independent of creation of the rustls server. It seems like it should be possible.

It was pretty easy to get it working with `nativetls`

iamjpotts:
This isn't a problem specific to the `rustls` implementation in my PR; the `nativetls` has the same problem.

I added `write_all` to that version (not pushed yet) and it has the same problem hanging at write_all in both client and server.

Thanks for your help, I'll keep digging.

est31:
Late to the discussion but +1 to using the `rustls` crate because it means less dependencies.

iamjpotts:
> Late to the discussion but +1 to using the `rustls` crate because it means less dependencies.

Agree.

iamjpotts:
@est31 / @cpu there's a difference in behavior between `rustls` and `nativetls` when the client and server are communicating over a bidirectional pipe from the `pipe` crate.

Let's assume the following behavior (implemented in all three `examples`):

* Client and server start independently, on separate threads
* Client sends a message and then reads, blocking waiting for a response
* Server blocks waiting for a message, then sends a response

With `nativetls`, both client and server run to completion, whether the `pipe` bipipe is buffered or unbuffered.

With `rustls`, both the client and server block indefinitely, whether the `pipe` bipipe is buffered or unbuffered. It seems `rustls` gets stuck handshaking.

`plumber` has the same client/server behaviors, but without any tls implementation on top. Client writes then reads, while server reads then writes. `plumber` will run to completion whether its buffered or unbuffered.

If `plumber` is altered so both client and server write before reading, only `pipe::bipipe_buffered` will allow both client and server to run to completion; `pipe::bipipe` (unbuffered) blocks indefinitely.

If `nativetls` is altered so both client and server write before reading, using `pipe::bipipe_buffered` does not fix the indefinite blocking.

Is there a bug in rustls - should it be able to run on a buffered bidirectional pipe with a rusttls client hooked up to one end of the pipe and a rusttls server hooked up to the other end?

iamjpotts:
Also, while I have included `plumber.rs` and a `nativetls` based version of the example, those are temporary for comparison to the rustls behavior, and will be removed once the issues with the rusttls example are resolved.

examples/auto-gen-ca-and-server-tls.rs
```diff
@@ -0,0 +1,217 @@
...
+    println!("Preparing to verify with tls");
+
+    let leaf_then_ca = vec![host_cert_rustls, rustls_ca.clone()];
+
+    println!();
+
+	// Does not work with either buffered or unbuffered bipipe, even when we use the same message
+	// pattern as with the nativetls example, where client sends then reads, and server reads and
+	// then sends. Client blocks sending a message; server blocks reading a message.
+    let (mut p_client, mut p_server) = pipe::bipipe_buffered();
```
djc, line 53: Note that there's really no need to use a pipe thing here. You can just directly use a `Vec<u8>` buffer (or two) that one side writes into and the other side then reads from.

iamjpotts: The vectors couldn't be moved into the server thread and client threads, because it would require moving each vector twice.

cpu:
Is the status of this branch still accurately reflected in the "Do not merge" PR title?

I appreciate the effort you've put into this work, and I do agree that better examples are needed. However, I think the approach taken here is too heavy-weight. It's more of an example of using the downstream TLS crates vs using `rcgen`. In general I worry we're duplicating code that would be better offered as an example maintained in those upstream repositories where it would be kept in-sync with project changes. I also think the choice to try and make a bi-directionally communicating client/server pair complicates the code and doesn't match what I expect most users want to do: create a client, or create a server, and use those with a TCP socket based transport.

I would feel a lot better about examples that focused exclusively on certificate generation, the domain of rcgen. If it's true that there are still gaps that make it challenging for a new user to use the generated certificates with Rustls then I think the best course of action would be to try and extend the [existing Rustls examples](https://github.com/rustls/rustls/tree/main/examples/src/bin) to better resolve those challenges.

Apologies for not voicing these opinions earlier.

iamjpotts:
No worries. I was actually starting to get that same feeling myself - like the scope of the proposed code changes had changed, and that at least some portion of what it became might belong in another crate.

It was here originally to show how to use rcgen to solve a problem, and provide an end-to-end solution.

I am a little concerned about the different behavior of the tls implementations I documented in a comment above, but that may belong elsewhere also:

"there's a difference in behavior between rustls and nativetls when the client and server are communicating over a bidirectional pipe from the pipe crate."

Let me think about it a bit and I may proactively refactor or close the PR.

PR #101, opened 2022-10-31
Add CidrSubnet::from_addr_and_mask

iamjpotts:
I address [my own comment](https://github.com/est31/rcgen/pull/96/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R907) on #96 at by adding a new method onto `CidrSubnet` to parse it from a byte slice.

Includes tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -312,6 +312,30 @@ impl CidrSubnet {
 			},
 		}
 	}
+
+	/// Extract the IP address and subnet mask.
+	///
+	/// For IPv4, the source is a four-byte IP and four-byte mask.
+	/// For IPv6, the source is a 16-byte IP and 16-byte mask.
+	#[allow(dead_code)]
+	fn from_addr_and_mask(octets: &[u8]) -> Result<Self, RcgenError> {
+		if let Ok(value) = <&[u8; 32]>::try_from(octets) {
+			if let Ok(addr) = <&[u8; 16]>::try_from(&value[..16]) {
+				if let Ok(mask) = <&[u8; 16]>::try_from(&value[16..]) {
+					return Ok(Self::V6(*addr, *mask));
+				}
+			}
+		} else if let Ok(value) = <&[u8; 8]>::try_from(octets) {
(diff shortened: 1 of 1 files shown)


est31:
I'm not really sure this is good as a `pub` fn, but as a non-pub fn it might make sense. Ultimately I'm fine with the unwrap-based solution in #96 as well. Also, style wise, the } should be on the same line as `else`.

iamjpotts:
> I'm not really sure this is good as a `pub` fn, but as a non-pub fn it might make sense. Ultimately I'm fine with the unwrap-based solution in #96 as well. Also, style wise, the } should be on the same line as `else`.

Fixed the brace location.

It would be nice if there was a way to reference a portion of an array as an array rather than as a slice. Both the unwrap and non-unwrap implementations are unideal.

I think refactoring the parsing of the byte slice into a `CidrSubnet` is something that should be its own standalone function in either case. It's easier to read than having every different type of parser in a match clause of a large function.

est31:
Closing this as the current code is fine by me. But feel free to open a PR that addresses [this](https://github.com/est31/rcgen/pull/96#discussion_r1008985441) suggestion of yours.

iamjpotts:
> Closing this as the current code is fine by me. But feel free to open a PR that addresses [this](https://github.com/est31/rcgen/pull/96#discussion_r1008985441) suggestion of yours.

Done, see #102

iamjpotts:
@est31 now that you can see where my head's at with #102 do you mind if I revisit pulling out the IP address to CidrSubnet conversion?

I can keep the landed implementation instead of the one I have in this PR.

I'll submit a new PR if you want me to move forward with this refactor.

est31:
What do you mean? It already calls the `ip_addr_from_octets` function, no?

iamjpotts:
> What do you mean? It already calls the `ip_addr_from_octets` function, no?

That change was for IP only. This change was for IP + CIDR mask parsing out of a single byte slice.

PR #102, opened 2022-10-31
Deduplicate GenericName to SanType conversion

iamjpotts:


est31 (approved):
Very nice, thank you!

PR #106, opened 2022-12-14
authority_key_identifier now possible on self-signed

### LINKED ISSUE #105: Authority Key Identifier not possible on self-signed certificates
The documentation for CertificatesParams mention for use_authority_key_identifier_extension:

> If true (and not self-signed)

As a reason for the self-signed restriction, the  [source code](https://docs.rs/rcgen/0.10.0/src/rcgen/lib.rs.html#827-834) cites a section from RFC 5280:

> The keyIdentifier field of the authorityKeyIdentifier extension MUST
   be included in all certificates generated by conforming CAs to
   facilitate certification path construction.  There is one exception;
   where a CA distributes its public key in the form of a "self-signed"
   certificate, the authority key identifier MAY be omitted.

But to my understanding this means the omission of the extension is optional and not that it has to be omitted.

Therefore the exception for self-signed certificates should be removed.

M0urban:
fixes #105

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -653,7 +653,7 @@ pub struct CertificateParams {
 	pub custom_extensions :Vec<CustomExtension>,
 	/// The certificate's key pair, a new random key pair will be generated if this is `None`
 	pub key_pair :Option<KeyPair>,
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks!

PR #107, opened 2023-02-01
Make sure that all CertificateParams can be serialized in a CSR

g2p:
In write_request, we now serialize custom extensions as extension requests.  We also check that all CertificateParams fields are either empty or that we can serialize them.

We do ignore a few fields: not_before, not_after, and key_identifier_method.
NotBefore and NotAfter cannot be specified in a CSR, a key identifier might be but would require a use case and API changes.

Fixes https://github.com/est31/rcgen/issues/64.


src/lib.rs
```diff
@@ -920,13 +920,48 @@ impl CertificateParams {
...
+			key_identifier_method,
+		} = self;
+		// - alg and key_pair will be used by the caller
+		// - not_before and not_after cannot be put in a CSR
+		// - There might be a use case for specifying the key identifier
+		// in the CSR, but in the current API it can't be distinguished
+		// from the defaults so this is left for a later version if
+		// needed.
+		let _ = (alg, key_pair, not_before, not_after, key_identifier_method);
+		if !(serial_number.is_none()
```
est31, line 950: Could you remove this negation by doing de morgan?

est31 (changes requested):

src/lib.rs
g2p: Sure!

est31 (approved):
Thanks!

PR #113, opened 2023-04-03
Change Certificate::from_params and generate_simple_self_signed to take SanType instead of String

rukai:
Breaking change.

Its quite unexpected when I do:
`CertificateParams::new(&["127.0.0.1".to_owned()])` and that generates a DNSName entry of 127.0.0.1


est31:
I understand the concern but I don't think this is a good way to address it. I want to keep that API as simple as possible. People shouldn't have to import the `SanType` struct. Ideally it would try to parse it as an ip first and then as a DNS name.

rukai:
IMO, making the user think about what they are doing is always better than trying to guess what they mean. Especially so for something as important as a certificate.

However your suggestion of trying to parse an ip address is still an improvement towards guessing and better than nothing, so I'll close this PR and attempt that instead.

PR #114, opened 2023-04-03
CertificateParams::new attempt to parse ip address

rukai:
supercedes https://github.com/est31/rcgen/pull/113

Guesses what the user meant in some circumstances, email addresses and uri's will still be misunderstood.


est31 (approved):
Thanks!

PR #116, opened 2023-04-05
Update pem to v2.0.1

koushiro:


est31 (approved):
Thanks!

est31:
> error: package `pem v2.0.0` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.63.0

Hmm I suppose this means I don't want to merge this at this point, because 1.67.0 is way too new. See also [this discussion](https://github.com/jcreekmore/pem-rs/commit/b19353d2849b234672a64cd80a80bdaae7836787#r107365915).

koushiro:
I don't know why @jcreekmore use 1.67.0 as msrv of [pem-rs](https://github.com/jcreekmore/pem-rs) 2.0.
Because the newest `pem-rs` use `dep` syntax in the `Cargo.toml`, so maybe `rust-version = "1.60.0"` is enough.

koushiro:
also there is a feature issue of pem-rs 2.0 https://github.com/jcreekmore/pem-rs/pull/45

jcreekmore:
Yeah, my whole philosophy behind MSRV is "if I am introducing a big breaking change, bring it up to whatever is current" because no one _has_ to update from 1.x to 2.x and I want to be as unconstrained as possible in development (I don't want to have to think about what version features got added to the compiler when I am developing). With all that said, I don't see any issues with dropping it to 1.60.0 because, as I said, it was fairly arbitrary and I don't foresee any new feature development anytime soon.

PR #117, opened 2023-04-06
apply rustfmt and clippy for the crate

koushiro:
- apply rustfmt and clippy for the crate
- add rustfmt and clippy check into ci
- add msrv (1.63.0) check into ci


est31:
Some comments:

* The MSRV is not fixed: it's always 7 months ago, so I like the "stable 7 months ago" check that is there currently
* I don't like rustfmt's output, so I don't want the application of rustfmt
* clippy can be added to CI, yes, but don't remove the allow declarations. They are there for a purpose.

PR #118, opened 2023-04-10
Use line endings appropriate for current platform

frjonsen:
PR in reference to https://github.com/est31/rcgen/issues/115

Adds platform checking to decide what line endings to use when serializing to PEM.

I tried activating Github Actions for Windows to test correctness for that platform too, but unfortunately it seems the dev-dependency `openssl` does not work: https://github.com/frjonsen/rcgen/actions/runs/4655417259/jobs/8238039795

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,12 @@ const OID_NAME_CONSTRAINTS :&[u64] = &[2, 5, 29, 30];
 // https://www.iana.org/assignments/smi-numbers/smi-numbers.xhtml#smi-numbers-1.3.6.1.5.5.7.1
 const OID_PE_ACME :&[u64] = &[1, 3, 6, 1, 5, 5, 7, 1, 31];
 
+#[cfg(feature = "pem")]
+const ENCODE_CONFIG: pem::EncodeConfig = match cfg!(target_family = "windows") {
+	true => pem::EncodeConfig { line_ending: pem::LineEnding::CRLF },
+	false => pem::EncodeConfig { line_ending: pem::LineEnding::LF },
+};
+
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 #[allow(missing_docs)]
 #[non_exhaustive]
@@ -629,7 +635,7 @@ impl CertificateSigningRequest {
 	pub fn serialize_pem_with_signer(&self, ca :&Certificate) -> Result<String, RcgenError> {
 		let contents = self.params.serialize_der_with_signer(&self.public_key, ca)?;
 		let p = Pem::new("CERTIFICATE", contents);
-		Ok(pem::encode(&p))
+		Ok(pem::encode_config(&p, ENCODE_CONFIG))
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -148,6 +148,12 @@ const OID_NAME_CONSTRAINTS :&[u64] = &[2, 5, 29, 30];
 // https://www.iana.org/assignments/smi-numbers/smi-numbers.xhtml#smi-numbers-1.3.6.1.5.5.7.1
 const OID_PE_ACME :&[u64] = &[1, 3, 6, 1, 5, 5, 7, 1, 31];
 
+#[cfg(feature = "pem")]
+const LINE_ENDING: pem::LineEnding = match cfg!(target_family = "windows") {
```
est31, line 152: Could you make this a `pem::EncodeConfig`, or alternatively, make another const that has declares that `EncodeConfig`? Then one would save on repetition.

frjonsen:
> I think one can add openssl via vcpkg: https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L94

This did indeed solve this particular issue, but I've been getting some follow up issues for other packages. [I've resolved one](https://github.com/est31/rcgen/actions/runs/4664562608/jobs/8257074170). but currently [I'm stuck on building `botan`](https://github.com/est31/rcgen/actions/runs/4665102920/jobs/8258857277?pr=118). Trying to resolve these other issues is unfortunately taking more time than I currently have to offer, so I'm not quite sure when I will be able to finish up this PR.

est31:
@frjonsen it's ok, is there a way to mark a github CI job as allow failure? Or alternatively, one could comment it all out.

frjonsen:
> @frjonsen it's ok, is there a way to mark a github CI job as allow failure? Or alternatively, one could comment it all out.

Could probably do that. I see though that you've been working on resolving this, so thanks for that. I was looking into this during work time, so unfortunately had to prioritize other things, since using the `pem` lib directly does work as a workaround.

est31:
> I see though that you've been working on resolving this, so thanks for that.

I've thought that your PR is useful and it's not really your fault that there is no CI for windows right now, so I thought I would help you. PR #103 I left open for a long time hoping for smaller fixes to it, but they didn't happen. Later on it turned out that one of the functions I wrote to support that PR had an issue.

botan's maintainer has been very helpful in resolving the build issues on windows, I think now we'll have to wait until a release of botan includes https://github.com/randombit/botan/pull/3504

est31:
Update: there has been no release of botan yet. There is also a PR https://github.com/randombit/botan-rs/pull/84 to add windows CI to botan-rs, which seems to be green for them.

randombit:
Hi @est31 just wanted to give an update here - newly released `botan 0.10.3` does include the support for setting the source file link method. But Windows CI is still stalled, it is "green" but actually a no-op. I need to write a new script for CI orchestration.

est31:
Hmm I wonder if it's possible to have the botan dev dependency and tests only for non-windows platforms. @frjonsen could you put botan into `[target.'cfg(not(windows))'.dependencies]` in Cargo.toml, and add cfg's to the test file?

est31:
Hmm I pushed it manually... now it complains about openssl not being found :/ Not sure why that is the case given that we install it just before via vcpkg.

est31:
I have successfully merged this PR by disabling botan on the windows CI. Sorry for the delay @frjonsen. I have opened #125 for enabling botan on the windows CI once it's ready.

frjonsen:
Thanks for finishing the PR. It looks like you had a very productive day yesterday!

est31:
Yes I did some trial and error with the CI to make it green. Sorry for the notification noise :).

PR #119, opened 2023-04-25
Updating cert param/key pair comments

Will-Low:
I was a little confused by this comment, since it wasn't clear what code was responsible for generating this key pair. https://github.com/est31/rcgen/blob/14dfa20ed86b78348ce9ddd4d88cc1dc04e702bd/src/lib.rs#L651

I might suggest removing the comment from this field, instead documenting this behavior on `Certificate::from_params`, which appears to be that the original comment was referring to.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -648,7 +648,6 @@ pub struct CertificateParams {
 	pub extended_key_usages :Vec<ExtendedKeyUsagePurpose>,
 	pub name_constraints :Option<NameConstraints>,
 	pub custom_extensions :Vec<CustomExtension>,
-	/// The certificate's key pair, a new random key pair will be generated if this is `None`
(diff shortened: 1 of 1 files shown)


est31:
I'd be okay with having both, but not with removing the parameter from the struct. Also, can you put an empty line between the first and second sentence of the function documentation?

Will-Low:
I was thinking about leaving both initially, but I didn't like that the comment refers to a piece of code that could change divergently. It seems to introduce some unnecessary coupling. What do you think?

> Also, can you put an empty line between the first and second sentence of the function documentation?

Sure thing

est31:
I like the docs on the field. The docs on the function aren't as clear without that additional hint.

Will-Low:
Updated per the above.

est31 (approved):
Thank you!

PR #125, opened 2023-06-15
Enable botan CI on windows

est31:
In #118, windows CI was added but the botan specific tests were disabled as the build for rust doesn't work.

I'm opening this PR for future discussion and work on botan support on the windows CI. The goal is to get the CI green, and until then it can't be merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,9 @@ jobs:
     - uses: actions/checkout@v4
       with:
         persist-credentials: false
+    - uses: egor-tensin/vs-shell@v2
+      with:
+        arch: amd64
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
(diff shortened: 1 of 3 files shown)


est31:
Some amazing news: https://github.com/randombit/botan-rs/issues/82#issuecomment-1636508346

cpu:
@est31 I pushed some revisions onto this branch and I think it's ready to go if you want to review.

* I rebased on tip of `main`.
* I dropped the `Cargo.lock` update commit. We're already using the latest `botan-rs`/`botan-src`.
* I added a commit (9a5424057f6ee01d3810781cd2321c713714e322) that matches the [upstream CI](https://github.com/randombit/botan-rs/blob/b95b871342c4c82d288f8bc272c408e91aafe3c6/.github/workflows/ci.yml#L124-L126)'s usage of [`egor-tensin/vs-shell@v2`](https://github.com/egor-tensin/vs-shell) for setting up env vars. The commit message has more detail but there's a compile time error without it and I'm not familiar enough with PowerShell/Windows to pull out a more minimal fix (it also seems smart to match how the upstream tests their builds).

est31:
@cpu thanks! I think with your changes, we can merge it.

PR #127, opened 2023-06-16
Turn botan back into a dev-dependency

mbrubeck:
Fixes a regression from d3ab04a84b78bed81c67d21873581302c787fc25 (#118).

See also c8eb1e689dbd1558738090f2a5e3e06117c66d2e (#43).


est31 (approved):
Wow I did the same mistake again as was done in ef63ca7235eb64ddc28243ec3b23e5eff0353756, lol. Good catch.

PR #128, opened 2023-06-18
Certificate Revocation List (CRL) generation support.

cpu:
## Description

This branch implements support for generating RFC 5280 Certificate Revocation Lists (CRLs) using rcgen. 

My primary motivation for this work is to offer the ecosystem a pure-Rust method for generating CRLs that will work with [Rustls ](https://github.com/rustls/rustls) as part of in-progress support for client certificate revocation checking. As such, this work doesn't attempt to offer all possible CRL features but a minimal subset I think makes the most sense as a starting point.

I've done my best to match the style of the crate for this new feature, but as this is my first contribution there may be room for further improvements.

Resolves https://github.com/est31/rcgen/issues/109

### CertificateParams: lift write_extension fn.
Previously the `CertificateParams` type defined a member function, `write_extension`, that could serialize X509 extensions per RFC 5280. 

Conveniently, Certificate revocation lists (CRLs) use the exact same mechanism for both CRL and revoked cert extensions.

In preparation for supporting generation of CRLs this commit lifts the existing function into a free-standing version, renamed to
`write_x509_extension`. The existing `CertificateParams` code is updated to use this function.

### CertificateParams: lift AKI extension fn.
Previously the `CertificateParams` type's `write_cert` fn had a code fragment used to write the authority key identifier (AKI) extension defined by [RFC 5280 Section 4.2.1.1](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.1).

Later, in [RFC 5280 Section 5.2.1](https://www.rfc-editor.org/rfc/rfc5280#section-5.2.1), there is text saying:
> Conforming CRL issuers MUST use the key identifier method, and MUST
> include this extension in all CRLs issued.

Conveniently, the AKI extension implementation for CRLs is identical to certificates. Both use the definition from 4.2.1.1.

In preparation for supporting generation of CRLs that adhere to the specification this commit lifts the existing code for AKI serialization to a free-standing version named `write_x509_authority_key_identifier`. The existing `CertificateParams.write_cert` code is updated to use this function.

### lib: add support for creating CRLs.
This commit introduces support for generating signed Certificate Revocation Lists (CRLs) according to the profile defined in [RFC 5280 Section 5](https://www.rfc-editor.org/rfc/rfc5280#section-5).

This implementation closely follows the feature set (_soon to be_) supported by the [Rustls `webpki` crate](https://github.com/rustls/webpki). It is _not_ a complete implementation of all possible CRL features. Notably it doesn't support:

* Indirect CRLs - that is, CRLs that have revoked certificates issued by an issuer separate from the issuer of the CRL.
* Delta CRLs - that is, CRLs that only include the differences from a previously issued CRL.

There is also limited flexibility for issuing CRLs that don't adhere to 5280. E.g. specifying an incorrect version, omitting a CRL number, omitting the authority key identifier extenion. These are required by 5280 so let's please not pollute the world with more out-of-spec DER. I assume anyone needing to generate malformed things on purpose can use tools like [`der-ascii`]( https://github.com/google/der-ascii) instead.

### tests: add certificate revocation list tests.
This commit adds tests that ensure rcgen created CRLs parse/validate successfully with OpenSSL and x509-parser.

Coverage is not added for Botan because it appears the Botan rust bindings don't expose the required `X509_CRL` type (https://github.com/randombit/botan-rs/issues/95).

Coverage is not (yet) added for Webpki because the upstream crate hasn't yet cut a release that includes CRL parsing support. I've started on this test coverage (see https://github.com/est31/rcgen/commit/37ada8d5132dd971977071be61bf5765838638c1) but since it will require a `Cargo` patch to use the unreleased `rustls/webpki` I've held off on completing that work or including it for review.


tests/generic.rs
```diff
@@ -77,3 +77,50 @@ mod test_convert_x509_subject_alternative_name {
 		assert!(actual.subject_alt_names.contains(&ip_san));
 	}
 }
+
+#[cfg(feature = "x509-parser")]
+mod test_x509_parser_crl {
```
cpu, line 82: I put these tests in `generic.rs` but perhaps it would be better to create a x509-parser.rs test file?

tests/generic.rs
```diff
@@ -77,3 +77,50 @@ mod test_convert_x509_subject_alternative_name {
...
+		// The properties of the CRL should match expected.
+		assert_eq!(x509_crl.version().unwrap(), X509Version(1));
+		assert_eq!(x509_crl.issuer(), x509_issuer.subject());
+		assert_eq!(x509_crl.last_update().to_datetime().unix_timestamp(),
+				   crl.get_params().this_update.unix_timestamp());
+		assert_eq!(x509_crl.next_update().unwrap().to_datetime().unix_timestamp(),
+				   crl.get_params().next_update.unix_timestamp());
+		// TODO(XXX): Waiting on https://github.com/rusticata/x509-parser/pull/144
+		// let crl_number = BigUint::from_bytes_be(crl.get_params().crl_number.as_ref());
+		// assert_eq!(x509_crl.crl_number().unwrap(), &crl_number);
```
cpu, line 114: At the cost of more test code we could also replicate what `crl_number()` is doing:

* iterating extensions
* finding the crl number OID (using the correct OID :stuck_out_tongue_closed_eyes: )
* unmarshaling the value

The effort/value tradeoff didn't seem that great so I left this as-is until the upstream crate lands my fix + publishes a release. Open to revisiting if you think it's worthwhile to have this assertion present today.

src/lib.rs
```diff
@@ -1463,6 +1508,253 @@ pub enum KeyIdMethod {
 	Sha512,
 }
 
+/// A certificate revocation list number.
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct CrlNumber {
```
est31, line 1513: Could you use `SerialNumber` instead? This is more or less a copy+rename of that type.

tests/generic.rs
est31: I don't think it's needed to have it from the get-go. If there is a hidden bug, we'll just fix it once we discover it, which at the latest is when a released version of upstream has your patch.

src/lib.rs
```diff
@@ -639,6 +650,75 @@ impl CertificateSigningRequest {
...
+///   crl_number: CrlNumber::from(1234),
+///   revoked_certs: vec![revoked_cert],
+///   alg: &PKCS_ECDSA_P256_SHA256,
+///   key_identifier_method: KeyIdMethod::Sha256,
+/// };
+/// let crl = CertificateRevocationList::from_params(crl).unwrap();
+/// println!("{}", crl.serialize_pem_with_signer(&issuer).unwrap());
+///# }
+pub struct CertificateRevocationList {
+	params : CertificateRevocationListParams,
```
est31, line 687: Style nit: you use `params :CertificateRevocationListParams` here? My idea is that I always do `<space>:<space>` if it's used like an assignment, and `<space>:` if it's indicating a type.

est31 (changes requested):
Initial review looks mostly good, I haven't given it a close look though (yet)

tests/generic.rs
est31: It's fine I think. the mod right above is doing the same.

src/lib.rs
cpu: Thanks, that one slipped through. Fixed.

src/lib.rs
cpu: Sure, I was on the fence about this one. 5280 describes it as a distinct thing (e.g. not calling it the CRL's serial number) but the specification of encoding is the same as serial so having it distinct in the type system is probably overkill.

One difference I had implemented in the `CrlNumber` type was erroring from the `from_slice` constructor if given something larger than 20 octets. The `SerialNumber` implementation doesn't do that but it seems like a minor point (and could be considered separately, perhaps with other validation).

In any case, I replaced `CrlNumber` with `SerialNumber` :+1:

tests/generic.rs
cpu: Sounds good. I did smoke-test my upstream fix with a `Cargo` patch on rcgen and the assertion worked as expected with the patched x509-parser.

djc:
@est31 do you have an idea of when you'll be able to look at this in more detail? We're discussing using this support in the rustls test suite.

est31 (approved):
Thanks!

est31:
@djc merged. Do you need a crates.io release with this PR?

djc:
Not for now, will ping you if/when we do. Thanks for the fast turnaround!

cpu:
> Thanks for the fast turnaround!

+1 - thank you :tada:

PR #131, opened 2023-06-28
Fix OID for ExtendedKeyUsagePurpose::Any

### LINKED ISSUE #130: Incorrect oid for ExtendedKeyUsagePurpose::Any?
When I create a certificate using these parameters:

```rust
let mut params = CertificateParams::new(vec![HOST.to_string()]);
params.extended_key_usages = vec![ExtendedKeyUsagePurpose::Any];
```

and then parse it using [x509-parser](https://docs.rs/x509-parser), the extended key attributes parse like this:

```
ExtendedKeyUsage {
        any: false,
        server_auth: false,
        client_auth: false,
        code_signing: false,
        email_protection: false,
        time_stamping: false,
        ocsp_signing: false,
        other: [
            OID(2.5.29.37),
        ],
}
```

It looks like rcgen is producing [`OID(2.5.29.37)`](https://github.com/est31/rcgen/blob/3a8535a0724963e62dc3828ad8c73c957396a883/src/lib.rs#L1431), but x509-parser is expecting [`OID(2.5.29.37.0)`](https://docs.rs/x509-parser/0.15.0/src/x509_parser/extensions/keyusage.rs.html#131). Going by https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12:

```
id-ce OBJECT IDENTIFIER  ::=  {joint-iso-ccitt(2) ds(5) 29}
id-ce-extKeyUsage OBJECT IDENTIFIER ::= { id-ce 37 }
anyExtendedKeyUsage OBJECT IDENTIFIER ::= { id-ce-extKeyUsage 0 }
```

it looks like x509-parser is correct and rcgen is missing the trailing 0.

jgallagher:
Fixes #130

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1428,7 +1428,7 @@ impl ExtendedKeyUsagePurpose {
 		use ExtendedKeyUsagePurpose::*;
 		match self {
 			// anyExtendedKeyUsage
-			Any => &[2, 5, 29, 37],
+			Any => &[2, 5, 29, 37, 0],
 			// id-kp-*
 			ServerAuth => &[1, 3, 6, 1, 5, 5, 7, 3, 1],
 			ClientAuth => &[1, 3, 6, 1, 5, 5, 7, 3, 2],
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks"

PR #133, opened 2023-07-04
switch to rustls/webpki, add test coverage for CRL parsing.

cpu:
## Description

This is a small follow-up to https://github.com/est31/rcgen/pull/128 now that `rustls/webpki` v0.101.0 has been released. Resolves https://github.com/est31/rcgen/issues/129

### deps: switch to rustls webpki v0.101.0.
The [original webpki](https://github.com/briansmith/webpki) crate is under-maintained, so the [Rustls project](https://github.com/rustls/rustls) has [forked it](https://github.com/rustls/webpki) and invested in maintenance and new features, including [CRL support](https://docs.rs/rustls-webpki/latest/webpki/struct.BorrowedCertRevocationList.html) for client certificate revocation checking.

This commit switches rcgen to use the latest version of the rustls/webpki fork instead of the original upstream webpki project. One small breaking change in the end entity name validation API is fixed in the existing rcgen webpki tests in order to make this switch build/test cleanly.

### tests: update webpki tests to include CRL support.

This commit updates the webpki test suite to include support for parsing and validating a CRL generated by rcgen, as well as a test that checks a revoked EE client cert is rejected when a CRL is provided revoking the cert.

### DIFF
--- a/tests/util.rs
+++ b/tests/util.rs
@@ -81,7 +81,7 @@ pub fn test_crl() -> (CertificateRevocationList, Certificate) {
 	let now = OffsetDateTime::now_utc();
 	let next_week = now + Duration::weeks(1);
 	let revoked_cert = RevokedCertParams{
-		serial_number: SerialNumber::from(9999),
+		serial_number: SerialNumber::from_slice(&[0xC0, 0xFF, 0xEE]),
 		revocation_time: now,
 		reason_code: Some(RevocationReason::KeyCompromise),
 		invalidity_date: None,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -503,8 +503,8 @@ dependencies = [
  "rand",
  "ring",
  "rsa",
+ "rustls-webpki",
  "time",
- "webpki",
  "x509-parser",
  "yasna",
  "zeroize",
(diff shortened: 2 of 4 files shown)


tests/util.rs
```diff
@@ -81,7 +81,7 @@ pub fn test_crl() -> (CertificateRevocationList, Certificate) {
 	let now = OffsetDateTime::now_utc();
 	let next_week = now + Duration::weeks(1);
 	let revoked_cert = RevokedCertParams{
-		serial_number: SerialNumber::from(9999),
+		serial_number: SerialNumber::from_slice(&[0xC0, 0xFF, 0xEE]),
```
cpu, line 84: I switched to using `from_slice(&[u8])` here because `from(u64)` was converting the `u64` input into a big endian encoded slice that was padded out to the size of a `u64`, where the webpki side was returning a serial number `&[u8]` slice that had no padding. Changing to using a raw slice as input to the `RevokedCertParams` seemed like the simplest fix.

est31 (changes requested):
The PR is really great so far, but would it be possible to also add a test if one tries to use a revoked cert, it is actually refused?

cpu:
> The PR is really great so far, but would it be possible to also add a test if one tries to use a revoked cert, it is actually refused?

👍 Good idea, I can do that (_probably ~tomorrow, I'm just logging off for the day_)

cpu:
> would it be possible to also add a test if one tries to use a revoked cert, it is actually refused?

Force pushed an updated revision to add a test case for this.

cpu:
Thanks for the reviews :-) I'll take a crack at the draft botan-rs test cases shortly.

PR #136, opened 2023-08-16
lib: add cert and CRL distribution point extension support.

cpu:
## Description

This branch adds support for certificate CRL distribution points extensions, and the corresponding CRL issuing distribution point extension.

Initially I opened this PR with just the former, but the latter is so closely related it felt more appropriate to make one branch with two commits. 

Test coverage is better for the certificate extension compared to the CRL extension, the upstream library support for the latter is quite thin.

I haven't updated the webpki CRL revocation checking tests to demonstrate the new extensions at this time, because it is only available in an alpha release. I can either do this as a follow up once the v/0.102.0 release is cut (likely in O(weeks)) or we could update the dev-dependency to the alpha now and add coverage in this branch. LMK what you would prefer :-)

### lib: add cert CRL distribution points ext. support.
This commit extends rcgen to allow generating certificates that contain an RFC 5280 certificate revocation list (CRL) distribution points extension. This is a useful mechanism for helping ensure CRL coverage when performing revocation checks, and is newly supported by rustls/webpki. See this upstream [webpki issue](https://github.com/rustls/webpki/issues/121) and [RFC 5280 §4.2.1.13](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.13) for more background.

Using the new `crl_distribution_points` field of the `CertificateParams` struct it's possible to encode one or more distribution points specifying URI general names where up-to-date CRL information for the certificate can be found.

Similar to existing rcgen CRL generation, the support for this extension is not extensive, but instead tailored towards usage in the web PKI with a RFC 5280 profile.

Notably this means:
* There's no support for specifying the `reasons` flag - RFC 5280 "RECOMMENDS against segmenting CRLs by reason code".
* There's no support for specifying a `cRLIssuer` in the DP - this is specific to indirect CRLs, and neither rcgen's CRL generation code or webpki's parsing/validation support these. 
* There's no support for specifying a `nameRelativeToCrlIssuer` in the DP name instead of a sequence of general names for similar reasons as above: 5280 says: "Conforming CAs SHOULD NOT use nameRelativeToCRLIssuer to specify distribution point names."
* There's no support for specifying general names of type other than URI within a DP name's full name. Other name types either don't make sense in the context of this extension, or are rarely useful in practice (e.g. directory name).

Test coverage is mixed based on the support of the relevant third party libraries. OpenSSL (openssl-rs) and x509-parser both parse this extension well, and so the `openssl.rs` and `generic.rs` test coverage is the most extensive. Webpki (v/0.102.0-alpha.0) recognizes this extension for use during revocation checking, but doesn't expose it externally so a simple parse test is added. Botan's rust bindings do not recognize the extension or offer a way to pull out arbitrary extensions, so no test coverage is added there.

### lib: add CRL issuing distribution point ext. support.
This commit extends rcgen to allow generating certificate revocation lists (CRLs) that contain an RFC 5280 CRL issuing distribution point extension. This is a useful mechanism for helping ensure CRL coverage when performing revocation checks, and is newly supported by rustls/webpki. See this upstream [webpki issue](https://github.com/rustls/webpki/issues/121) and [RFC 5280 §5.2.5](https://www.rfc-editor.org/rfc/rfc5280#section-5.2.5) for more background.

Using the new optional `issuing_distribution_point` field of the `CertificateRevocationListParams` struct it's possible to encode a
issuing distribution point specifying URI general names where up-to-date CRL information for the CRL can be found.

Similar to existing rcgen CRL generation, the support for this extension is not extensive, but instead tailored towards usage in the web PKI with a RFC 5280 profile.

Notably this means:
* There's no support for specifying the `indirectCRL` bool - neither rcgen's existing CRL generation code or webpki's parsing/validation supports these.
* There's no support for specifying the `onlySomeReasons` field - RFC 5280 "RECOMMENDS against segmenting CRLs by reason code". 
* There's no support for specifying a `onlyContainsAttributeCerts` bool - RFC 5280 says "Conforming CRLs issuers MUST set the onlyContainsAttributeCerts boolean to FALSE." and the DER encoding of 'false' requires eliding the value.
* There's no support for specifying a 'nameRelativeToCrlIssuer' in the  DP name instead of a sequence of general names for similar reasons as  above: 5280 says: "Conforming CAs SHOULD NOT use nameRelativeToCRLIssuer to specify distribution point names."
* There's no support for specifying general names of type other than URI  within a DP name's full name. Other name types either don't make sense in the context of this extension, or are rarely useful in practice (e.g. directory name).

Compared to test coverage of the certificate CRL distribution points extension this commit can't offer too much. OpenSSL (openssl-rs) doesn't expose arbitrary CRL extensions, or the issuing distribution point. The `x509-parser` crate can pull out the extension, but doesn't decompose the value (I may attempt to land code for this upstream in the future, stay tuned). Webpki (v/0.102.0-alpha.0) recognizes this extension for use during revocation checking, but doesn't expose it externally. Botan's rust bindings do not recognize the extension or offer a way to pull out arbitrary extensions, so no test coverage is added there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -163,6 +163,10 @@ const OID_CRL_REASONS :&[u64] = &[2, 5, 29, 21];
 // https://www.rfc-editor.org/rfc/rfc5280#section-5.3.2
 const OID_CRL_INVALIDITY_DATE :&[u64] = &[2, 5, 29, 24];
 
+// id-ce-issuingDistributionPoint
+// https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.5
+const OID_CRL_ISSUING_DISTRIBUTION_POINT :&[u64] = &[2, 5, 29, 28];
+
 #[cfg(feature = "pem")]
 const ENCODE_CONFIG: pem::EncodeConfig = match cfg!(target_family = "windows") {
 	true => pem::EncodeConfig { line_ending: pem::LineEnding::CRLF },
@@ -680,6 +684,7 @@ impl CertificateSigningRequest {
 ///   this_update: date_time_ymd(2023, 06, 17),
 ///   next_update: date_time_ymd(2024, 06, 17),
 ///   crl_number: SerialNumber::from(1234),
+///   issuing_distribution_point: None,
 ///   revoked_certs: vec![revoked_cert],
 ///   alg: &PKCS_ECDSA_P256_SHA256,
 ///   key_identifier_method: KeyIdMethod::Sha256,
@@ -1404,7 +1409,8 @@ impl NameConstraints {
 }
 
 /// A certificate revocation list (CRL) distribution point, to be included in a certificate's
-/// [distribution points extension](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.13).
+/// [distribution points extension](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.13) or
+/// a CRL's [issuing distribution point extension](https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.5)
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct CrlDistributionPoint {
 	/// One or more URI distribution point names, indicating a place the current CRL can
@@ -1596,6 +1602,11 @@ pub struct CertificateRevocationListParams {
 	pub next_update :OffsetDateTime,
 	/// A monotonically increasing sequence number for a given CRL scope and issuer.
 	pub crl_number :SerialNumber,
+	/// An optional CRL extension identifying the CRL distribution point and scope for a
+	/// particular CRL as described in RFC 5280 Section 5.2.5[^1].
+	///
+	/// [^1]: <https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.5>
+	pub issuing_distribution_point :Option<CrlIssuingDistributionPoint>,
 	/// A list of zero or more parameters describing revoked certificates included in the CRL.
 	pub revoked_certs :Vec<RevokedCertParams>,
 	/// Signature algorithm to use when signing the serialized CRL.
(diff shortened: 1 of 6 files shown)


tests/generic.rs
```diff
@@ -123,4 +123,27 @@ mod test_x509_parser_crl {
...
+		// We should find a CRL DP extension was parsed.
+		let crl_dps = parsed_cert.get_extension_unique(&x509_parser::oid_registry::OID_X509_EXT_CRL_DISTRIBUTION_POINTS)
+			.expect("malformed CRL distribution points extension")
+			.expect("missing CRL distribution points extension");
+
+		// The extension should not be critical.
+		assert!(!crl_dps.critical);
+
+		// TODO(XXX): x509-parser does not parse this extension yet, limiting what we can test from
+		// 	here. See `openssl.rs` for more thorough coverage.
```
cpu, line 147: Noticed while working on the follow-up that this isn't correct. I will rework.

cpu:
> cpu force-pushed the cpu-cert-crl-distrib-points branch from 9e5f117 to 96bf5f8 

* I updated the cert distribution points extension serialization to pull out the code for writing a sequence of URI general names since this can be shared between the cert + CRL extensions.
* I fixed the x509-parser test coverage for the cert CRL distribution points extension after realizing x509-parser does have support for fully parsing this extension (_it is however missing the same for the **CRL** extension - I will follow-up on this upstream_).
* I rolled the CRL issuing distribution point support into this branch as a 2nd commit instead of making a follow-up PR. I think the changes are so closely related it will be easier to review it all in one go.

I also called out one question in the updated PR desc about whether it makes sense to pull in the alpha webpki release that has support for these extensions in order to get better test coverage.

Thanks!

src/lib.rs
```diff
@@ -1641,6 +1719,54 @@ impl CertificateRevocationListParams {
...
+	fn write_der(&self, writer :DERWriter) {
+		// IssuingDistributionPoint SEQUENCE
+		writer.write_sequence(|writer| {
+			// distributionPoint [0] DistributionPointName OPTIONAL
+			write_distribution_point_name_uris(writer.next(), &self.distribution_point.uris);
+
+			// -- at most one of onlyContainsUserCerts, onlyContainsCACerts,
+			// -- and onlyContainsAttributeCerts may be set to TRUE.
+			if let Some(scope) = self.scope {
+				match scope {
```
est31, line 1742: I think this could maybe be deduplicated? As in, the tag number could be stored in a variable and then one only needs one `write_tagged_implicit` call.

tests/generic.rs
```diff
@@ -120,7 +120,68 @@ mod test_x509_parser_crl {
 		let (_, reason_code) = x509_revoked_cert.reason_code().unwrap();
 	 	assert_eq!(reason_code.0, revoked_cert.reason_code.unwrap() as u8);
 
+		// The issuing distribution point extension should be present and marked critical.
+		let issuing_dp_ext = x509_crl.extensions().iter()
+			.find(|ext| ext.oid == x509_parser::oid_registry::OID_X509_EXT_ISSUER_DISTRIBUTION_POINT)
+			.expect("failed to find issuing distribution point extension");
+		assert!(issuing_dp_ext.critical);
+		// TODO(XXX): x509-parser does not yet parse the CRL issuing DP extension for further examination.
```
est31, line 128: No need for `XXX` here, `TODO:` is enough

tests/generic.rs
```diff
@@ -120,7 +120,68 @@ mod test_x509_parser_crl {
...
+			match dp.distribution_point.as_ref().expect("missing distribution point name") {
+				DistributionPointName::FullName(general_names) => general_names.iter(),
+				DistributionPointName::NameRelativeToCRLIssuer(_) => panic!("unexpected name relative to cRL issuer")
+			}
+		}).collect::<Vec<_>>();
+
+		// All of the general names should be URIs.
+		let uris = general_names.iter().map(|general_name| {
+			match general_name {
+				x509_parser::extensions::GeneralName::URI(uri) => uri,
```
est31, line 179: Maybe do `*uri` here, it's a reference to a reference, then you don't need to do `&"foo"` below.

est31 (changes requested):

tests/generic.rs
cpu: Sounds good. Switched to `TODO:`.

tests/generic.rs
cpu: That is nicer, thank you. Switched.

est31 (approved):

est31:
I've merged this. I will work on a PR to split up rcgen into multiple files soon.

cpu:
Ty!

PR #144, opened 2023-08-24
Hold back version of time crate to 0.3.23 to support Rust 1.66

iamjpotts:
Latest version of `time` create has a higher minimum rust version that `rcgen` does.

Limit the version of `time` to the latest version with a minimum rust version compatible with `rcgen`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,24 +67,24 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "botan"
-version = "0.10.4"
+version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9728e14ac65b5b40061c56abdbc3c2ac13e8caca19ac117f26e6caa0932a1175"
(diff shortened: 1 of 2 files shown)


iamjpotts:
Build failure related to parsing a revoked certificate might be fixed by https://github.com/rustls/rcgen/pull/142.

est31:
Needs a rebase, but otherwise looks good.

est31 (approved):
This also does a `cargo update`, which is a good idea as we haven't done one in a long time. In general, doing `cargo update` is not really useful though, as it incurs a cost in terms of requiring us to rebuild a large chunk of dependencies each time there is an update.

rukai:
This change will break my build.
I stumbled across this change and was concerned because I've seen these kinds of version restrictions break things many times.
So I tried pointing my project to the latest rcgen git and sure enough:
image

Please do not introduce fragility into version resolution in order to satisfy MSRV.
If you really want to maintain MSRV you could keep CI passing by relying on your Cargo.lock to pull in the right time version and then expect users to do the same.

I guess the rust-aws-sdk project is just as much at fault here, I'll raise an issue with them too.

est31:
@rukai I think the time maintainers would be interested in your user report: https://github.com/time-rs/time/discussions/592

djc:
@rukai see #148.

PR #145, opened 2023-08-24
project: misc CI improvements.

cpu:
I recommend reviewing this branch commit-by-commit.

There are no functional changes to the code, just CI improvements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
   build-windows:
     runs-on: windows-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v3
       with:
         persist-credentials: false
     - name: Install Rust
@@ -42,7 +42,7 @@ jobs:
             toolchain: stable 7 months ago
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v3
       with:
         persist-credentials: false
     - name: Install Rust
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@master
+      with:
+        persist-credentials: false
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
@@ -41,6 +43,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@master
+      with:
+        persist-credentials: false
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,3 +77,27 @@ jobs:
       run: cargo test --verbose --features x509-parser
     - name: Run cargo doc
       run: cargo doc --all --all-features
+
+  coverage:
+    name: Measure coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools
+      - name: Measure coverage
+        run: cargo llvm-cov --all-features --lcov --output-path ./lcov.info
+      - name: Report to codecov.io
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./lcov.info
+          fail_ci_if_error: false
+          verbose: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: ci
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  merge_group:
+  schedule:
+    - cron: '0 18 * * *'
 
 env:
   RUSTFLAGS: -D warnings
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,20 @@ env:
   RUSTFLAGS: -D warnings
 
 jobs:
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - run: cargo clippy --all-features
+
   build-windows:
     runs-on: windows-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   RUSTFLAGS: -D warnings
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
```
est31, line 1: I feel that for most projects, dependabot has negative utility, because:

* it creates a *lot* of noise in the commit log
* this noise also translates to having to rebuild all dependencies each time there is an update. The situation of rust being so heavy on dependencies is bad enough, no need to make it worse by requiring developers to rebuild dependencies so regularly.
* most updates are not really useful for end users, instead you have stuff like fixing typos of rustdocs. in most instances, old versions are just as good as long as they don't get years out of date.

.github/workflows/ci.yml
```diff
@@ -1,15 +1,52 @@
...
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt
+      - name: Check formatting
+        run: cargo fmt --all -- --check
```
est31, line 28: Please, let's do this in a separate PR. It's already a quite big PR.

est31 (changes requested):

.github/dependabot.yml
djc: I agree that this is an important downside for repositories that commit the `Cargo.lock` file. In our other libraries we don't yet commit the `Cargo.lock` file, but the recommendations from the Cargo team have recently changed/are changing soon to suggest that library projects, too, should commit their lockfiles. We'd ideally want to restrict Dependabot to semver-incompatible updates only. Unfortunately I don't think this is an option for their Cargo ecosystem plugin (although AIUI their npm plugin does support it). The other option that I've used in the past is to manually put in place a list of ignores, like this:

```
    ignore:
      - dependency-name: ahash
        versions:
          - "> 0.7.2, < 0.8"
      - dependency-name: anyhow
        versions:
          - "> 1.0.40, < 2"
```

This of course has the downside of having to maintain the ignore list, but if semver-incompatible updates don't happen to often for rcgen's dependencies (which seems like a reasonable assumption) that might not be a big burden.

.github/workflows/ci.yml
```diff
@@ -11,6 +11,20 @@ env:
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - run: cargo clippy --all --all-features
```
djc, line 42: It doesn't actually deny warnings? This has `--all` which I don't normally use, do you mean `--all-targets`?

.github/workflows/ci.yml
```diff
@@ -77,3 +77,26 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools
+      - name: Measure coverage
```
djc, line 111: IIRC CodeCov is pretty flaky unless you provide a token.

.github/workflows/ci.yml
est31: `--all` is equivalent to `--workspace` (but shorter). It was deprecated in favour of latter because of the confusion potential with `--all-targets` :).

I think `RUSTFLAGS` handles the denying of warnings already, so there is no need to also ask it in the CLI.

djc: But `--workspace` is also the default already, so might as well omit it IMO.

.github/dependabot.yml
est31: I started committing Cargo.lock due to ring's practice of yanking old releases, as explained in fb7343490ed00396fe3abfecbe557606d6191d64 . I wanted to be able to check out an older release, and compile it. If some crate has all semver compatible releases yanked for the version you depend on, that doesn't work.

To expand on the noise in the commit log point: sadly dependabot has no feature to unify updates, as in, have one commit that does the equivalent of `cargo update` maybe once per month or so: it has to make a separate commit/PR/etc for every single update, even of minor semver updates (semver incompatible ones I get, but not the patch version updating ones).

So every single release of any transitive dependency causes one commit in my log, which is bloating it up to an extreme and undeserved degree. Now, if there is security updates it might make sense, but for rcgen, a library, it makes no sense at all because users have their own `Cargo.lock`. All the `Cargo.lock` in rcgen does is aid development, and even if rcgen were a binary crate, the main point in time when `Cargo.lock` should be up to date is shortly before a release.

There is other arguments to have up-to-date `Cargo.lock` files, like to verify that there is no breakage with using newest versions of some crates. Obviously, we want the newest released versions to work with our library, but on the other hand, we also want older versions to work. This doesn't mean that we should make it a priority to keep the versions in rcgen's Cargo.lock as old as possible, I just think that this is a good counter argument to making using the latest version a priority.

To ensure that the latest versions of crates work, we could use a CI job that runs `cargo update` before running, maybe re-using one of the CI jobs for existing platforms.

For `cargo-udeps`, people have also brought up dependabot, and we agreed upon merging updates of especially security relevant dependencies as a compromise (if there is security reasons for the update or not). But I noted also to them that for the users who use the `cargo-udeps` release from crates.io, there is zero effect to what is going on on `cargo-udeps`'s github. The experience of enabling dependabot there made me like it even less because openssl often updates multiple of their crates in succession, and dependabot makes multiple PRs for each of the openssl crates, causing tons of noise. I suppose some people use `cargo-udeps` from git though and have a job running that checks for updates of the `master` branch of `cargo-udeps`. At least that's the only group of people (outside of people developing `cargo-udeps`) who would benefit from updates that are not immediately before a crates.io release.

djc: To be clear, I think @cpu put this in here in order to generate notifications about semver-incompatible updates (as you would get without a committed lockfile). I agree that automated lockfile bumps aren't helpful.

cpu: Is there any opposition to use Dependabot for CI actions? The downsides being raised seem particular to using the Cargo integration.

.github/workflows/ci.yml
cpu: Removed from this branch.

.github/workflows/ci.yml
cpu: I've omitted it from the clippy task added in this branch.

.github/workflows/ci.yml
cpu: Added a token.

.github/dependabot.yml
cpu: I've removed the Dependabot addition. Pending input on using it for CI actions I'll consider a follow-up PR.

.github/workflows/ci.yml
cpu: https://github.com/rustls/rcgen/pull/146

.github/dependabot.yml
est31: Yeah for actions it is good to have dependabot I think, unless the version updates are extremely regular, because otherwise it's very easy to get the versions out of date. That's also why I generally prefer doing `@master` instead of pinning the version. But given that one of the deps has set this insecure default, it makes sense to pin it. Adding it in a followup would be better also, yeah.

cpu: > unless the version updates are extremely regular

IME they're quite infrequent.

est31 (approved):

djc (approved):

Issue #85, opened 2022-06-16
Open to a formatting with rustfmt?


clangenb:
Formatting with `cargo fmt` has been widely adopted by now. I was wondering if you were open to introduce formatting with rustfmt?

You could introduce a `.rustfmt.toml`, which almost maintains the current formatting config.

est31:
Not at the moment, thank you. Last time I checked, rustfmt still mixes tabs and spaces, but I want this to have tabs only until the first non-whitespace character on each line, after which it should be spaces only.

clangenb:
Oh, I see, I was not aware of this. Thanks for explaining!

And I double-checked: You are right, rustfmt still mixes tabs and spaces.

cpu:
Is there any chance you might be willing to reconsider this decision? :angel: 

I don't want to start a heated debate but from the perspective of an outside contributor the current manual formatting is a challenge. Having to think about formatting adds a fair bit of cognitive overhead + review time that could be avoided entirely with `rustfmt`. 

Perhaps use of this crate has grown to the point where it makes sense to prioritize consistency with the broader Rust ecosystem to facilitate ease of outside contributions?

est31:
Hmm yeah originally in most releases, most of the commits have come from me with a few commits from outside contributors. The 0.11.0 release was different though as most commits have come from contributors. Maybe I can write a small script that calls rustfmt internally that un-does the tab/space mixing...

cpu:
Thanks for giving it some consideration.

> Maybe I can write a small script that calls rustfmt internally that un-does the tab/space mixing...

That sounds like an OK middle ground. Personally I'd probably just embrace `rustfmt` and its warts instead of having to maintain additional tooling but maybe I've got brain worms from too many years of universally applying `go fmt`.

PR #146, opened 2023-08-25
project: apply cargo fmt universally, verify in CI.

cpu:
Self-explanatory. Applies `cargo fmt` universally, adds a CI job to ensure consistent formatting going forward. No functional changes.


djc (approved):
To be explicit, here's why I think this is a good idea:

1. The default `cargo fmt` settings represent idiomatic style, that is used across most Rust projects
2. Automatic formatting tooling makes life easier for contributors
3. Style enforcement in CI avoids (some) additional review roundtrips due to stylistic concerns

cpu:
> To be explicit, here's why I think this is a good idea

Thanks for adding motivating context. I should have done that in my PR description :+1: Your points echo the feelings I expressed [in this comment](https://github.com/rustls/rcgen/issues/85#issuecomment-1596156604).

est31:
Btw, would it be possible to remove these coverage comments? they are put on every PR, even ones that don't touch anything related to tests. I think it's cool to have coverage reports you can browse online and look at graphs, but I don't think one needs to have a comment for each PR on how it changes the doc coverage.

As for autoformatters, I think there are some options that make cargo fmt nicer, I will make some experiments and will then propose something.

cpu:
> Btw, would it be possible to remove these coverage comments?

I'll look into that.

> As for autoformatters, I think there are some options that make cargo fmt nicer, I will make some experiments and will then propose something.

@est31 Does this PR need to wait for that adjustment? The benefit of machine formatting is that if you change the rules of the formatter, you can run it again.

djc:
> > As for autoformatters, I think there are some options that make cargo fmt nicer, I will make some experiments and will then propose something.
> 
> @est31 Does this PR need to wait for that adjustment? The benefit of machine formatting is that if you change the rules of the formatter, you can run it again.

@est31 friendly ping on this? Would be nice for further changes to get the formatting done first.

est31 (approved):

est31:
It's a pity that `overflow_delimited_expr` isn't stable. Idk about you but I like former way more:

```
-		assert_eq!(uris, &[
-			"http://example.com/crl.der",
-			"http://crls.example.com/1234",
-			"ldap://example.com/crl.der"
-		]);
+		assert_eq!(
+			uris,
+			&[
+				"http://example.com/crl.der",
+				"http://crls.example.com/1234",
+				"ldap://example.com/crl.der"
+			]
+		);
```

iamjpotts:
> It's a pity that `overflow_delimited_expr` isn't stable. Idk about you but I like former way more:
> 
> ```
> -		assert_eq!(uris, &[
> -			"http://example.com/crl.der",
> -			"http://crls.example.com/1234",
> -			"ldap://example.com/crl.der"
> -		]);
> +		assert_eq!(
> +			uris,
> +			&[
> +				"http://example.com/crl.der",
> +				"http://crls.example.com/1234",
> +				"ldap://example.com/crl.der"
> +			]
> +		);
> ```

I like the latter way, but I'm also glad that `rustfmt` isn't `black`

PR #147, opened 2023-08-25
ci: add dependabot config for GitHub actions.

cpu:
This will let Dependabot keep the CI GitHub action versions up to date.

For the time being we choose not to configure Dependabot for the cargo ecosystem since its generally more noise/work to keep up with.


est31 (approved):
Let's hope it doesn't catch all the patch versions

PR #149, opened 2023-08-29
project: codecov comments only for changes.

cpu:
This commit adds a `codecov.yml` configuration file that instructs Codecov.io to only add a comment to pull requests when coverage changes.

See the [codecov docs](https://docs.codecov.com/docs/pull-request-comments) for more information.

If this is still too noisy for our preferences we can adjust the config to `comment: false` in a follow-up.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+comment:
+  require_changes: true


djc (approved):

est31 (approved):
Thanks for the PR, also good strategy to maybe always disable the comment.

PR #153, opened 2023-09-09
misc. cargo doc improvements.

cpu:
Small improvements to the existing rustdocs. I recommend reviewing commit-by-commit.

* Fixing redundant link targets in `src/key_pair.rs` flagged by beta rust doc.
* Fixing a typo in `src/key_pair.rs.
* Updating the permitted/excluded subtree name constraint docs to be consistent in their descriptions.
* Applying consistent comment standards to the crate internal `src/oid.rs` comments.
* Adding rustdoc comments to `CertificateSigningRequest`, removing an allow for missing docs.
* Pulling out a separate rust doc CI job, testing with stable, beta, nightly and MSRV toolchains on one platform.
* Adding `--document-private-items` to the CI rust doc job to catch crate-internal doc errs.
* Enabling `doc_auto_cfg` to get automatic feature requirement display in the doc.rs docs
* Removing redundant manually maintained feature requirement comments in favour of `doc_auto_cfg`'s automatic support.


src/oid.rs
```diff
@@ -1,74 +1,68 @@
-/// pkcs-9-at-extensionRequest in RFC 2985
+/// pkcs-9-at-extensionRequest in <https://www.rfc-editor.org/rfc/rfc2985#appendix-A>
 pub const OID_PKCS_9_AT_EXTENSION_REQUEST: &[u64] = &[1, 2, 840, 113549, 1, 9, 14];
 
-/// id-at-countryName in RFC 5280
+/// id-at-countryName in <https://www.rfc-editor.org/rfc/rfc5280#appendix-A>
```
est31, line 4: How is this rendered? Having the full URL visible? Then maybe a link like ```[`RFC 5280`](https://www.rfc-editor.org/rfc/rfc5280#appendix-A)``` would be better. Note how the URLs were non-doc comments.

src/lib.rs
```diff
@@ -1259,13 +1256,11 @@ impl CertificateParams {
 /// (only relevant for CA certificates)
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct NameConstraints {
-	/// If non-empty, a whitelist of subtrees that the
-	/// domain has to match.
+	/// If non-empty, an allow list of subtrees that the domain has to match.
 	pub permitted_subtrees: Vec<GeneralSubtree>,
-	/// A list of excluded subtrees.
+	/// If non-empty, a forbid-list of excluded subtrees that the domain must not match.
```
est31, line 1261:
```suggestion
	/// If non-empty, a list of excluded subtrees that the domain must not match.
```

src/lib.rs
```diff
@@ -1259,13 +1256,11 @@ impl CertificateParams {
 /// (only relevant for CA certificates)
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct NameConstraints {
-	/// If non-empty, a whitelist of subtrees that the
-	/// domain has to match.
+	/// If non-empty, an allow list of subtrees that the domain has to match.
```
est31, line 1259:
```suggestion
	/// If non-empty, a list of subtrees that the domain has to match.
```

est31 (approved):
Okay except comments.

.github/workflows/ci.yml
```diff
@@ -41,6 +40,26 @@ jobs:
           components: clippy
       - run: cargo clippy --all-features
 
+  rustdoc:
+    name: Check for documentation errors
```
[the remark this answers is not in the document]
cpu: Renamed.

src/oid.rs
cpu: Switched to titled links.

src/lib.rs
cpu: Updated.

src/lib.rs
cpu: Updated.

cpu:
> Okay except comments.

Thanks for the review. I believe I addressed all of your comments.

src/lib.rs
```diff
@@ -1259,13 +1259,11 @@ impl CertificateParams {
 /// (only relevant for CA certificates)
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct NameConstraints {
-	/// If non-empty, a whitelist of subtrees that the
-	/// domain has to match.
+	/// If non-empty, a list of subtrees that the domain has to match.
 	pub permitted_subtrees: Vec<GeneralSubtree>,
-	/// A list of excluded subtrees.
+	/// If non-empty, a list of excluded subtrees that the domain must not match.
```
djc, line 1261: I'd rather remove the "If non-empty," prefix from `permitted_subtrees`? At least I'm not sure what value it adds. If you're going to add "that the domain must not match", maybe remove "excluded"? Otherwise this feels like it is trying to say the same thing twice (I agree that the prose explanation is more clear than merely using "excluded").

djc (approved):
Nice!

src/lib.rs
cpu: Changed to remove `If non-empty` prefixes, and the "excluded" from the excluded subtrees comment.

Thanks, agreed it reads better :+1:

cpu:
@est31 Just wanted to mention that I enabled the "require merge queue" setting for the repository. It's convenient to be able to queue PRs to automatically merge when CI passes (after the required reviews are in) vs. having to babysit their progress. As usual, repository administrators can bypass this requirement.

Issue #122, opened 2023-05-30
X509v3 extensions from certificate not transferred to CSR


nicopal:
Hi,

After creating a certificate with several X509v3 extensions (e.g. X509v3 Extended Key Usage: TLS Web Client Authentication), I create a certificate signing request using e.g. 
`certificate.serialize_request_der()`

However, the certificate signing request **does not** contain the extensions. This is in line with an earlier issue in OpenSSL, where extensions from the certificate were not transferred to the CSR and vice versa. This appears to have been fixed by now: https://github.com/openssl/openssl/issues/10458

Do you plan to add a feature to copy X509v3 extensions from the certificate into the CSR?

Thanks.

cpu:
EKU params are being passed into CSRs as of https://github.com/rustls/rcgen/pull/264 

We can leave this open for a more general solution but I thought it would be helpful for you to know this since extended key usages was the usecase mentioned in the original issue text.

PR #154, opened 2023-09-10
lib: use ext helper in more places, test custom exts.

cpu:
This branch updates a couple places that were writing x509 extensions by hand to use the existing `write_x509_extension` helper instead. 

Since custom extensions are a bit trickier a unit test for custom extensions in CSRs and certificates is added before refactoring the existing code to use the helper. This helps ensure the change is a no-op from the perspective of users.

With the updated test, it was possible to demonstrate a bug similar to https://github.com/rustls/rcgen/issues/122. The CSR serialization logic requiring SANs to be present in order to emit the custom extensions into the CSR PKCS9 extension request attribute. This branch fixes the issue.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -999,11 +999,7 @@ impl CertificateParams {
 
 						// Write standard key usage
 						if !self.key_usages.is_empty() {
-							writer.next().write_sequence(|writer| {
-								let oid = ObjectIdentifier::from_slice(OID_KEY_USAGE);
-								writer.next().write_oid(&oid);
-								writer.next().write_bool(true);
-
+							write_x509_extension(writer.next(), OID_KEY_USAGE, true, |writer| {
 								let mut bits: u16 = 0;
 
 								for entry in self.key_usages.iter() {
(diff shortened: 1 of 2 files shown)


est31 (approved):
I only gave this a cursory glance due to limited time but the general idea of the PR is good, thanks. Maybe it would be good to get another review?

cpu:
> Maybe it would be good to get another review?

Sounds good. @djc when you have some time I'd appreciate a review pass.

FWIW I'm also working on a subsequent refactoring of extension handling to fix #122 - I'll open that as a follow-up shortly.

cpu:
> FWIW I'm also working on a subsequent refactoring of extension handling to fix https://github.com/rustls/rcgen/issues/122 - I'll open that as a follow-up shortly.

General idea is in https://github.com/cpu/rcgen/pull/1 I haven't done a self-review pass yet and I'd like to add more tests.

src/lib.rs
```diff
@@ -911,7 +911,7 @@ impl CertificateParams {
 			// Write extensions
 			// According to the spec in RFC 2986, even if attributes are empty we need the empty attribute tag
 			writer.next().write_tagged(Tag::context(0), |writer| {
-				if !subject_alt_names.is_empty() {
+				if !subject_alt_names.is_empty() || !custom_extensions.is_empty() {
```
djc, line 914: Given the commit message text about a test failure, should this commit be moved around to appear before some of the other commits in this PR?

djc (approved):

djc:
Nice!

src/lib.rs
cpu: I made the test change + fix in the same commit so they would all build green. The commit message text describes a manually created intermediate step.

PR #160, opened 2023-09-20
Release 0.11.2

### LINKED ISSUE #156: Publish a 0.11.2 release
I think it's a good time to cut a `0.11.2` release - I've done a scan of the contents of `main` since the last tag and don't believe anything consists of a breaking API change (but it would be helpful for someone to verify that line of thinking w.r.t the new CRL support types). This would also let a few Rustls projects remove `git` pins they're carrying for rcgen w/ CRL support.

The immediate changes I have in mind for addressing a few outstanding issues in the issue tracker are likely to require breaking changes (e.g. because they'll change what extensions are included in a CSR in a way that might break assumptions downstream code has made based on previous API contracts).

est31:
Closes #156.

This is the first release of rcgen at its new home in the `rustls` github organization. I want to use this opportunity to thank @cpu , @djc and the other rustls members for their help with maintaining rcgen and letting rcgen join the `rustls` organization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 
 # Changes
 
-## Release 0.11.2 - Pending Release
+## Release 0.11.2 - September 21, 2023
 
-- Project repository is now part of the [rustls](https://github.com/rustls) organization.
+- `rcgen` has joined the umbrella of the [rustls](https://github.com/rustls) organization.
 - Support for retrieving signature algorithm from `KeyPair`s. Contributed by [tindzk](https://github.com/tindzk).
 - Fix for writing certificate signing requests (CSRs) with custom extensions from parameters without subject alternative names.
 - Support for certificate CRL distribution points extension.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -498,7 +498,7 @@ dependencies = [
 
 [[package]]
 name = "rcgen"
-version = "0.11.1"
+version = "0.11.2"
 dependencies = [
  "botan",
  "openssl",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rcgen"
-version = "0.11.1"
+version = "0.11.2"
 description = "Rust X.509 certificate generator"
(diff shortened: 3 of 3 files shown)


cpu (approved):
Thanks! 🎉🥳

est31:
The CI failure is legit... seems due to a problem with the new webpki release [which added the budget param](
https://diff.rs/rustls-webpki/0.101.4/0.101.5/src/crl.rs) to a public function?

```

error[E0061]: this method takes 3 arguments but 2 arguments were supplied
   --> tests/webpki.rs:542:4
    |
542 |         .verify_signature(&[&webpki::ECDSA_P256_SHA256], &raw_spki.value())
    |          ^^^^^^^^^^^^^^^^-------------------------------------------------- an argument of type `&mut webpki::verify_cert::Budget` is missing
    |
note: method defined here
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-webpki-0.101.5/src/crl.rs:42:8
    |
42  |     fn verify_signature(
    |        ^^^^^^^^^^^^^^^^
help: provide the argument
    |
542 |         .verify_signature(&[&webpki::ECDSA_P256_SHA256], &raw_spki.value(), /* &mut webpki::verify_cert::Budget */)
    |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

cpu:
> The CI failure is legit... seems due to a problem with the new webpki release [which added the budget param](https://diff.rs/rustls-webpki/0.101.4/0.101.5/src/crl.rs) to a public function?

@est31 Looks like I broke semver backporting that to 0.101.x without realizing it :crying_cat_face: I will work on cutting a new point release that undoes that mistake, but I won't be able to do it until ~tomorrow. Thanks for flagging the issue.

cpu:
> Looks like I broke semver backporting that to 0.101.x without realizing it

This is fixed in [0.101.6](https://github.com/rustls/webpki/releases/tag/v%2F0.101.6).

PR #166, opened 2023-10-03
Upgrade to `ring` `0.17`

thomaseizinger:
Reboot of https://github.com/rustls/rcgen/pull/98.

This PR is a breaking change because the `RcgenError::RingKeyRejected` variant is changed AND because `ring` is part of the public API via the `From` impl for `RcgenError`.

### DIFF
--- a/src/key_pair.rs
+++ b/src/key_pair.rs
@@ -106,6 +106,7 @@ impl KeyPair {
 		pkcs8: &[u8],
 		alg: &'static SignatureAlgorithm,
 	) -> Result<Self, Error> {
+		let rng = &SystemRandom::new();
 		let pkcs8_vec = pkcs8.to_vec();
 
 		let kind = if alg == &PKCS_ED25519 {
@@ -114,11 +115,13 @@ impl KeyPair {
 			KeyPairKind::Ec(EcdsaKeyPair::from_pkcs8(
 				&signature::ECDSA_P256_SHA256_ASN1_SIGNING,
 				pkcs8,
+				rng,
 			)?)
 		} else if alg == &PKCS_ECDSA_P384_SHA384 {
 			KeyPairKind::Ec(EcdsaKeyPair::from_pkcs8(
 				&signature::ECDSA_P384_SHA384_ASN1_SIGNING,
 				pkcs8,
+				rng,
 			)?)
 		} else if alg == &PKCS_RSA_SHA256 {
 			let rsakp = RsaKeyPair::from_pkcs8(pkcs8)?;
@@ -146,14 +149,15 @@ impl KeyPair {
 	pub(crate) fn from_raw(
 		pkcs8: &[u8],
 	) -> Result<(KeyPairKind, &'static SignatureAlgorithm), Error> {
+		let rng = SystemRandom::new();
 		let (kind, alg) = if let Ok(edkp) = Ed25519KeyPair::from_pkcs8_maybe_unchecked(pkcs8) {
 			(KeyPairKind::Ed(edkp), &PKCS_ED25519)
 		} else if let Ok(eckp) =
-			EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P256_SHA256_ASN1_SIGNING, pkcs8)
+			EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P256_SHA256_ASN1_SIGNING, pkcs8, &rng)
 		{
 			(KeyPairKind::Ec(eckp), &PKCS_ECDSA_P256_SHA256)
 		} else if let Ok(eckp) =
-			EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P384_SHA384_ASN1_SIGNING, pkcs8)
+			EcdsaKeyPair::from_pkcs8(&signature::ECDSA_P384_SHA384_ASN1_SIGNING, pkcs8, &rng)
 		{
 			(KeyPairKind::Ec(eckp), &PKCS_ECDSA_P384_SHA384)
 		} else if let Ok(rsakp) = RsaKeyPair::from_pkcs8(pkcs8) {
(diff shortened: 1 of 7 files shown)


thomaseizinger:
All commits pass clippy and formatting. Let me know if you want any of these to be split out into other PRs!

src/key_pair.rs
```diff
@@ -55,19 +55,19 @@ impl KeyPair {
 	/// Parses the key pair from the DER format
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
-	pub fn from_der(der: &[u8]) -> Result<Self, RcgenError> {
-		Ok(der.try_into()?)
+	pub fn from_der(der: &[u8], rng: &dyn SecureRandom) -> Result<Self, RcgenError> {
```
iamjpotts, line 58: This extra required parameter unnecessarily complicates the api for most use cases.

I suggest keeping the original method signature, and providing an alternate method where the random number generator can be provided. Have the method with the original signature use a reasonable default.

Also, why is it `&dyn` ?

iamjpotts (commented):
See discussion in https://github.com/rustls/rcgen/pull/98 - was that drawn to a consensus wrt the additional parameter?

src/key_pair.rs
thomaseizinger: > Also, why is it `&dyn` ?

Because it is a trait and therefore has to be passed by reference otherwise it is not `Sized` :)

> This extra required parameter unnecessarily complicates the api for most use cases.

Most usecases use `rcgen::generate_simple_self_signed` or `Certificate::from_params`, right? I'd argue that parsing keypairs etc is already low-level functionality for advanced usecases. If we duplicate all of these APIs with and without an RNG parameter, we end up with a lot of noise.

djc: Different question: is *ring* already part of the public API of the crate?

src/key_pair.rs
```diff
@@ -143,20 +144,22 @@ impl KeyPair {
 		})
 	}
 
-	pub(crate) fn from_raw(
-		pkcs8: &[u8],
+	pub(crate) fn from_raw<'b>(
+		pkcs8: impl Into<Cow<'b, [u8]>>,
```
djc, line 151: How is this helpful? IMO this kind of trick is usually good for public APIs, but for a private API I'd probably just take `&[u8]` -- it doesn't look like anything is making good use of the possibility of `Cow::Owned` anyway.

src/key_pair.rs
thomaseizinger: I thought it was but upon checking again, it might not actually be!

thomaseizinger: > I thought it was but upon checking again, it might not actually be!

That changes the game entirely. Any suggestions for how we can make `KeyPair`'s deterministic? I guess arguably, that should probably be done in a different PR.

djc: (Note that the use of ring error types in `rcgen::Error` also makes it public.)

I'm not saying making it a part of the public API is necessarily bad (not sure how intentional it was to keep it private), but just making sure we make this an explicit choice.

thomaseizinger: > (Note that the use of ring error types in `rcgen::Error` also makes it public.)

Yeah good call. I'll make it private again.

djc:
(FWIW, you had a nice commit history before but now there's a bunch of back and forth there... maybe use `rebase -i` to skip the commits that introduce the `Cow` and introduce ring to the public API, instead of reverting them with later commits?)

thomaseizinger:
> (FWIW, you had a nice commit history before but now there's a bunch of back and forth there... maybe use `rebase -i` to skip the commits that introduce the `Cow` and introduce ring to the public API, instead of reverting them with later commits?)

Sure! I did see that you are not squash-merging in here so I figured you'd care about the commit history :)

thomaseizinger:
Sorry about the closing stuff, I fuffed around with rebase too much :sweat_smile: 

This PR is now entirely focused on upgrading `ring`.

src/key_pair.rs
```diff
@@ -211,23 +216,25 @@ impl TryFrom<Vec<u8>> for KeyPair {
 
 impl KeyPair {
 	/// Generate a new random key pair for the specified signature algorithm
-	pub fn generate(alg: &'static SignatureAlgorithm) -> Result<Self, RcgenError> {
-		let system_random = SystemRandom::new();
+	pub fn generate(
+		alg: &'static SignatureAlgorithm,
+		rng: &dyn SecureRandom,
```
djc, line 221: This is still exposing ring as part of the public API.

est31: We could create our own `SecureRandom` trait, which is a copy of ring's. Then we could make a (private) struct wrapper of that, which then implements ring's `SecureRandom` trait. We would then take anything that impls our `SecureRandom` trait.

src/lib.rs
```diff
@@ -1492,15 +1492,16 @@ fn write_general_subtrees(writer: DERWriter, tag: u64, general_subtrees: &[Gener
 impl Certificate {
 	/// Generates a new certificate from the given parameters.
 	///
-	/// If there is no key pair included, then a new key pair will be generated and used.
+	/// If there is no key pair included, then a new key pair will be randomly generated and used.
+	/// If you want to control the [`KeyPair`] or the randomness used to generate it, set it ahead of time before calling this function.
```
cpu, line 1496: I think this might be a touch more user friendly if we explicitly link to the params field in question:

```suggestion
	/// If you want to control the [`KeyPair`] or the randomness used to generate it, set the [`CertificateParams::key_pair`]
	/// field ahead of time before calling this function.
```

src/key_pair.rs
thomaseizinger: I'll defer it to another PR.

est31: @thomaseizinger fine by me!

est31 (approved):
Okay except for the two comments by djc and cpu, where the one by djc can also be addressed in a subsequent PR (as long as it's done before a release).

thomaseizinger:
> Okay except for the two comments by djc and cpu, where the one by djc can also be addressed in a subsequent PR (as long as it's done before a release).

Fixed! It is a breaking change so you might want to release a non-breaking one before merging this one.

PR #167, opened 2023-10-03
Rename `Error` type

thomaseizinger:
Error types of upstream crates often need to be referred to in several places like composite `Error` enums. Having the type just named `Error` and exported at the root makes this really easy by just fully-qualifying it, in our case `rcgen::Error`. This saves the user from having to add an additional `use rcgen::` line and is unambiguous.

We provide a type-alias to make the rename a non-breaking change, meaning it can be released immediately before other breaking changes like the [upgrade to `ring 0.17`](https://github.com/rustls/rcgen/pull/166) for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 # Changes
 
+## Unreleased
+
+- Rename `RcGenError` to `Error` to avoid stuttering when used fully-qualified via `rcgen::`.
+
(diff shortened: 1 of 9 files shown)


djc (approved):
Thanks, I think this is the idiomatic way to do things these days!

cpu:
Thank you, this is a nice change :+1:

PR #168, opened 2023-10-03
Add caching to CI

thomaseizinger:
`botan-sys` is quite a heavy dependency to build. To accelerate CI times, add some caching of build artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,7 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
         toolchain: stable
+    - uses: Swatinem/rust-cache@v2
     - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
     - run: vcpkg install openssl:x64-windows-static-md
     - name: Run cargo check --all
@@ -111,6 +112,7 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.toolchain }}
+    - uses: Swatinem/rust-cache@v2
     - name: Run cargo check --all
       run: cargo check --all
     - name: Run the tests
@@ -134,6 +136,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: llvm-tools
+      - uses: Swatinem/rust-cache@v2
       - name: Measure coverage
         run: cargo llvm-cov --all-features --lcov --output-path ./lcov.info
       - name: Report to codecov.io


djc (approved):

est31:
I wonder, is it possible to do caching without using dependencies? From what I can tell, you can do it with just 1-2 lines as well. The issue about dependencies is that you have to upgrade them and they also sometimes can become unmaintained.

djc:
The Swatinem action has been around for a long time and is widely used (IIRC it's already in use in some of the other rustls repositories?). For updates you can enable Dependabot for the Actions your workflows depend on.

The actions-rs stuff was painful, but I feel like caching ends up complicated enough that taking a dependency for this is probably the right solution.

cpu:
> (IIRC it's already in use in some of the other rustls repositories?)

Hmm, I don't think so? I grepped across my local checkouts and didn't come up with any YAML hits. It looks like Quinn, H3 and Reqwest are, but nothing in the Rustls org.

I had been [playing around](https://github.com/cpu/rcgen/commit/9ab70dfde12d4b6f817d811cbde8f15bfb02ba8e) with caching in an old CI cleanup branch using `actions/cache@v3` based on the [upstream docs for Cargo projects](https://github.com/actions/cache/blob/main/examples.md#rust---cargo). I think I paused on that because it was hard to demonstrate speedup when the `main` branch hadn't populated the cache based on how GitHub shared caches between forks/branches (IIRC).

Would using the 1st party Cache action dependency be more palatable to you @est31 ?

cpu:
> For updates you can enable Dependabot for the Actions your workflows depend on.

FWIW that's already [in place](https://github.com/rustls/rcgen/blob/e468600e068920d9a8cddb359f4e42e94cbc1ae0/.github/dependabot.yml).

est31:
> I had been [playing around](https://github.com/cpu/rcgen/commit/9ab70dfde12d4b6f817d811cbde8f15bfb02ba8e) with caching in an old CI cleanup branch using actions/cache@v3 based on the [upstream docs for Cargo projects](https://github.com/actions/cache/blob/main/examples.md#rust---cargo).

Mhh that looks pretty good. Of course, if a pull request is not sped up by it while using `Swatinem`'s crate speeds up PR builds, then we should use latter, but generally I'd prefer using something from github instead.

cpu:
>  Of course, if a pull request is not sped up by it while using Swatinem's crate speeds up PR builds, then we should use latter,

I'll have to page the context back in, but I believe the issue wasn't the action being used, but the fact that the `main` convergence branch cache isn't populated until we merge one of the approaches. Afterwards (based on my reading [of the docs](https://github.com/actions/cache#cache-scopes)) it will be available to other branches.

est31:
Okay @cpu let's go with your branch then, try it out on main, and if it has problems, then we can always come back to this PR. It is a little bit noisier but is lower in terms of supply chain risk, which I like.

cpu:
> Try it out on main, and if it has problems, then we can always come back to this PR

PTAL: https://github.com/rustls/rcgen/pull/172 

Just to be explicit; I agree strongly that speeding up CI is a worthwhile direction no matter which approach we end up taking. Thanks for getting the ball rolling @thomaseizinger

thomaseizinger:
> I wonder, is it possible to do caching without using dependencies? From what I can tell, you can do it with just 1-2 lines as well. The issue about dependencies is that you have to upgrade them and they also sometimes can become unmaintained.

FWIW, I have very good experiences with this action. We use it extensively in quite a big workspace here: https://github.com/libp2p/rust-libp2p. The setup is quite elaborate in that we only write to caches on master: https://github.com/libp2p/rust-libp2p/blob/master/.github/workflows/cache-factory.yml and then use them in PRs: https://github.com/libp2p/rust-libp2p/blob/399eaa39156f756956f38fab20083293da826fdd/.github/workflows/ci.yml#L44

I've also add interactions with the maintainer and they were very responsive :)

I don't mind which way you go, personally, I just prefer not to maintain these configs if I can do it with a one-liner. For example, the action also automatically cleans your target directory of workspace artifacts which are usually useless.

est31:
Closing in favour of #172. As I've said above, if the approach chosen by #172 turns out to cause major issues or limitations, we can always reconsider this one, but for now i'd like to go with #172. @thomaseizinger thanks for getting the ball rolling!

PR #170, opened 2023-10-03
Refactor `KeyPair` constructor

thomaseizinger:
The `TryFrom` impl doesn't communicate that these bytes need to be DER-encoded. Forcing the user to use the constructor makes this obvious.

This is a breaking change because it removes trait implementations.

### DIFF
--- a/src/key_pair.rs
+++ b/src/key_pair.rs
@@ -55,7 +55,7 @@ impl KeyPair {
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
 	pub fn from_der(der: &[u8]) -> Result<Self, Error> {
-		Ok(KeyPair::from_raw(der)?)
+		Ok(KeyPair::guess_kind_from_der(der)?)
 	}
 	/// Returns the key pair's signature algorithm
 	pub fn algorithm(&self) -> &'static SignatureAlgorithm {
@@ -66,7 +66,7 @@ impl KeyPair {
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)?;
 		let private_key_der: &[_] = private_key.contents();
-		Ok(KeyPair::from_raw(private_key_der)?)
+		Ok(KeyPair::guess_kind_from_der(private_key_der)?)
 	}
 
 	/// Obtains the key pair from a raw public key and a remote private key
@@ -142,7 +142,7 @@ impl KeyPair {
 		})
 	}
 
-	pub(crate) fn from_raw(pkcs8: &[u8]) -> Result<KeyPair, Error> {
+	pub(crate) fn guess_kind_from_der(pkcs8: &[u8]) -> Result<KeyPair, Error> {
 		let (kind, alg) = if let Ok(edkp) = Ed25519KeyPair::from_pkcs8_maybe_unchecked(pkcs8) {
 			(KeyPairKind::Ed(edkp), &PKCS_ED25519)
 		} else if let Ok(eckp) =
--- a/src/key_pair.rs
+++ b/src/key_pair.rs
@@ -3,7 +3,6 @@ use pem::Pem;
 use ring::rand::SystemRandom;
 use ring::signature::KeyPair as RingKeyPair;
 use ring::signature::{self, EcdsaKeyPair, Ed25519KeyPair, RsaEncoding, RsaKeyPair};
-use std::convert::TryFrom;
 use std::fmt;
 use yasna::DERWriter;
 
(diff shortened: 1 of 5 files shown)


djc (commented):
I'd squash these commits together. Maybe `KeyPair::from_raw()` should be called `from_der()`, too?

est31:
If what @djc said is implemented, this can be merged I think.

est31:
(also needs a rebase)

thomaseizinger:
> I'd squash these commits together. Maybe `KeyPair::from_raw()` should be called `from_der()`, too?

I've named it `guess_kind_from_der` and refactored it to use early-returns. Let me know what you think.

thomaseizinger:
> I don't find all the refactoring very convincing because it still ends up with quite a bit of duplication.

Fair enough. I am surprised you mention duplication as the reason because the current version has _less_ duplication than what is in `main`. At least when it comes to the semantic coupling of the different algorithm identifiers. Those are in a single place now whereas before they were duplicated between two functions.

> Would be happy to merge something closer to the original form though, I'd maybe rename `guess_kind_from_der()` to `try_from_der()` which feels more idiomatic (and more concise).

I went with `guess` because I deemed it important to convey to the caller that this function isn't necessarily failing because the bytes are malformed but simply because we might not support a particular combination. Additionally, if you pass malformed bytes, the error is not going to tell you what is wrong because we are hiding this by simply trying the next one. To me, that is very different from what I'd expect from a typical `try_from` function, hence the special name.

cpu:
> I went with guess because I deemed it important to convey to the caller that this function isn't necessarily failing because the bytes are malformed but simply because we might not support a particular combination. Additionally, if you pass malformed bytes, the error is not going to tell you what is wrong because we are hiding this by simply trying the next one. To me, that is very different from what I'd expect from a typical try_from function, hence the special name.

By this logic shouldn't `KeyPair::from_der` be renamed to `KeyPair::guess_from_der`? The `KeyPair::guess_from_der` fn is `pub(crate)` so the extra information being conveyed to the caller through the updated name is only helpful to other `rcgen` developers, not external users right?

thomaseizinger:
> > I went with guess because I deemed it important to convey to the caller that this function isn't necessarily failing because the bytes are malformed but simply because we might not support a particular combination. Additionally, if you pass malformed bytes, the error is not going to tell you what is wrong because we are hiding this by simply trying the next one. To me, that is very different from what I'd expect from a typical try_from function, hence the special name.
> 
> By this logic shouldn't `KeyPair::from_der` be renamed to `KeyPair::guess_from_der`? The `KeyPair::guess_from_der` fn is `pub(crate)` so the extra information being conveyed to the caller through the updated name is only helpful to other `rcgen` developers, not external users right?

Yeah good thinking! Happy to change that if there is consensus on it :)

cpu:
> Yeah good thinking! Happy to change that if there is consensus on it :)

It seems fine to me. Does anyone else have an opinion? I think it would be a good idea to take an action on this PR one way or the other.

thomaseizinger:
> the updated name is only helpful to other `rcgen` developers, not external users right?

Not sure about others but I very often have a quick glance at the "first" layer of abstraction that I am calling from my code so in addition to there being value for contributors, I think there is also value for users. Arguably, not as much as the name of a public function of course.

> > Yeah good thinking! Happy to change that if there is consensus on it :)
> 
> It seems fine to me. Does anyone else have an opinion? I think it would be a good idea to take an action on this PR one way or the other.

I am open to either! There is also an open concern from @djc about whether the early-return style and separate functions are desirable. Should I revert those changes so we are just removing the `TryFrom` impls?

src/key_pair.rs
```diff
@@ -106,65 +105,110 @@ impl KeyPair {
...
-			panic!("Unknown SignatureAlgorithm specified!");
-		};
+				der,
+			)?),
+			alg: &PKCS_ECDSA_P384_SHA384,
+			serialized_der: der.to_vec(),
+		})
+	}
 
+	fn pkcs_rsa_sha256(der: &[u8]) -> Result<KeyPair, Error> {
```
est31, line 183: Could you remove these functions in favour of the old state which assigned `kind`? This saved us a lot of repetitive code.

est31:
I'm fine with the changes to the removal of `TryFrom` and the rename of `from_raw` to `guess_kind_from_der`, but I don't like the addition of these constructor functions and the changes of the implementations of `guess_kind_from_der` and `from_der_and_sign_algo`. Could we have former without the latter?

est31:
Also probably needs a rebase once #183 is merged.

thomaseizinger:
Closing as stale. Unfortunately, I don't have the resources to continue on this, sorry!

cpu:
Sorry that this one ended up sitting around for so long. Thanks for the contribution.

Issue #89, opened 2022-09-02
API is not very clear about how to sign one `Certificate` with another one.


vi:
The only mention of signing new certificates I see is `serialize_*_with_signer` functions.

I am expected to serialize, then immediately deserialize certificate even though there is no intention to save it to file or transmit.

Documentation may be more clear about why signing and serialization are one step. I expected there be something like `rcgen::Certificate::from_params_with_signer(params: CertificateParams, ca: &Certificate)`, so I can sign certificate, then serialize it if needed (or use somehow directly).

est31:
PRs are welcome, and also open to discuss solutions.

szguoxz:
This is interesting, are you saying we can't do it yet? I have tried the whole day without success!

est31:
You can sign certificates with another one. The `_from_signer` functions are relevant here. You can grep the tests for uses of that function.
E.g.: https://docs.rs/rcgen/0.10.0/rcgen/struct.Certificate.html#method.serialize_der_with_signer

cpu:
I think the new example from https://github.com/rustls/rcgen/pull/174 and the upcoming CLI tool are probably sufficient to close this. We can revisit if there are other suggestions to make the API clearer.

PR #164, opened 2023-09-29
wip: refactoring extension handling

cpu:
## Description

Work in progress refactor of extension handling (for certificates, CSRs and CRLs).

Goals:
* Pull extension specific code out of `lib.rs`
* Centralize logic for whether or not extensions need to be omitted
* Prevent emitting duplicate extension OIDs (resolves https://github.com/rustls/rcgen/issues/155)
* Ensure all extensions from `CertificateParams` are written to both certificates and CSRs (resolves https://github.com/rustls/rcgen/issues/122)
* Support reading custom extensions from CSRs (resolves https://github.com/rustls/rcgen/issues/150)

Todo:
* [ ] Support for reading more supported CSR exts into params
  * [ ] basic constraints
  * [ ] extended key usages
  * [ ] name constraints
  * [ ] ...
* [ ] optional fn to allow filtering exts emitted from params into CSR
* [ ] optional fn to allow filtering exts emitted into params from CSR
* [ ] Test coverage for duplicate extension rejection
* [ ] Test coverage for specifying SKI
* [ ] Changelog updates for breaking changes

PR #174, opened 2023-10-09
Add example of generating a cert chain

tbro:
Generate two certificates and sign the second with the first. Addresses #79 and #89.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,10 @@ required-features = ["pem"]
 name = "rsa-irc-openssl"
 required-features = ["pem"]
 
+[[example]]
+name = "sign-leaf-with-ca"
+required-features = ["pem", "x509-parser"]
+
 [dependencies]
 yasna = { version = "0.5.2", features = ["time", "std"] }
 ring = "0.17"
--- a/examples/sign-leaf-with-ca.rs
+++ b/examples/sign-leaf-with-ca.rs
@@ -1,0 +1,58 @@
+use rcgen::{
+	BasicConstraints, Certificate, CertificateParams, DnType, DnValue::PrintableString,
+	ExtendedKeyUsagePurpose, IsCa, KeyUsagePurpose,
+};
+use time::{Duration, OffsetDateTime};
+
+/// Example demonstrating signing end-endity certificate with ca
+fn main() {
+	let ca = new_ca();
+	let end_entity = new_end_entity();
+
+	let end_entity_pem = end_entity.serialize_pem_with_signer(&ca).unwrap();
+	println!("directly signed end-entity certificate: {end_entity_pem}");
+
+	let ca_cert_pem = ca.serialize_pem().unwrap();
+	println!("ca certificate: {ca_cert_pem}",);
+}
+
+fn new_ca() -> Certificate {
+	let mut params = CertificateParams::new(Vec::default());
+	let (yesterday, tomorrow) = validity_period();
+	params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+	params
+		.distinguished_name
+		.push(DnType::CountryName, PrintableString("BR".into()));
+	params
+		.distinguished_name
+		.push(DnType::OrganizationName, "Crab widgits SE");
+	params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+	params.key_usages.push(KeyUsagePurpose::KeyCertSign);
+	params.key_usages.push(KeyUsagePurpose::CrlSign);
+
+	params.not_before = yesterday;
+	params.not_after = tomorrow;
+	Certificate::from_params(params).unwrap()
+}
+
+fn new_end_entity() -> Certificate {
+	let name = "entity.other.host";
+	let mut params = CertificateParams::new(vec![name.into()]);
+	let (yesterday, tomorrow) = validity_period();
+	params.distinguished_name.push(DnType::CommonName, name);
+	params.use_authority_key_identifier_extension = true;
+	params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+	params
+		.extended_key_usages
+		.push(ExtendedKeyUsagePurpose::ServerAuth);
+	params.not_before = yesterday;
+	params.not_after = tomorrow;
+	Certificate::from_params(params).unwrap()
+}
+
+fn validity_period() -> (OffsetDateTime, OffsetDateTime) {
+	let day = Duration::new(86400, 0);
+	let yesterday = OffsetDateTime::now_utc().checked_sub(day).unwrap();
+	let tomorrow = OffsetDateTime::now_utc().checked_add(day).unwrap();
+	(yesterday, tomorrow)
+}


Cargo.toml
```diff
@@ -26,6 +26,10 @@ required-features = ["pem"]
 name = "rsa-irc-openssl"
 required-features = ["pem"]
 
+[[example]]
+name = "sign-leaf-with-ca"
+required-features = ["x509-parser"]
```
djc, line 31: It looks to me like x509-parser is actually not needed here?

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,33 @@
+use rcgen::{BasicConstraints, Certificate, CertificateParams, IsCa};
+use std::fs;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let mut ca_params: CertificateParams = Default::default();
```
djc, line 5: I think the idiomatic way to write these would be `CertificateParams::default()`.

cpu: I think we should aim to have this example produce an issuer certificate that tries to meet the requirements of RFC 5280 and the CABF baseline requirements. I don't think the Rcgen defaults do a great job of this presently, but we can at least account for this in the example code.

In particular I think we should:

* Include a country name and organization name in the root certificate subject [c.f. baseline reqs "7.1.2.10.2 CA Certificate Naming"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712102-ca-certificate-naming).

* Include a `keyUsage` extension with `digitalSignature`, `keyCertSig`n and `cRLSign` asserted. [c.f. baseline reqs "7.1.2.10.7 CA Certificate Key Usage"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712107-ca-certificate-key-usage)

* Fix the validity period so that the `notBefore` is set to one day in the past (instead of a date in 1975), and fix the `notAfter` so that it is no more than 25 years in the future (instead of a date in 4096) c.f. baseline reqs [7.1.2.1.1 Root CA Validity](https://github.com/cabforum/servercert/blob/main/docs/BR.md#71211-root-ca-validity)

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,33 @@
...
+	let mut leaf_params: CertificateParams = Default::default();
```
cpu: Similar to my feedback about the root certificate parameters, let's update the end entity certificate parameters to generate something more closely matching the CABF subscriber certificate profile.

That means:

* Including the authority key identifier ext (e.g. by setting `use_authority_key_identifier_extension` in the params), c.f. baseline reqs ["7.1.2.11.1 Authority Key Identifier"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712111-authority-key-identifier)
* Adding a SAN extension, c.f. baseline reqs ["7.1.2.7.12 Subscriber Certificate Subject Alternative Name"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712712-subscriber-certificate-subject-alternative-name)
* Including a key usage extension that asserts only `digitalSignature` (since this is an ECC public key), c.f baseline reqs ["7.1.2.7.11 Subscriber Certificate Key Usage"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712711-subscriber-certificate-key-usage)
* Including an EKU extensions that asserts `serverAuth`, c.f. baseline reqs ["7.1.2.7.10 Subscriber Certificate Extended Key Usage"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712710-subscriber-certificate-extended-key-usage)
* Setting the `notBefore` to no more than 48 hours before the date of issuance.
* Setting the `notAfter` to no more than 397 days in the future [c.f. baseline reqs "6.3.2 Certificate operational periods and key pair usage periods"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#632-certificate-operational-periods-and-key-pair-usage-periods)

That would still leave two omissions from the server cert profile:

* The authority information access extension, c.f. baseline reqs ["7.1.2.7.7 Subscriber Certificate Authority Information Access"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#71277-subscriber-certificate-authority-information-access)
* The subscriber certificate policies extension, c.f. baseline reqs ["7.1.2.7.9 Subscriber Certificate Certificate Policies"](https://github.com/cabforum/servercert/blob/main/docs/BR.md#71279-subscriber-certificate-certificate-policies)

We don't have code to emit those two extensions and they're the most specific to the context of a proper root program. I think it's fair to defer those for now :+1:

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,33 @@
...
+
+	// in order to sign leaf certificate we pass the ca cert in when
+	// serializing
+	let leaf_serialized = leaf_cert.serialize_pem_with_signer(&ca_cert)?;
+	let ca_serialized = ca_cert.serialize_pem()?;
+
+	println!("{ca_serialized}");
+	println!("{leaf_serialized}");
+	println!("{}", leaf_cert.serialize_private_key_pem());
+	std::fs::create_dir_all("certs/")?;
```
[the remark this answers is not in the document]
tbro: Agree. I was just mimicking what I saw in the other examples. The latest commit removes all the file-system writes.

cpu (changes requested):
Thanks for the PR! 

I'm interested in moving `rcgen` towards generating more realistic and standards compliant certificates by default (in particular, following the guidance set by the CA/Browser forum since it dictates the direction of the most important TLS ecosystem). 

I haven't started on updating any rcgen defaults to follow this model, but in the meantime I think making the _examples_ do so would be a good first step. As the defaults in the library improve the example code can be trimmed down.

As a result of this larger goal most of my feedback is related to ways to make the example produce more complaint certificates. The base code looks good otherwise :-)

tbro:
@cpu I have added the parameters to comply with standards. @djc I basically copied the  same structure from your other example in order to bring in the CSR logic. Unfortunately, `serialize_request_pem` panics with the extra parameters added in the latest commit. So I have removed serialization of the `indirect` certificate. I can think a bit more about this later, but for now I've pushed the changes that resolve the rest of the comments.

cpu:
Thanks!

> Unfortunately, serialize_request_pem panics with the extra parameters added in the latest commit.

I think that will be resolved w/ https://github.com/rustls/rcgen/pull/164 - right now not every extension that can be created from `CertificateParams` can be written to a CSR. There's code in `main` [rejecting KU, EKU, and AKI](https://github.com/rustls/rcgen/blob/948c3b54eae1db242c7966cfa7338001c3928c2f/src/lib.rs#L868-L877) when serializing a CSR from params.

> So I have removed serialization of the indirect certificate. 

That makes sense to me. We can extend the example when the CSR limitation is addressed by #164 (or another PR).

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+		.unwrap();
+	println!("{direct}");
+
+	println!("ca certificate:");
+	let pem = ca.certificate.serialize_pem().unwrap();
+
+	println!("{}", pem);
+}
+
+struct Ca {
```
cpu, line 27: I think these types aren't adding very much and `Ca::new` and `Entity::new` could be written more clearly as free-standing `new_ca` and `new_end_entity` fn's. WDYT?

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+fn main() {
+	let ca = Ca::new();
+	let entity = Entity::new();
+
+	println!("directly signed end-entity certificate:");
+	let direct = entity
+		.certificate
+		.serialize_pem_with_signer(&ca.certificate)
+		.unwrap();
+	println!("{direct}");
```
cpu, line 19: I think it would be cleaner to have only one `println!` by re-arranging the order of things slightly. I also think `end_entity_pem` as a var name is probably easier to understand than `direct`.

```suggestion
	let end_entity_pem = entity
		.certificate
		.serialize_pem_with_signer(&ca.certificate)
		.unwrap();
	println!("directly signed end-entity certificate: {end_entity_pem}");
```

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+
+	println!("ca certificate:");
+	let pem = ca.certificate.serialize_pem().unwrap();
+
+	println!("{}", pem);
```
cpu, line 24: Same feedback here RE: re-arranging for one `println!`.

If you keep the separate `println!`'s I think this second instance should use an in-line var like `println!("{pem}")` for consistency with L19.

I also think `ca_cert_pem` might be a better name for the var as opposed to `pem`.

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+struct Ca {
+	certificate: Certificate,
+}
+
+impl Ca {
+	fn new() -> Self {
+		let mut params = CertificateParams::new(vec!["ca.some.host".to_owned()]);
```
cpu, line 33: A CA certificate shouldn't have a Subject Alternate Name extension, so I think this should be `CertificateParams::new(Vec::default())`.

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+		let (yesterday, tomorrow) = y_t();
+		params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+		params.distinguished_name.push(DnType::CountryName, "BR");
+		params
+			.distinguished_name
+			.push(DnType::OrganizationName, "Crab widgits SE");
```
cpu, line 39: Both the country name and organization name should be using `DnValue::PrintableString` values based on the ASN.1 module in RFC 5280. I think using this form is resulting in `UTF8String` values.

Try:

```rust
params
			.distinguished_name
			.push(DnType::CountryName, DnValue::PrintableString("BR".into()));
```

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+impl Entity {
+	fn new() -> Self {
+		let name = "entity.other.host";
+		let mut params = CertificateParams::new(vec!["entity.other.host".to_owned()]);
+		let (yesterday, tomorrow) = y_t();
+		params.distinguished_name.push(DnType::CommonName, name);
+		params.use_authority_key_identifier_extension = true;
+		params
+			.subject_alt_names
+			.push(rcgen::SanType::DnsName(name.into()));
```
cpu, line 77: Because you gave the same `nam` to the `CertificateParams` constructor this is actually adding a _second_ instance of the same domain name to the SAN extension. We'll want to avoid that. Probably deleting this explicit `subject_alt_names.push` is the best bet :+1:

examples/sign-leaf-with-ca.rs
```diff
@@ -0,0 +1,99 @@
...
+			certificate: Certificate::from_params(params).unwrap(),
+		}
+	}
+
+	fn create_csr(&self) -> String {
+		self.certificate.serialize_request_pem().unwrap()
+	}
+}
+
+fn y_t() -> (OffsetDateTime, OffsetDateTime) {
```
cpu, line 94: I think it'd be clearer to write this function name out fully, or perhaps just call it `validity_period` ?

cpu (commented):
Thanks for the iterations,

> tbro requested review from cpu and djc 32 minutes ago

Here's some initial feedback, but I think the [build failures](https://github.com/rustls/rcgen/actions/runs/6473914181/job/17577670667?pr=174) in CI are also true positives to work through.

examples/sign-leaf-with-ca.rs
cpu: > Both the country name and organization name should be using DnValue::PrintableString

Ah, my mistake, just `CountryName` needs to be a `PrintableString`.

Org name is a `CHOICE` that allows `utf8String` so can be left as-is, but country name should be changed to a printable string [based on 5280 A.1](https://www.rfc-editor.org/rfc/rfc5280#appendix-A.1):

```
-- Naming attributes of type X520countryName (digraph from IS 3166)

id-at-countryName       AttributeType ::= { id-at 6 }

X520countryName ::=     PrintableString (SIZE (2))
```

examples/sign-leaf-with-ca.rs
tbro: I agree

tbro:
> Here's some initial feedback, but I think the [build failures](https://github.com/rustls/rcgen/actions/runs/6473914181/job/17577670667?pr=174) in CI are also true positives to work through.

I addressed your comments and builds seems to be passing now.

est31:
Sorry for getting so late to the discussion. First a general point: I must say that I disagree with the opinion of @cpu to generally move rcgen to conform by CA/Browser forum rules. Yes, they are made by the professionals in the CA industry, those who issue certificates as a paid service. And it is valuable to have these rules apply to most TLS certificates you use to connect to random web services.

But it feels to me that most users of rcgen are *not* the "professionals", so they don't really *need* to conform by these rules. And in my opinion, the professional's rules are quite harmful for users.

* The created certificates are now way bigger than they used to be, causing bloat. This means you cannot paste them as easily in the text form and are more likely to move them around as files.
* Any additional complexity makes it hard to customize the certificates to the use case needed by the user of rcgen.
* The worst part is probably not allowing certificates which last thousands of years. In many settings you don't really *want* to do key rotation or certificate renewal, and have no plans set up for that. I've heard many people reject all of TLS as a *technology* because of this strong enforcement of certificate renewal. In rcgen's use cases, often you have not really planned key rotation or certificate renewal. You deploy some device in the field, a certificate becomes invalid, and then boom, the device doesn't work any more.

To me, it feels that the CA/Browser forum rules are mostly relevant if you are either a member of that forum, or if you otherwise want to create certificates that are trusted by browsers by default. I don't think most users of rcgen are/want to do either, and I built rcgen to serve their needs (which was also my need).

The big browsers don't enforce most of these rules, at least for locally trusted certificates. For trusted-by-default certificates, it's a different thing, which is again fine and a good idea. But locally trusted certificates don't need this info, you know already that they have local trust.

I am in favour of adding a "CA/Browser Forum compliance" mode which enables the appropriate extensions for various scenarios, it could e.g. be different constructors of `CertificateParams`. But I don't think rcgen should output certificates that expire less than 1 year after issuance by default. That's way more btw than what chrome requires for (newly issued) globally trusted certificates, so it would really just be something that ticks a random checkbox somewhere while disrupting users in the field.

-----------

Now to the specific point. I think it's fine to have one or two examples that create CA/Browser Forum compliant certificates, just to show users who need it (they should be in a minority) how it's done. But I don't think we should use it in all examples, or even a majority.

cpu:
>  First a general point: I must say that I disagree with the opinion of @cpu to generally move rcgen to conform by CA/Browser forum rules.

@est31 Fair, but I don't want to side-track this PR arguing my perspective. It sounds like it would be an uphill battle in either case.

> But it feels to me that most users of rcgen are not the "professionals"

For what its worth I think this is a strong argument for why opinionated defaults are in order.

> Now to the specific point. I think it's fine to have one or two examples that create CA/Browser Forum compliant certificates, just to show users who need it (they should be in a minority) how it's done. But I don't think we should use it in all examples, or even a majority.

It sounds like the changes I suggested are fine for the example at hand? If not I think it would be helpful to the OP if you leave some specific suggestions about what needs to change to land this example.

djc:
I feel like we should offer CA/B forum defaults, but under a specific API (like `CertificateParams::web_server()` or some such). I do think I agree that getting this example to go through the CA/B hoops might not be worth it, especially if we're planning to offer an API in the library that handles these details for you. It would be great if we can facilitate all three use cases:

* Test certificates (simple, long-lived)
* Custom deployment (diverse concerns, but making long-lived certificates easy)
* CA/B-compliant certificates

Which seems feasible enough to me?

est31:
Sorry for the off-topic discussion. I am approving this PR.

est31 (approved):

djc (approved):

djc:
> > getting this example to go through the CA/B hoops might not be worth it/increases scope too much for this PR
> 
> The changes I requested have already been made and don't seem to increase the scope by much to me. It's a handful of small changes.
> 
> I think we should defer discussion of future API changes to a separate issue and focus on providing actionable feedback to the contributor. If there are parts of the work that are out of scope, let's suggest changes to bring it in-scope.

Fair enough, sorry!

cpu (approved):
Thank you! Sorry that some of my feedback ended up being controversial :sweat:

cpu:
> cpu force-pushed the sign-leaf-with-ca-example branch from a75cbc3 to aa2caf9 

Squashed the review commits so the merge queue rebase merge will be tidy.

Issue #175, opened 2023-10-11
Command line tool


tbro:
I've seen some talk about the possibility of bringing a command-line-tool to rcgen. [minica](https://github.com/jsha/minica) is an example of this kind of tool for Golang. I recently wrote [myca](https://github.com/tbro/myca) to address my own needs. It uses **rcgen** library. My hope with this issue is to spark a discussion about bringing a **CLI** to **rcgen**. The first question might be to ask if  **rcgen** wants such a thing? If it does, then we can discuss what it should look like. I think it would be relatively easy to provide a tool to create throw away keys and certificates for use in development and testing. And that is what I have tried to solve for myself. Once that bridge has been crossed, it shouldn't be much more effort to provide a few options and switches for compliance in more restricted contexts. I might be oversimplifying and I'm sure there are other opinions. Discussion is welcome.

djc:
rcgen already includes a CLI; check `src/main.rs`. So the question is, how can we better document that a CLI exists, and how can we evolve the CLI to support your use cases?

tbro:
Oh yes sorry for my poor phrasing. As for evolving the existing CLI, I imagine such a tool could accept some parameters to set common parameters. The first thing that comes to mind are setting the `DnsName` and signing/encryption algorithm. It would probably be useful to allow the user to provide an output directory as well.

cpu:
> The first question might be to ask if rcgen wants such a thing?

I'm interested! :-) Thanks for opening the issue.

I think it would probably make sense to offer in a crate separate from the main library, but in the same project workspace/repo. There might be cause for taking quality of life dependencies for the tool that wouldn't be appropriate for the library (and it would help keep us honest w.r.t using only the public API of the library).

> rcgen already includes a CLI; check src/main.rs. So the question is, how can we better document that a CLI exists, and how can we evolve the CLI to support your use cases?

The existing tool can only spit out a self-signed certificate with a hard-coded set of names. I think it would be better to frame it as an example as opposed to documenting it as a tool for end users. It's scope of usefulness seems fairly narrow in the current state.

>  As for evolving the existing CLI, I imagine such a tool could accept some parameters to set common parameters. The first thing that comes to mind are setting the DnsName and signing/encryption algorithm. It would probably be useful to allow the user to provide an output directory as well.

Both of those ideas sound good to me. Looking at `minica` for inspiration it seems like it would also be helpful to consider IP address subjects as well as domain names.

[...]

Issue #137, opened 2023-08-17
Collaborative maintenance offer.


cpu:
Hi @est31,

This crate is a valuable part of the Rust X.509/TLS ecosystem and I'm very grateful for the work you've put in. I'm writing this issue to float the idea of collaborative maintenance of `rcgen`, potentially under the `rustls` organization banner. Is this something you would be open to?

I think having a community supporting `rcgen` will help realize some untapped potential:

* Pull request review. We can help get PRs missing review (e.g. #99) or requiring a final push (e.g. #125) across the line. It will also help move larger features (e.g. #136, #128) forward faster.
* Issue triage and support. We can help respond to user questions (e.g. #123, #126), and prioritize improvements requested through the issue tracker that may remain.
* Feature gaps. We can help implement making *ring* optional (e.g. #74), add more examples (e.g. #110, #89, #79), fix bugs (e.g. #62, #73), and increase test coverage (e.g. #34).

From the Rustls side we can bring a lot of Rust, X.509, TLS, and PKI expertise. We also have resources that can help fund sustained contributions to projects like this as part of a broader mission to replace memory unsafe code in the TLS ecosystem. I see `rcgen` as an important part of this mission and both @ctz and @djc are supportive of this idea.

Some immediate priorities that come to mind:

* Applying the best practices we use throughout the Rustls ecosystem (applying rustfmt, measuring test coverage, creating smaller modules, documenting a release process, etc).
* Augmenting the parameter struct API model with an easier to understand builder model.
* Implementing a command line tool atop the library ala [`minica`](https://github.com/jsha/minica) to offer a paved road for consumers to make a quick standalone PKI without writing Rust code.
* Helping tune generated certificates to have no findings/warnings with tools like [Zlint](https://github.com/zmap/zlint).

If you'd like to join the [Rustls Discord server](https://discord.gg/CsKrm5PH6c) we can talk through any hesitations you might have.

Thanks for considering this!

est31:
Thanks @cpu for reaching out.

First, I've seen your PR #136 and I really like the idea, I just didn't have the time yet to review it in detail. My plan is to do this in the weekend: My new job gives me less free time than before (I like it a lot though).

This brings us to your request. It is great to see that over the years (is it really already almost 5 years? Where is that time gone :o?), rcgen has evolved into something not just useful, but also used by many people in the ecosystem. 

[...]

PR #176, opened 2023-10-13
Add cli workspace crate

tbro:
Names are provisional, please suggest better ones. `rcgen`'s `main.rs` is being used as a provisional `main.rs` in the new crate, but will be replaced and moved to top level examples later.

  * Include some workspace configuration to confirm expectations
  * move `src/main.rs` to `myca/src/main.rs`
  * addresses #175


djc (commented):
I would prefer to have a workspace-only top-level `Cargo.toml` and have the two crates live in separate directories.

myca feels like a too generic name, maybe rustls-cert-gen?

rustls-cert-gen/src/main.rs
```diff
@@ -36,3 +36,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 	fs::write("certs/key.der", &cert.serialize_private_key_der())?;
 	Ok(())
 }
+
+#[test]
```
djc, line 40: This probably shouldn't get added?

rustls-cert-gen/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "rustls-cert-gen"
+version = "0.1.0"
+license.workspace = true
+edition.workspace = true
+
+[[bin]]
```
djc, line 7: The `[[bin]]` section isn't needed -- Cargo should pick it up and do the right thing implicitly.

djc (commented):
It's unfortunate that `Cargo.toml` -> `rcgen/Cargo.toml` isn't recognized as a rename by Git...

rustls-cert-gen/src/main.rs
tbro: Its only there to demonstrate that tests are being run on both crates. I can remove it now, but that whole file is going to get replaced in the next PR anyway.

est31:
> I would prefer to have a workspace-only top-level Cargo.toml and have the two crates live in separate directories.

Given that we only want to have two crates, I don't think that having the extra subdirectory level here makes sense. In other words, I would prefer having one Cargo.toml shared to both be workspace level and be responsible for rcgen.

tbro:
> > I would prefer to have a workspace-only top-level Cargo.toml and have the two crates live in separate directories.
> 
> Given that we only want to have two crates, I don't think that having the extra subdirectory level here makes sense. In other words, I would prefer having one Cargo.toml shared to both be workspace level and be responsible for rcgen. Edit: the binary can be in a separate directory.

I tend to agree. It certainly makes for a smaller diff. What are your thoughts @djc ?

djc:
I find Cargo manifests that mix a workspace and crate metadata very confusing, but if I'm in the minority here I don't think I get to enforce my preferences.

cpu:
> I find Cargo manifests that mix a workspace and crate metadata very confusing, but if I'm in the minority here I don't think I get to enforce my preferences.

I lightly prefer the workspace approach and don't think an extra subdirectory is a meaningful disadvantage (but maybe that's from bad years dealing with very deeply nested Java source trees :grimacing: )

tbro:
Feels like opinions aren't **too** strong on the workspace structure question, yet a decision needs to be made. So I'm going to return to original structure of this PR (top level Cargo.toml will be responsible for both rcgen crate and workspace config). If I'm making the wrong call, we can always change it back again :).

cpu:
> Feels like opinions aren't too strong on the workspace structure question, yet a decision needs to be made

Do the advantages/disadvantages outlined by @tbro in his edited comment above sway your opinion on the subject @est31 ? If not, do you feel strongly enough to consider it blocking feedback for merging this PR? Thanks!

est31:
I'm fine with having a top-level virtual workspace it if we eliminate the src directory in both of these crates and have the .rs files at the top level, like it used to be done in rustc. It's doable by manually specifying a lib/bin section.

djc:
> I'm fine with having a top-level virtual workspace it if we eliminate the src directory in both of these crates and have the .rs files at the top level, like it used to be done in rustc. It's doable by manually specifying a lib/bin section.

IMO this just deviates from conventions more, making the crate harder to work on/understand for outside contributors. I honestly don't understand why you think the added hierarchy levels are such a big problem.

est31:
Extra subdirectories mean longer paths, and more things to click on in editors like vs code or such: you now have to expand the crate dir and then the src dir inside of it. the benefit is negligible because there aren't that many files in the src directory. Each time I have to expand the src dir in the editor I'm annoyed. The benefit that directories provide, bundling up things, is already there thanks to having per-crate directories.

The convention is great for projects that have a bunch of stuff like README, CI files, or license files at the top level but in multi crate projects you don't have those in each crate subdir so you don't need the `src` directory.

tbro:
> The convention is great for projects that have a bunch of stuff like README, CI files, or license files at the top level but in multi crate projects you don't have those in each crate subdir so you don't need the `src` directory.

I think you don't have *all* those files in each crate, but you do have some. To me it makes sense to share a LICENSE, but not so much a README. It seems more digestible to have  separate out documentation. Also the published package on crates.io would need documentation.

cpu:
> Extra subdirectories mean longer paths, and more things to click on in editors like vs code or such: you now have to expand the crate dir and then the src dir inside of it.

My expectation is that most IDEs would preserve the folder expansion state between opening/closing the project. Certainly CLion does this. With that in mind, isn't it a one-time cost? I'm not sure I see this as a significant frustration point.

Cargo.toml
```diff
@@ -1,59 +1,10 @@
...
-
-[[example]]
-name = "rsa-irc"
-required-features = ["pem"]
-
-[[example]]
-name = "rsa-irc-openssl"
-required-features = ["pem"]
+[workspace]
+members = ["rustls-cert-gen", "rcgen"]
```
cpu, line 2: nit: alphabetize the members.

rcgen/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "rcgen"
+version = "0.11.3"
+description = "Rust X.509 certificate generator"
+repository = "https://github.com/est31/rcgen"
```
cpu, line 5: Can you update this to the current URL while you're here? :-)

rustls-cert-gen/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
cpu, line 1: WDYT about adding a `description`? Did you want to wait for reworking the tool to add some of that type of metadata?

cpu (approved):

rustls-cert-gen/Cargo.toml
tbro: The idea was just to add minimal metadata mainly just to test expectations about workspace functionality. But having a description doesn't hurt anything, so I added one in latest commit.

cpu (approved):
Thanks, LGTM.

I'll hold merge until we can confirm djc and est31 are +1 too.

est31:
> I'll hold merge until we can confirm djc and est31 are +1 too.

This is a larger PR so I definitely want it to only be merged with my approval.

> My expectation is that most IDEs would preserve the folder expansion state between opening/closing the project. 

VS code (which I use) does preserve the open state, but it still takes away vertical space. You also have to type `src/` in other settings, like for example I use the git cli for most of my git based interactions instead of the IDE, and there you'd have to type `rcgen/src/lib.rs` instead of `rcgen/lib.rs` or `src/lib.rs`. Sometimes I also use the github interface to interact with a repo, e.g. to browse the tree at a certain commit without touching my local checkout. There too the additional nesting just makes things worse. So it's not really a one time cost but one you pay over and over again.

That's why I deviate from the convention here, and the benefits of the convention are only minimal. Very often I've visited projects and opened the Cargo.toml just to discover that it is a workspace top level toml and I had to guess where the main crate of the project was. In fact, this is one of the least standardized areas of Rust, some people make a `crates/` directory, others have a `lib` directory, in multi language projects you can find `rust` directories. I don't want others to get a similar experience when dealing with rcgen. For a git repo named rcgen, it's reasonable to assume that the top level src directory is for the rcgen crate.

In other words, I don't think the currently proposed directory layout is good. The two layouts I'd be okay with I've outlined in https://github.com/rustls/rcgen/pull/176#issuecomment-1768625586 . 

Should we ever have 4-5+ crates in rcgen, and a 50+ line workspace top level section, we can rethink that choice, but for now I would really prefer having the `src/` dir for rcgen at the top level.

Also, looking at this PR, Cargo.lock should definitely not exist in the `rcgen` subdirectory, nor should codecov.yml. For README, I don't think we need one specific to rcgen either (we could have one specific to the cli tool tho).

cpu:
> This is a larger PR so I definitely want it to only be merged with my approval.

SGTM.

> There too the additional nesting just makes things worse. So it's not really a one time cost but one you pay over and over again.

I think there are strategies to overcome all of these issues (tab completion, using fuzzy file finders, various shortcuts, etc). It still feels to me like a minor issue and not strong justification for why we should do something unique compared to what I see other multi-crate workspaces doing.

> In other words, I don't think the currently proposed directory layout is good. The two layouts I'd be okay with I've outlined in https://github.com/rustls/rcgen/pull/176#issuecomment-1768625586 .

This feels like a place where there's consensus among @tbro, @djc and myself that the current layout implement in-branch is OK. It gives me pause for concern that you want to override that consensus, especially for something relatively small in nature. If in the future there are larger points of contention (particularly in the domain of Rust code as opposed to the project layout) would consensus be overruled again? I can accept one of your proposed alternatives if you feel very strongly, but (respectfully) I do worry that this indicates you want to continue operating as if this is your project and not a community effort.

est31:
I agree that it is a minor issue and if you folks really insist, we can have it the "conventional" way, not because you managed to convince me that it is a better solution technically but because of social reasons (I value you a lot as co-maintainers). The Cargo.lock, codecov yaml and README should definitely be deleted from the rcgen dir though.

Note that I have never really believed in following conventions unless they deliver explicit value. This is probably why I created rcgen in the first place: the convention would have been to use openssl instead of reading the spec and implementing it myself in a library. You know the "don't run your own crypto" saying...

I think since rcgen's creation, the gathering momentum behind rust has attracted attention from increasingly serious people. I suppose these are the folks you @cpu see as the future, currently underserved, potential rcgen users, that you want to attract. My personal use cases were quite non-serious (I just wanted something that outputs something I can feed to quic libraries that forces usage of TLS). Now in 2023 it's way less of a weird choice to pick Rust for some project than it was in 2019.

But while the community grew, I haven't changed. To me, following conventions is just as weird as it was in 2019, although I think I do have gained a larger appreciation for social cohesion in that time period. This non-following of conventions is partially driven by my desire to use my projects as a way to move conventions forward and improve them, but also because I want to do the technically best thing.

I'm typing this precisely because of the concerns you expressed above about the potential disagreements we could have about future Rust changes. I think some of them can also be seen in the "following conventions" vs "trying to set trends"/"doing something tailored for a small target group" angle.

I definitely want to avoid some forking event i.e. what happened to webpki, and therefore am fine with even more collaboration than what I originally thought about #137. Doesn't mean that I will stop saying my opinion though :).

cpu:
Thanks for writing that out @est31. I really appreciate you sharing your opinions and don't want that to stop :+1: In my mind having these discussions seems like a healthy part of a navigating the evolution of an established project.

I also think it's a good practice to question why we're doing something and consider alternatives. I'm definitely not arguing for blindly taking all convention - I'm always open for discussion.

I think when you're working on a project in isolation it's easier to discard convention without increasing barriers to contribution, but when there's more interest in getting help from a broader audience the calculus can change. In my experience the conventions make it easier to move between projects without as much friction. That sentiment is maybe hard to quantify but does feel like explicit value to me, especially as someone trying to contribute across many repositories in a given week. I think the other aspect is that conventions are well known, but your preferences aren't. I think we'll get a better sense of your preferences over time but right now we sort of bump into them on a case-by-case basis more unexpectedly.

You're right that I'm interested in a more serious direction (though maybe we have different ideas of what serious means :-)). I think you're also right that historically I've been more interested in serving the majority use-case vs making sure other use-cases for smaller groups of enthusiasts are served. With that said like you I'm also interested in making this a collaboration that works for everyone. I think having your perspective as a counter balance to mine is very helpful. The world is better for everyone if we can find solutions that help with both sets of users and I'm not unwilling to meet somewhere in the middle.

Back to the PR at hand :sweat_smile: ...

> I agree that it is a minor issue and if you folks really insist, we can have it the "conventional" way

For a small issue, my gut says we should go with the majority if you're willing to bend here. I appreciate that you're making a compromise and I hope we can make the need for such a rare event. I also see it as a great reminder for why I should compromise on some things I might have taken a harder stance on in the future. It should be a two way street - so please call me out if I don't remember that.

>  The Cargo.lock, codecov yaml and README should definitely be deleted from the rcgen dir though.

That sounds uncontroversial to me if it doesn't break something :+1:

tbro:
> > The Cargo.lock, codecov yaml and README should definitely be deleted from the rcgen dir though.
> 
> That sounds uncontroversial to me if it doesn't break something 👍

Those have been moved as requested. Are there any other pending issues with this PR?

cpu (approved):
LGTM!

rcgen/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+version = "0.11.3"
+description = "Rust X.509 certificate generator"
+repository = "https://github.com/rustls/rcgen"
+documentation = "https://docs.rs/rcgen"
+readme.workspace = true
+license.workspace = true
+edition.workspace = true
+keywords = ["mkcert", "ca", "certificate"]
+
+[lib]
```
est31, line 12: this section isn't needed, cargo picks it up automatically.

rcgen/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "rcgen"
+version = "0.11.3"
+description = "Rust X.509 certificate generator"
+repository = "https://github.com/rustls/rcgen"
```
est31, line 5: can this be a workspace-level thingy as well?

est31 (changes requested):
LICENSE is still missing in the new tool's directory.

Also wondering about the CI. Have you looked over the file and verified that all the commands work as they did before?

rcgen/Cargo.toml
tbro: `name` cannot. I moved the rest in latest commit. I also moved `keywords` up a level. reference for keys allowed in `workspace.package`: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

tbro:
> LICENSE is still missing in the new tool's directory.

I copied it to `rustls-cert-gen` crate in latest commit. I think the workspace inheritance was doing this for us anyway, but I guess its best to cover all the bases. Question, since we are including `LICENSE` in the repo, should we be using `license_file` key instead of `license`? https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields.

> Also wondering about the CI. Have you looked over the file and verified that all the commands work as they did before?

I have. Basically, cargo commands will iterate all the crates of the workspace, so CI commands continue to work as before. There are some details in this comment: https://github.com/rustls/rcgen/pull/176#issuecomment-1766853616

est31 (approved):

est31:
> since we are including LICENSE in the repo, should we be using license_file key instead of license

this breaks many tools that do license based validation. having a field that is easily parseable is better for them. I asked for the file to be copied so that it ends up in the package.

cpu:
@tbro Thanks for sticking with this. I appreciate it!

PR #177, opened 2023-10-18
Release 0.11.4

heiher:
Make a new release 0.11.4 to resolve compile issues for new targets that added for `ring`. (LoongArch/RISC-V/S390)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 
 # Changes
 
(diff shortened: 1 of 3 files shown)


djc:
IIRC the ring upgrade from #166 is a semver-incompatible change because we expose the *ring* dependency via the `Error` type, so we probably shouldn't release this as 0.11.x.

PR #179, opened 2023-10-27
Fix error compatibility alias

g2p:
In #167, the error type was renamed from RcgenError to rcgen::Error.

Make sure the compatibility alias uses the correct name.


cpu (approved):
Good catch, thank you!

cpu:
This feels uncontroversial to merge on one review.

djc (approved):

Issue #181, opened 2023-10-27
Tighten up string type representations to prevent illegal values


cpu:
In several places Rcgen defines representations of ASN.1 string types, but in a way that doesn't enforce the values meet the restrictions imposed by the type in use. 

For example, consider the `DnValue` enum with `TeletexString`, `PrintableString`, `UniversalString`, `Utf8String` and `BmpString` variants. Those variants express their values as either `String` or `Vec<u8>`, but in many cases the ASN.1 definitions for these string types introduce further restrictions. As one example, a `PrintableString` can only contain `A-Z, a-z, 0-9, '()+,-./:=?` and `<SPACE>` where as a Rust `String` can contain any valid UTF-8.

Because of this mismatch invalid values can be expressed in `CertificateParams`, and won't be caught by any validation. In the cases where we use a yasna writer and `write_bytes` generically, we'll emit an invalid encoding for the type in use. In the cases where we use a more specific yasna helper like `write_ia5_string` or `write_printable_string`, invalid values will cause a panic.

I think we should look at creating a different representation that can impose more validation in order to reject invalid values at construction time as opposed to emitting invalid encodings or panicing when serializing.

Tudyx:
Does using newtype would make the job?

Something in the idea of the [nutype crate](https://github.com/greyblake/nutype) but without using derive macro

djc:
Yeah, we could probably do newtypes for each of the inner variants, and have `TryFrom` impls for them?

cpu:
That sounds like the right approach to me :+1:

Tudyx:
The created types will look like a lot to the one from the [der::asn1 module](https://docs.rs/der/latest/der/asn1/index.html) of the `der` crate. See for instance the [`PrintableString struct`](https://docs.rs/der/latest/der/asn1/struct.PrintableString.html).
Should we consider adding the `der` crate as a dependency or it's not worth it?

est31:
I don't think we should depend on the der crate for that alone. It has a huge list of dependencies of its own.

Tudyx:
Actually, it has no dependency for our use case. All the dependencies listed in [their ` Cargo.toml`](https://github.com/RustCrypto/formats/blob/master/der/Cargo.toml) are optional, and we only need the `alloc` feature that doesn't activate any. I double-check with `cargo tree` to be sure.

PR #182, opened 2023-10-27
basic Ia5String support for DistinguishedName values

### LINKED ISSUE #180: Ia5String support
Add support for Ia5String in Distinguished Name. Thank you.

cpu:
This branch adds basic support emitting and parsing distinguished name values that are Ia5Strings. For example, email address attributes in a certificate subject distinguished name.

Note that because of #181 this code will panic when emitting invalid Ia5String values. This problem is general to rcgen's handling of ASN.1 string types and so isn't addressed with additional care in this branch. A broader rework is required.

Along the way I also fixed a warning from https://github.com/rustls/rcgen/pull/176 related to where we were defining the custom `profile.dev.package.num-bigint-dig` profile metadata.

Resolves #180

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,9 @@ version = "0.11.3"
 description = "Rust X.509 certificate generator"
 repository = "https://github.com/rustls/rcgen"
 keywords = ["mkcert", "ca", "certificate"]
+
+# This greatly speeds up rsa key generation times
+# (only applies to the dev-dependency of rcgen because cargo
+# ignores profile overrides for non leaf packages)
+[profile.dev.package.num-bigint-dig]
+opt-level = 3
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -363,6 +363,8 @@ impl DnType {
 pub enum DnValue {
 	/// A string encoded using UCS-2
 	BmpString(Vec<u8>),
+	/// An ASCII string.
+	Ia5String(String),
 	/// An ASCII string containing only A-Z, a-z, 0-9, '()+,-./:=? and `<SPACE>`
 	PrintableString(String),
 	/// A string of characters from the T.61 character set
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -13,3 +13,9 @@ version = "0.11.3"
 description = "Rust X.509 certificate generator"
 repository = "https://github.com/rustls/rcgen"
 keywords = ["mkcert", "ca", "certificate"]
+
+# This greatly speeds up rsa key generation times
+# (only applies to the dev-dependency because cargo
```
est31, line 18:
```suggestion
# (only applies to the dev-dependency of rcgen because cargo
```

rcgen/src/lib.rs
```diff
@@ -361,16 +361,18 @@ impl DnType {
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 #[non_exhaustive]
 pub enum DnValue {
-	/// A string of characters from the T.61 character set
```
est31, line 364: the prior order was inspired by the ordering of the `DirectoryString` ASN.1 type in RFC 5280.

Which doesn't list `Ia5String`, but there is a paragraph in the RFC about implementations using using `Ia5String` for emails due to it being able to represent the `@`.

In general, I'm not a fan of reorders like these, as they make review harder and we don't have tools to enforce that the alphabetic order is kept. I'm fine with it this time though.

rcgen/src/lib.rs
```diff
@@ -471,20 +473,15 @@ impl DistinguishedName {
 				.ok_or(Error::CouldNotParseCertificate)?;
 			let dn_type = DnType::from_oid(&attr_type_oid.collect::<Vec<_>>());
 			let data = attr.attr_value().data;
+			let try_str =
```
est31, line 476: very nice

est31 (approved):
see comments.

rcgen/src/lib.rs
cpu: > the prior order was inspired by the ordering of the DirectoryString ASN.1 type in RFC 5280.

Ah I see. In general I would be fine not re-ordering things that are ordered based on an external specification, but in this case since we need a type included that isn't in the list being referenced doing a sort feels like the right call.

> In general, I'm not a fan of reorders like these, as they make review harder

I specifically separated out the reorder into a separate commit to make it easier to review. If you review commit by commit I don't think it's too much of a burden.

> we don't have tools to enforce that the alphabetic order is kept.

True, but we can try to catch it in review. While imperfect, it still feels better than just haphazardly adding to the end each time. 

Since it sounds like you're OK with this instance I'll resolve the feedback and we can revisit case-by-case.

cpu:
@djc Did you want to give this branch a pass before merge?

djc (approved):

PR #184, opened 2023-10-30
Don't use workspace-wide crate versions

djc:
Given that rustls-cert-gen already specifies its version number (such that only one crate actually uses the workspace `version`) and that version numbers are generally pretty closely tied to specific crate's API, IMO it's better to avoid having a workspace-level crate version.


cpu (approved):
This makes sense to me.

est31 (approved):
Yes, this makes sense

PR #185, opened 2023-10-31
Rustls-cert-gen support basic parameters

### LINKED ISSUE #175: Command line tool

tbro:
This commit adds basic functionality for rustls-cerg-gen crate. A
small wrapper library has been added in order to organize code into
small modules and to provide a simple API that can easily be updated
as new functionality is added in the future. There are some rough
edges, some missing documentation for example. But I hope to get the
broader structure approved before refining the documentation. Several
parameters have been added with the idea of supporting a broad range
of use-cases and test expectations around design and
maintainability. The easiest way to view currently supported options
is with `cargo run -- --help`.

Closes #175

### DIFF
--- a/rustls-cert-gen/README.md
+++ b/rustls-cert-gen/README.md
@@ -1,0 +1,30 @@
+# rustls-cert-gen
+
+`rustls-cert-gen` is a tool to generate TLS certificates. In its
+current state it will generate a Root CA and an end-entity
+certificate, along with private keys. The end-entity certificate will
+be signed by the Root CA.
+
+## Usage
+Having compiled the binary you can simply pass a path to output
+generated files.
+
+	cargo run -- -o output/dir
+
+In the output directory you will find these files:
+
+  * `cert.pem`  (end-entity's X.509 certificate, signed by `root-ca`'s key)
+  * `cert.key.pem` (end-entity's private key)
+  * `root-ca.pem` (ca's self-signed x.509 certificate)
+
+For a complete list of supported options:
+
+	rustls-cert-gen --help
+
+## FAQ
+
+#### What signature schemes are available?
+
+  * `pkcs_ecdsa_p256_sha256`
+  * `pkcs_ecdsa_p384_sha384`
+  * `pkcs_ed25519`
--- a/rustls-cert-gen/src/args.rs
+++ b/rustls-cert-gen/src/args.rs
@@ -1,0 +1,75 @@
+//! Command Line argument parsing
+#![allow(missing_docs)]
+
+use std::{net::IpAddr, path::PathBuf};
+
+use bpaf::Bpaf;
+use rcgen::SanType;
+
+#[derive(Clone, Debug, Bpaf)]
+#[bpaf(options)]
+/// rustls-cert-gen Tls Certificate Generator
+pub struct Options {
+	/// Output Directory for generated files
+	#[bpaf(short, long, argument("output/path/"))]
+	pub output: PathBuf,
+	/// Signature algorithm
+	#[bpaf(short, long, fallback("ecdsa_p256".into()), display_fallback)]
+	pub sig_algo: String,
+	#[bpaf(external)]
+	/// Extended Key Usage Purpose: ClientAuth
+	#[bpaf(long)]
+	pub client_auth: bool,
+	/// Extended Key Usage Purpose: ServerAuth
+	#[bpaf(long)]
+	pub server_auth: bool,
+	/// Basename for end-entity cert/key
+	#[bpaf(long, fallback("cert".into()), display_fallback)]
+	pub cert_file_name: String,
+	/// Basename for ca cert/key
+	#[bpaf(long, fallback("root-ca".into()), display_fallback)]
+	pub ca_file_name: String,
+	/// Subject Alt Name (apply multiple times for multiple names/Ips)
+	#[bpaf(many, long, argument::<String>("san"), map(parse_sans))]
+	pub san: Vec<SanType>,
+	/// Common Name (Currently only used for end-entity)
+	#[bpaf(long, fallback("Tls End-Entity Certificate".into()), display_fallback)]
+	pub common_name: String,
+	/// Country Name (Currently only used for ca)
+	#[bpaf(long, fallback("BR".into()), display_fallback)]
+	pub country_name: String,
+	/// Organization Name (Currently only used for ca)
+	#[bpaf(long, fallback("Crab widgits SE".into()), display_fallback)]
+	pub organization_name: String,
+}
+
+/// Parse cli input into SanType. Try first `IpAddr`, if that fails
+/// declare it to be a DnsName.
+fn parse_sans(hosts: Vec<String>) -> Vec<SanType> {
+	hosts.into_iter().map(parse_san).collect()
+}
+
+fn parse_san(host: String) -> SanType {
+	if let Ok(ip) = host.parse::<IpAddr>() {
+		SanType::IpAddress(ip)
+	} else {
+		SanType::DnsName(host)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn test_parse_san() {
+		let hosts = vec!["my.host.com", "localhost", "185.199.108.153"];
+		let sans: Vec<SanType> = hosts.into_iter().map(Into::into).map(parse_san).collect();
+		assert_eq!(SanType::DnsName("my.host.com".into()), sans[0]);
+		assert_eq!(SanType::DnsName("localhost".into()), sans[1]);
+		assert_eq!(
+			SanType::IpAddress("185.199.108.153".parse().unwrap()),
+			sans[2]
+		);
+	}
+}
--- a/rustls-cert-gen/src/cert/ca.rs
+++ b/rustls-cert-gen/src/cert/ca.rs
@@ -1,0 +1,75 @@
+use rcgen::{
+	BasicConstraints, Certificate, CertificateParams, DnType, DnValue::PrintableString, IsCa,
+	KeyUsagePurpose,
+};
+
+use super::PemCertifiedKey;
+
+pub struct CaParams {
+	params: CertificateParams,
+}
+
+impl CaParams {
+	pub fn new(mut params: CertificateParams) -> Self {
+		params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		params.key_usages.push(KeyUsagePurpose::KeyCertSign);
+		params.key_usages.push(KeyUsagePurpose::CrlSign);
+		Self { params }
+	}
+	/// Return `&self.params`.
+	pub fn params(&self) -> &CertificateParams {
+		&self.params
+	}
+	pub fn country_name(mut self, country: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::CountryName, PrintableString(country.into()));
+		self
+	}
+	pub fn organization_name(mut self, name: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::OrganizationName, name);
+		self
+	}
+	pub fn build(self) -> Result<Ca, rcgen::Error> {
+		let cert = Certificate::from_params(self.params)?;
+		let cert = Ca { cert };
+		Ok(cert)
+	}
+}
+
+pub struct Ca {
+	cert: Certificate,
+}
+
+impl Ca {
+	/// Self-sign and serialize
+	pub fn serialize_pem(&self) -> Result<PemCertifiedKey, rcgen::Error> {
+		let cert_pem = self.cert.serialize_pem()?;
+		let key_pem = self.cert.serialize_private_key_pem();
+		Ok(PemCertifiedKey {
+			cert_pem,
+			private_key_pem: key_pem,
+		})
+	}
+	pub fn cert(&self) -> &Certificate {
+		&self.cert
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn init_ca() {
+		let params = CertificateParams::default();
+		let cert = CaParams::new(params);
+		assert_eq!(
+			cert.params().is_ca,
+			IsCa::Ca(BasicConstraints::Unconstrained)
+		)
+	}
+}
--- a/rustls-cert-gen/src/cert/entity.rs
+++ b/rustls-cert-gen/src/cert/entity.rs
@@ -1,0 +1,140 @@
+use super::PemCertifiedKey;
+use rcgen::{
+	Certificate, CertificateParams, DnType, ExtendedKeyUsagePurpose, IsCa, KeyUsagePurpose, SanType,
+};
+
+/// End-entity [Certificate]
+pub struct EndEntity {
+	cert: Certificate,
+}
+
+impl EndEntity {
+	/// Sign with `self.signer` and serialize.
+	pub fn serialize_pem(&self, signer: &Certificate) -> Result<PemCertifiedKey, rcgen::Error> {
+		let cert_pem = self.cert.serialize_pem_with_signer(signer)?;
+		let key_pem = self.cert.serialize_private_key_pem();
+		Ok(PemCertifiedKey {
+			cert_pem,
+			private_key_pem: key_pem,
+		})
+	}
+	pub fn cert(&self) -> &Certificate {
+		&self.cert
+	}
+}
+
+/// [CertificateParams] from which an [EndEntity] [Certificate] can be built
+pub struct EndEntityParams {
+	params: CertificateParams,
+}
+
+impl EndEntityParams {
+	/// Initialize `EndEntityParams`
+	pub fn new(mut params: CertificateParams) -> Self {
+		params.is_ca = IsCa::NoCa;
+		params.use_authority_key_identifier_extension = true;
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		Self { params }
+	}
+	/// Return `&self.params`.
+	pub fn params(&self) -> &CertificateParams {
+		&self.params
+	}
+	pub fn common_name(mut self, name: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::CommonName, name);
+		self
+	}
+	/// `SanTypes` that will be recorded as `subject_alt_names`
+	pub fn subject_alternative_names(mut self, sans: Vec<SanType>) -> Self {
+		for san in sans {
+			self.params.subject_alt_names.push(san);
+		}
+		self
+	}
+	/// Add ClientAuth to `extended_key_usages`.
+	pub fn client_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ClientAuth;
+		self.params.extended_key_usages.push(usage);
+		self
+	}
+	/// Add ServerAuth to `extended_key_usages`.
+	pub fn server_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ServerAuth;
+		self.params.extended_key_usages.push(usage);
+		self
+	}
+	pub fn build(self) -> Result<EndEntity, rcgen::Error> {
+		let cert = Certificate::from_params(self.params)?;
+		let cert = EndEntity { cert };
+		Ok(cert)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::CertificateBuilder;
+	use rcgen::{ExtendedKeyUsagePurpose, IsCa};
+	#[test]
+	fn init_end_endity() {
+		let params = CertificateParams::default();
+		let cert = EndEntityParams::new(params);
+		assert_eq!(cert.params().is_ca, IsCa::NoCa)
+	}
+	#[test]
+	fn client_auth_end_entity() {
+		let _ca = CertificateBuilder::new()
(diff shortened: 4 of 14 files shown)


rustls-cert-gen/README.md
```diff
@@ -0,0 +1,40 @@
+# rustls-cert-gen
+
+`rustls-cert-gen` is a tool to generate TLS certificates.  In its
+current state it will generate a a Root CA and and an end-entity
```
est31, line 4:
```suggestion
current state it will generate a Root CA and an end-entity
```

rustls-cert-gen/README.md
```diff
@@ -0,0 +1,40 @@
...
+certificate, along with private keys. End-entity will be signed by
```
est31, line 5:
```suggestion
certificate, along with private keys. The end-entity certificate will be signed by
```

rustls-cert-gen/src/args.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Clone, Debug, Bpaf)]
+#[bpaf(options)]
+/// rustls-cert-gen Tls Certificate Generator
+pub struct Options {
+	/// Output Directory for generated files
+	#[bpaf(short, long, argument("output/path/"))]
+	pub output: PathBuf,
+	/// Persist ca signing-key to `output` dir
+	#[bpaf(long, switch)]
+	pub insecure: bool,
```
est31, line 31: I'm not sure about calling this mode "insecure" or not doing it by default (IMO it should by default emit all files, including the signing key)

rustls-cert-gen/src/args.rs
```diff
@@ -0,0 +1,95 @@
...
+	#[bpaf(long, fallback("Crab widgits SE".into()), display_fallback)]
+	pub organization_name: String,
+}
+
+/// Parse cli input into SanType. Try first `IpAddr`, if that fails
+/// declare it to be a DnsName.
+fn parse_sans(hosts: Vec<String>) -> Vec<SanType> {
+	hosts
+		.into_iter()
+		.map(|s| s.to_string())
```
est31, line 63: This isn't needed if you make `parse_san` take a `&str` instead of `String`.

est31 (changes requested):
this is a partial review. I'm happy with trying out non-clap options for CLI arg parsing, we can always switch to clap later.

rustls-cert-gen/src/args.rs
tbro: I am happy to change it, but I would like to share my thought process first. I think the signing key is only useful for "advanced" usage. Namely, as input to another tool (or a possible future iteration of this one)  to sign more certificates. Someone looking for a quick way to generate certificates for a TLS connection won't not need it. I think such a person might also be less likely to have a complete grasp of the why and wherefore of all the files. They might also be less informed about secure practices for key management and the implications of leaking the signing key  of their ad-hoc certificate-authority. That's how I came to naming the switch `insecure`. So the idea was to make it easy for those that want something easy and provide a path to more advanced usage for those that know what they are doing.

rustls-cert-gen/src/args.rs
tbro: In that case I think we would need something like `.iter().map(|s|parse_san(s))`. The function call is needed to get the `&str` (I think). But it turns out that `s.to_string()` was unnecessary even without changing the signature. So I just removed that line in the latest commit.

rustls-cert-gen/src/args.rs
djc: I don't like the name here because it doesn't feel like persisting a private key to disk is inherently "insecure". Unless we have functionality for generating new certificates that are signed with an existing on-disk CA private key, I'm not sure it makes sense to persist the CA private key to disk because it'll be relatively hard to use for anything. Adding such functionality (to sign with a persisted CA private key) makes sense to me and probably shouldn't be too hard, but also probably doesn't need to be in this PR.

est31: IMO it makes sense to have the key around even if such a functionality doesn't exist yet, as then the key is around once we add that functionality, instead of being lost.

And yeah, full agree, persisting a private key to disk is not inherently insecure. This is a tool in the end, and yeah if you deploy the CA into a root store then the private key becomes very powerful. But that's up to the users.

tbro: OK `--secure` switch was removed in latest commit and all files will always be persisted.

cpu:
Just leaving a comment to say I haven't forgotten this PR. I'll try and give it a review pass in the next day or two. Thanks!

rustls-cert-gen/src/cert/ca.rs
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+pub struct Ca {
+	cert: Certificate,
+}
+
+impl Cert for Ca {
+	/// Self-sign and serialize
+	fn serialize(&self, _: Option<&Certificate>) -> Result<SerializedEntity, rcgen::Error> {
```
cpu, line 51: I think this name should indicate that it's serializing **PEM**.

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+		let cert_path = dir.join(format!("{name}.pem"));
+		let mut cert_out = File::create(cert_path)?;
+		write!(cert_out, "{}", &self.cert_pem)?;
+
+		Ok(())
+	}
+}
+
+/// Methods common to all Certificate types
+pub trait Cert {
```
cpu, line 39: I'm not sure this trait is worthwhile. For one thing, it creates an awkward situation where `serialize` has to take an `Option` for the signer even though the `Ca` implementation always throws the arg away.

It also doesn't seem like `main.rs` ever handles "Certs" generically, it either has a `Ca` or an `EndEntity` and if both implemented their own `serialize` functions they could be called as required without needing this indirection.

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+mod entity;
+mod signature;
+
+pub trait CertParams {
+	fn params(&self) -> &CertificateParams;
+}
+
+#[derive(Debug, Clone)]
+/// Struct to represent a Serialized Cert/Key pair
+pub struct SerializedEntity {
```
cpu, line 17: I think we could come up with a better name for this. In Rustls-land we call something like this a [CertifiedKey](https://docs.rs/rustls/latest/rustls/sign/struct.CertifiedKey.html).

Maybe `PemCertifiedKey` with rustdoc explaining it as a PEM serialized Certificate and its corresponding PEM serialized private key.

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+	pub cert_pem: String,
+	pub key_pem: String,
```
cpu, line 19: nit: I think we should emphasize this is the private key portion of a key _pair_ to distinguish from a public key.

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl SerializedEntity {
+	pub fn write(&self, dir: &Path, name: &str) -> Result<(), io::Error> {
```
cpu, line 23: I think this will fail if `dir` doesn't exist. Maybe add a `std::fs::create_dir_all(dir)?;`

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+		use std::io::Write;
+
+		let key_path = dir.join(format!("{name}.key.pem"));
+		let mut key_out = File::create(key_path)?;
+		write!(key_out, "{}", &self.cert_pem)?;
```
cpu, line 28: I think this is a bug:
```suggestion
		write!(key_out, "{}", &self.key_pem)?;
```

Maybe there should be a unit test that parses the serialized PEM files and does something with them?

rustls-cert-gen/src/cert/signature.rs
```diff
@@ -0,0 +1,196 @@
+use std::fmt::Display;
+
+#[derive(Debug)]
+pub enum KeyGenerationError {
```
cpu, line 4: I think if possible it would be best to use the `rcgen::Error` type.

rustls-cert-gen/src/cert/params.rs
```diff
@@ -0,0 +1,146 @@
...
+		params.distinguished_name = distinguished_name;
+		Self { params }
+	}
+	/// Set signature algorithm (instead of default). Returns `crate::Result<Self>`.
+	/// # Example
+	/// ```
+	/// # use rustls_cert_gen::BuildParams;
+	/// let cert = BuildParams::new().sig_algo("pkcs_ed25519");
+	/// ```
+	pub fn sig_algo(mut self, alg: &str) -> crate::Result<Self> {
```
cpu, line 39: nit: let's expand this name:
```suggestion
	pub fn signature_algorithm(mut self, alg: &str) -> crate::Result<Self> {
```

rustls-cert-gen/src/cert/params.rs
```diff
@@ -0,0 +1,146 @@
...
+		Ok(self)
+	}
+	/// Set options for generating a CA cert. In this setup
+	/// CA = ROOT (CA certs are self-signed).
+	/// # Example
+	/// ```
+	/// # use rustls_cert_gen::BuildParams;
+	/// let cert = BuildParams::new().ca();
+	/// ```
+	pub fn ca(self) -> CaParams {
```
cpu, line 52: I think this might also be nicer as `certificate_authority` for parity with `end_entity`

rustls-cert-gen/src/cert/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+mod params;
+pub use params::BuildParams;
+
+mod ca;
+mod entity;
+mod signature;
+
+pub trait CertParams {
+	fn params(&self) -> &CertificateParams;
+}
```
cpu, line 13: This feels like another trait that I'm not sure is necessary: it's probably clearer to have `Ca` and `EndEntity` define their own `params()` fn. Like the `Cert` trait there isn't anywhere in the existing `main.rs` where you want to treat both EE and CAs uniformly without knowing which type you have in-hand.

rustls-cert-gen/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+#![warn(missing_docs)]
+//! This library wraps [rcgen] to provide a simple API to generate TLS
+//! certificate-chains. It's primary intent is to ease development of
```
cpu, line 3: typo:
```suggestion
//! certificate-chains. Its primary intent is to ease development of
```

rustls-cert-gen/src/cert/entity.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl EndEntityParams {
+	/// Initialize `EndEntityParams`
+	pub fn new(mut params: CertificateParams) -> Self {
+		params.is_ca = IsCa::NoCa;
+		params.use_authority_key_identifier_extension = true;
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		Self { params }
+	}
+	pub fn cn(mut self, name: &str) -> Self {
```
cpu, line 42: I think we should avoid using acronyms in the API if possible
```suggestion
	pub fn common_name(mut self, name: &str) -> Self {
```

rustls-cert-gen/src/cert/entity.rs
```diff
@@ -0,0 +1,127 @@
...
+		self.params
+			.distinguished_name
+			.push(DnType::CommonName, name);
+		self
+	}
+	/// `SanTypes` that will be recorded as `subject_alt_names`
+	pub fn sans(mut self, sans: Vec<SanType>) -> Self {
```
cpu, line 49: ditto re: acronyms
```suggestion
	pub fn subject_alternative_names(mut self, sans: Vec<SanType>) -> Self {
```

cpu (commented):
Thanks @tbro. I had a fair amount of feedback, but I don't think any of it is particularly complex. LMK what you think.

rustls-cert-gen/README.md
```diff
@@ -0,0 +1,40 @@
...
+	rustls-cert-gen -o output/dir/
+
+In the output directory you will find these files:
+
+  * `cert.pem`  (end-entity's X509 certificate, signed by `root-ca`'s key)
+  * `cert.key.pem` (end-entity's private key)
+  * `root-ca.pem` (ca's self-signed x509 certificate)
+
+If you want to also output root-ca's signing key, you can supply the
+`--insecure` option.
```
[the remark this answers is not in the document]
tbro: correct

rustls-cert-gen/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+With `cargo run`:
+
+	cargo run -- -o output/dir
```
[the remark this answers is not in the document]
tbro: Yes I think the cargo run route is the mostly likely path currently. Possibly worth revisiting when a binary gets published or package is published to crates.io...

rustls-cert-gen/src/cert/params.rs
```diff
@@ -0,0 +1,146 @@
...
+	/// let cert = BuildParams::new().sig_algo("pkcs_ed25519");
+	/// ```
+	pub fn sig_algo(mut self, alg: &str) -> crate::Result<Self> {
+		let sig = Signature::new(alg)?;
+		self.params.alg = sig.key_pair.algorithm();
+		self.params.key_pair = Some(sig.key_pair);
+		Ok(self)
+	}
+	/// Set options for generating a CA cert. In this setup
+	/// CA = ROOT (CA certs are self-signed).
```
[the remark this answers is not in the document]
tbro: I think I was trying to indicate that there is no concept of intermediary certificate. But I don't remember why I wanted to try to say that. I think that that the doc here should just match `end-entity` doc: `Set options for `CA` Certificates.

rustls-cert-gen/src/cert/params.rs
```diff
@@ -0,0 +1,146 @@
+use rcgen::{CertificateParams, DistinguishedName};
+
+use super::ca::CaParams;
+use super::entity::EndEntityParams;
+use super::signature::Signature;
+
+/// Builder to configure TLS [CertificateParams] to be finalized
+/// into either a Ca or an End-Entity.
+pub struct BuildParams {
```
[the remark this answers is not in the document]
tbro: `rcgen::CertficateParams` does not implement `Debug` or `Clone`, so we can't derive them here.

rustls-cert-gen/src/cert/params.rs
```diff
@@ -0,0 +1,146 @@
...
+	params: CertificateParams,
+}
+
+impl Default for BuildParams {
+	fn default() -> Self {
+		Self::new()
+	}
+}
+
+impl BuildParams {
```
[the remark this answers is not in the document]
tbro: I don't like `BuildParams` either, but the underlying type being built is `rcgen::CertficateParams`, so having *Params* in the name seems appropriate to me.

rustls-cert-gen/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Options {
+	/// Output Directory for generated files
+	#[bpaf(short, long, argument("output/path/"))]
+	pub output: PathBuf,
+	/// Signature/Encryption algorithm
+	#[bpaf(short, long, fallback("pkcs_ecdsa_p256_sha256".into()), display_fallback)]
+	pub sig_algo: String,
+	#[bpaf(external)]
+	/// Extended Key Usage Purpose
+	pub purpose: Purpose,
```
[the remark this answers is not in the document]
tbro: You are right. I was confused about that.

rustls-cert-gen/src/cert/params.rs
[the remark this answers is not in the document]
cpu: Oops, you're right. 👍

tbro:
> Thanks @tbro. I had a fair amount of feedback, but I don't think any of it is particularly complex. LMK what you think.

Thanks for being thorough. Unless I missed something, I've taken  all of your suggestions except one. I don't think we can simply derive  `Clone` and `Debug` for `CertficateBuilder` (was `BuildParams`). But if you'd like to pursue another route we can consider more options.

tbro:
> Thanks for being thorough. Unless I missed something, I've taken all of your suggestions except one. I don't think we can simply derive `Clone` and `Debug` for `CertficateBuilder` (was `BuildParams`). But if you'd like to pursue another route we can consider more options.

Double checking and I definitely did miss a few comments, I should be able to address the rest tomorrow.

rustls-cert-gen/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Clone, Debug, Bpaf)]
+#[bpaf(options)]
+/// rustls-cert-gen Tls Certificate Generator
+pub struct Options {
+	/// Output Directory for generated files
+	#[bpaf(short, long, argument("output/path/"))]
+	pub output: PathBuf,
+	/// Signature/Encryption algorithm
+	#[bpaf(short, long, fallback("pkcs_ecdsa_p256_sha256".into()), display_fallback)]
```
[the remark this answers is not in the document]
tbro: Yes I wasn't sure which way to go with the naming so I thought the easiest thing was to expose the underlying definitions to the consumer. I believe it is technically possible to use p256 with a different hash size, but simplifying is probably the better option for a starting point.

rustls-cert-gen/src/cert/mod.rs
tbro: so just `private_key_pem`?

djc:
Thanks for working on this!

IMO it would be good to clean up the commit history a bit (or split into separate PRs) to make this easier to review.

In particular, here are some logical chunks of changes I'm seeing:

* Change original CLI to be an example
* Introduce new rustls-cert-gen crate
* Add README for rustls-cert-gen

I'm surprised `cargo fmt` isn't introducing empty lines between methods in all the new code, I think that's the more idiomatic style.

I also think right now the module sizes in rustls-cert-gen are so small as to make the code harder to follow. In particular, IMO the contents of the `args` module should just be inlined into `main.rs`, same for the contents of `lib.rs`. I'd concatenate all the modules in `cert` into a single file which I guess might be called `cert.rs`.

Personally I'm not fond of `Result` type aliases and I think we should just write it out, which is much more obvious when reviewing. Similarly I've found that type aliases in general are kind of annoying and it might be better to add a newtype wrapper if we want to use `Box<dyn Error>` as the type in the CLI crate (which otherwise certainly seems a decent choice). It might make sense to just use anyhow directly, which I think is geared for this use case directly.

tbro:
> IMO it would be good to clean up the commit history a bit (or split into separate PRs) to make this easier to review.
> 
> In particular, here are some logical chunks of changes I'm seeing:
> 
>     * Change original CLI to be an example
> 
>     * Introduce new rustls-cert-gen crate
> 
>     * Add README for rustls-cert-gen

I can easily squash the history, so I'll start with that. While I generally agree that having small PRs is easier to review, I think this might be an exception to that rule.  Sure, we can have a couple of tiny PRs like those you mention, but there is still going to be a huge one with all the "beef".  Also, I think it is fairly common to initialize new crates with a README, so splitting that off doesn't make much sense to me.  Also moving the former `main.rs` to `examples` is also very much a part of this. I'm happy to do those splits if the team *really* wants it, but I'll wait for more push back (push forward?).

> I also think right now the module sizes in rustls-cert-gen are so small as to make the code harder to follow. In particular, IMO the contents of the `args` module should just be inlined into `main.rs`, same for the contents of `lib.rs`. I'd concatenate all the modules in `cert` into a single file which I guess might be called `cert.rs`.

I disagree that having a few very large files is easier to follow than having smaller dedicated modules. Possibly is related to choice of IDE? I have certainly seen both styles employed by different crates. But I am happy to conform to the preferred style of this team, which I take to be what you have outlined here. 

> Personally I'm not fond of `Result` type aliases and I think we should just write it out, which is much more obvious when reviewing. Similarly I've found that type aliases in general are kind of annoying and it might be better to add a newtype wrapper if we want to use `Box<dyn Error>` as the type in the CLI crate (which otherwise certainly seems a decent choice). It might make sense to just use anyhow directly, which I think is geared for this use case directly.

Given these choices,  I would reach for anyhow. Does anyone disagree with this?

tbro:
> I'm surprised `cargo fmt` isn't introducing empty lines between methods in all the new code, I think that's the more idiomatic style.

There is [blank_lines_lower_bound](https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#blank_lines_lower_bound) but that puts a new line even between every line within a method.

cpu:
@tbro It feels like we should close this and focus on https://github.com/rustls/rcgen/pull/190 - do you agree?

tbro:
> @tbro It feels like we should close this and focus on #190 - do you agree?

agreed :+1:

Issue #186, opened 2023-11-01
codecov bot


est31:
Can we disable the codecov bot? I'm seeing it spamming PRs like #185 with information that is extremely useless. I don't think I've ever found any comment of that bot useful. It's fine if we have a codecov link somewhere with a graph or such, but the github integration is a nuisance.

cpu:
> I'm seeing it spamming PRs like https://github.com/rustls/rcgen/pull/185

I'm not sure I see the spam. There's only one comment from the bot on that PR.

> with information that is extremely useless.

Why do you think the information is useless? I find the summary of coverage changes helpful: it's possible to see all of the changes to coverage across all of the touched files without having to go to a separate service/browser tab.

est31:
I mean the stuff in the diff tab:

![Screenshot_20231101_221228](https://github.com/rustls/rcgen/assets/8872119/9d017f8e-1626-4dbf-bc70-8f0fbaad9732)

> Why do you think the information is useless?

a) I don't really think that chasing some coverage number goal is important. Tests are important, the number isn't. You can make shitty tests that have high coverage numbers, and you can make good tests that increase it by little only. If a PR is adding no tests even though it should add them then you can see that by inspecting the diff, so you see it regardless of whether many lines are added or whether it's just few lines. I always scroll over the codecov bot's comment in the thread, but if it's helpful to you folks we can keep it.

My bigger problem though is with the comments it adds to the diff of the newly added lines (see screenshot): some derive impls not being covered by tests, that's info that I don't care about. A comment on github is an extremely high level of information, and I just want to inspect the diff.

cpu:
> I mean the stuff in the diff tab

Ah ok. I see what you're talking about. Thanks, a screenshot makes this a lot clearer.

> a) I don't really think that chasing some coverage number goal is important

I think you're making a big assumption that this is my motivation. It definitely isn't, which is also why there's no required coverage goal configured.

> Tests are important, the number isn't. 

I agree tests are important. This is a tool for seeing what your tests are and aren't testing. You can try and do that by visual inspection but this is time consuming and error prone. In my experience it's easy to think a test is covering certain branches when it isn't. The comments are meant to be making that explicit and easier to spot in the code review context.

>  You can make shitty tests that have high coverage numbers

Have you seen that happen with PRs being submitted to the repo? That should be review feedback. Let's not add shitty tests :laughing: 

> and you can make good tests that increase it by little only.

[...]

PR #187, opened 2023-11-01
codecov: disable comments

cpu:
Turn off codecov leaving comments on PRs based on the [doc suggestion](https://docs.codecov.com/docs/pull-request-comments#disable-comment).

Resolves https://github.com/rustls/rcgen/issues/186

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,1 @@
-comment:
-  require_changes: true
+comment: false


est31 (approved):
I'm both okay with disabling diff comments as well as disabling them completely, and if latter is needed to get former, then I prefer not having comments at all.

cpu:
> I'm both okay with disabling diff comments as well as disabling them completely

:+1: Let's land this config and see how it goes. If it leaves comments that are still noisy I'll tweak it. We can try more complicated partially enabled configurations down the road.

I don't think we need to wait on djc for this so I'm going to merge.

PR #188, opened 2023-11-13
new `simple` example for rcgen

tbro:
This takes what was formerly `rcgen/src/main.rs` and moves it to the examples folder as `simple.rs`. It was split off from #185

### DIFF
--- a/rcgen/examples/simple.rs
+++ b/rcgen/examples/simple.rs
@@ -1,0 +1,33 @@
+use rcgen::{date_time_ymd, Certificate, CertificateParams, DistinguishedName, DnType, SanType};
+use std::fs;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let mut params: CertificateParams = Default::default();
+	params.not_before = date_time_ymd(1975, 1, 1);
+	params.not_after = date_time_ymd(4096, 1, 1);
+	params.distinguished_name = DistinguishedName::new();
+	params
+		.distinguished_name
+		.push(DnType::OrganizationName, "Crab widgits SE");
+	params
+		.distinguished_name
+		.push(DnType::CommonName, "Master Cert");
+	params.subject_alt_names = vec![
+		SanType::DnsName("crabs.crabs".to_string()),
+		SanType::DnsName("localhost".to_string()),
+	];
+
+	let cert = Certificate::from_params(params)?;
+
+	let pem_serialized = cert.serialize_pem()?;
+	let pem = pem::parse(&pem_serialized)?;
+	let der_serialized = pem.contents();
+	println!("{pem_serialized}");
+	println!("{}", cert.serialize_private_key_pem());
+	fs::create_dir_all("certs/")?;
+	fs::write("certs/cert.pem", pem_serialized.as_bytes())?;
+	fs::write("certs/cert.der", der_serialized)?;
+	fs::write("certs/key.pem", cert.serialize_private_key_pem().as_bytes())?;
+	fs::write("certs/key.der", cert.serialize_private_key_der())?;
+	Ok(())
+}


rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
+#![allow(clippy::complexity, clippy::style, clippy::pedantic)]
```
cpu, line 1: I think this could be removed. I only see a handful of clippy findings in this code and they look easy to fix.

rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use rcgen::{date_time_ymd, Certificate, CertificateParams, DistinguishedName, DnType, SanType};
+use std::fs;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let mut params: CertificateParams = Default::default();
+	params.not_before = date_time_ymd(1975, 01, 01);
+	params.not_after = date_time_ymd(4096, 01, 01);
```
cpu, line 9: Let's avoid octal constants:

```suggestion
	params.not_before = date_time_ymd(1975, 1, 1);
	params.not_after = date_time_ymd(4096, 1, 1);
```

rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
...
+
+	let cert = Certificate::from_params(params)?;
+
+	let pem_serialized = cert.serialize_pem()?;
+	let pem = pem::parse(&pem_serialized)?;
+	let der_serialized = pem.contents();
+	println!("{pem_serialized}");
+	println!("{}", cert.serialize_private_key_pem());
+	std::fs::create_dir_all("certs/")?;
+	fs::write("certs/cert.pem", &pem_serialized.as_bytes())?;
```
cpu, line 30:
```suggestion
	fs::write("certs/cert.pem", pem_serialized.as_bytes())?;
```

rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
...
+	fs::write("certs/cert.der", &der_serialized)?;
```
cpu, line 31:
```suggestion
	fs::write("certs/cert.der", der_serialized)?;
```

rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
...
+	fs::write(
+		"certs/key.pem",
+		&cert.serialize_private_key_pem().as_bytes(),
```
cpu, line 34:
```suggestion
		cert.serialize_private_key_pem().as_bytes(),
```

rcgen/examples/simple.rs
```diff
@@ -0,0 +1,38 @@
...
+	];
+
+	let cert = Certificate::from_params(params)?;
+
+	let pem_serialized = cert.serialize_pem()?;
+	let pem = pem::parse(&pem_serialized)?;
+	let der_serialized = pem.contents();
+	println!("{pem_serialized}");
+	println!("{}", cert.serialize_private_key_pem());
+	std::fs::create_dir_all("certs/")?;
```
cpu, line 29:
```suggestion
	fs::create_dir_all("certs/")?;
```

cpu (commented):
> This takes what was formerly rcgen/src/main.rs and moves it to the examples folder as simple.rs

This code is now duplicated instead of moved right (and I think from a different location than described)? It's in `rustls-cert-gen/src/main.rs` and also `rcgen/examples/simple.rs`.

Would it make more sense for this to be a `git mv rustls-cert-gen/src/main.rs rcgen/examples/simple.rs` and then putting an ~empty `main.rs` in its place?

tbro:
> > This takes what was formerly rcgen/src/main.rs and moves it to the examples folder as simple.rs
> 
> This code is now duplicated instead of moved right (and I think from a different location than described)? It's in `rustls-cert-gen/src/main.rs` and also `rcgen/examples/simple.rs`.

This was the previous `rcgen/src/main.rs`. Then it was moved as a placeholder to `rustls-cert-gen/src/main.rs`.

> Would it make more sense for this to be a `git mv rustls-cert-gen/src/main.rs rcgen/examples/simple.rs` and then putting an ~empty `main.rs` in its place?

I think the `git mv` doesn't make much sense because we also have to edit `rustls-cert-gen/src/main.rs` (add an empty `main()` to prevent breaking `cargo test` and such. so it won't end up being a `rename` in git history anyway.

djc:
> I think the `git mv` doesn't make much sense because we also have to edit `rustls-cert-gen/src/main.rs` (add an empty `main()` to prevent breaking `cargo test` and such. so it won't end up being a `rename` in git history anyway.

I think you can have a rename in the same commit as introducing a new file in the place of the old one? IMO it would be very helpful to make sure this is recorded as a git commit, so that `blame` and tools like that can see through the move.

cpu:
> I think you can have a rename in the same commit as introducing a new file in the place of the old one? 

That's what I had in mind, sorry - I could have been more explicit.

tbro:
> I think you can have a rename in the same commit as introducing a new file in the place of the old one? IMO it would be very helpful to make sure this is recorded as a git commit, so that `blame` and tools like that can see through the move.

I agree with the intent, unfortunately git wants to outsmart us here. I've tried this and double checked and it does not work. We can achieve what you want with two commits, but  I feel there is a good chance those will get squashed/rebased away at some point. If not, just say so and I'll do the two commits.

Here is the best description I could find on git's behavior in this case:
https://stackoverflow.com/a/29822890

cpu:
Ah that's a bummer :-( In that case I think what you have here is fine unless someone else has a stronger preference. Thanks for trying!

cpu (approved):

djc (approved):

est31 (approved):

PR #189, opened 2023-11-13
Add a README.md to rustls-cert-gen

tbro:
Adds a README.md to `rustls-cert-gen`. It was split off from #185.

### DIFF
--- a/rustls-cert-gen/README.md
+++ b/rustls-cert-gen/README.md
@@ -1,0 +1,30 @@
+# rustls-cert-gen
+
+`rustls-cert-gen` is a tool to generate TLS certificates. In its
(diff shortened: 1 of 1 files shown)


cpu (commented):
I don't think splitting this off from the code is a helpful division. The code in `rustls-cert-gen` at the time of this commit won't do the things described in the README, and if further code changes are suggested when reviewing the implementation it will be easy for it to fall out of sync.

cpu:
This content is part of https://github.com/rustls/rcgen/pull/190 and I think it's easier to review it alongside the code. I'm going to close this PR with only the README but if someone disagrees we can reopen.

PR #190, opened 2023-11-13
A functional `rustls-cert-gen` with basic parameters.

### LINKED ISSUE #175: Command line tool

tbro:
This is basically #185 minus #188 and #189. The structure also differs as sub modules have been inlined in `main.rs` and `cert.rs`. `anyhow` has also been added as a dependency to replace the `Result` alias.

Closes #175

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,10 @@ members = ["rcgen", "rustls-cert-gen"]
 resolver = "2"
 
 [workspace.dependencies]
-pem = { version = "3.0.2" }
+pem = "3.0.2"
+rand = "0.8"
+ring = "0.17"
+x509-parser = "0.15.1"
 
 [workspace.package]
 license = "MIT OR Apache-2.0"
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -24,16 +24,17 @@ required-features = ["pem", "x509-parser"]
 [dependencies]
 aws-lc-rs = { version = "1.0.0", optional = true }
 yasna = { version = "0.5.2", features = ["time", "std"] }
-ring = { version = "0.17", optional = true }
+ring = { workspace = true, optional = true }
 pem = { workspace = true, optional = true }
 time = { version = "0.3.6", default-features = false }
-x509-parser = { version = "0.15", features = ["verify"], optional = true }
+x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
 default = ["pem", "ring"]
 aws_lc_rs = ["dep:aws-lc-rs"]
 ring = ["dep:ring"]
+
 
 [package.metadata.docs.rs]
 features = ["x509-parser"]
@@ -46,9 +47,9 @@ allowed_external_types = [
 
 [dev-dependencies]
 openssl = "0.10"
-x509-parser = { version = "0.15", features = ["verify"] }
+x509-parser = { workspace = true, features = ["verify"] }
 rustls-webpki = { version = "0.101.0", features = ["std"] }
 botan = { version = "0.10", features = ["vendored"] }
-rand = "0.8"
+rand = { workspace = true }
 rsa = "0.9"
-ring = "0.17"
+ring = { workspace = true }
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -23,10 +23,10 @@ required-features = ["pem", "x509-parser"]
 
 [dependencies]
 yasna = { version = "0.5.2", features = ["time", "std"] }
-ring = "0.17"
+ring = { workspace = true }
 pem = { workspace = true, optional = true }
 time = { version = "0.3.6", default-features = false }
-x509-parser = { version = "0.15", features = ["verify"], optional = true }
+x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
@@ -43,8 +43,8 @@ allowed_external_types = [
 
 [dev-dependencies]
 openssl = "0.10"
-x509-parser = { version = "0.15", features = ["verify"] }
+x509-parser = { workspace = true, features = ["verify"] }
 rustls-webpki = { version = "0.101.0", features = ["std"] }
 botan = { version = "0.10", features = ["vendored"] }
-rand = "0.8"
-rsa = "0.9"
+rand = { workspace = true }
+rsa = { version = "0.9" }
--- a/rustls-cert-gen/README.md
+++ b/rustls-cert-gen/README.md
@@ -1,0 +1,30 @@
+# rustls-cert-gen
+
+`rustls-cert-gen` is a tool to generate TLS certificates. In its
+current state it will generate a Root CA and an end-entity
+certificate, along with private keys. The end-entity certificate will
+be signed by the Root CA.
+
+## Usage
+Having compiled the binary you can simply pass a path to output
+generated files.
+
+	cargo run -- -o output/dir
+
+In the output directory you will find these files:
+
+  * `cert.pem`  (end-entity's X.509 certificate, signed by `root-ca`'s key)
+  * `cert.key.pem` (end-entity's private key)
+  * `root-ca.pem` (ca's self-signed X.509 certificate)
+
+For a complete list of supported options:
+
+	rustls-cert-gen --help
+
+## FAQ
+
+#### What signature schemes are available?
+
+  * `pkcs_ecdsa_p256_sha256`
+  * `pkcs_ecdsa_p384_sha384`
+  * `pkcs_ed25519`
--- a/rustls-cert-gen/src/cert.rs
+++ b/rustls-cert-gen/src/cert.rs
@@ -222,7 +222,8 @@ impl KeypairAlgorithm {
 
 				let rng = ring::rand::SystemRandom::new();
 				let alg = &rcgen::PKCS_ED25519;
-				let pkcs8_bytes = Ed25519KeyPair::generate_pkcs8(&rng)?;
+				let pkcs8_bytes =
+					Ed25519KeyPair::generate_pkcs8(&rng).or(Err(rcgen::Error::RingUnspecified))?;
 
 				rcgen::KeyPair::from_der_and_sign_algo(pkcs8_bytes.as_ref(), alg)
 			},
@@ -233,7 +234,8 @@ impl KeypairAlgorithm {
 				let rng = ring::rand::SystemRandom::new();
 				let alg = &rcgen::PKCS_ECDSA_P256_SHA256;
 				let pkcs8_bytes =
-					EcdsaKeyPair::generate_pkcs8(&ECDSA_P256_SHA256_ASN1_SIGNING, &rng)?;
+					EcdsaKeyPair::generate_pkcs8(&ECDSA_P256_SHA256_ASN1_SIGNING, &rng)
+						.or(Err(rcgen::Error::RingUnspecified))?;
 				rcgen::KeyPair::from_der_and_sign_algo(pkcs8_bytes.as_ref(), alg)
 			},
 			KeypairAlgorithm::EcdsaP384 => {
@@ -243,7 +245,8 @@ impl KeypairAlgorithm {
 				let rng = ring::rand::SystemRandom::new();
 				let alg = &rcgen::PKCS_ECDSA_P384_SHA384;
 				let pkcs8_bytes =
-					EcdsaKeyPair::generate_pkcs8(&ECDSA_P384_SHA384_ASN1_SIGNING, &rng)?;
+					EcdsaKeyPair::generate_pkcs8(&ECDSA_P384_SHA384_ASN1_SIGNING, &rng)
+						.or(Err(rcgen::Error::RingUnspecified))?;
 
 				rcgen::KeyPair::from_der_and_sign_algo(pkcs8_bytes.as_ref(), alg)
 			},
--- a/rustls-cert-gen/src/cert.rs
+++ b/rustls-cert-gen/src/cert.rs
@@ -1,0 +1,431 @@
+use bpaf::Bpaf;
+use rcgen::{
+	BasicConstraints, Certificate, CertificateParams, DistinguishedName, DnType,
+	DnValue::PrintableString, ExtendedKeyUsagePurpose, IsCa, KeyUsagePurpose, SanType,
+};
+use std::{fmt, fs::File, io, path::Path};
+
+#[derive(Debug, Clone)]
+/// PEM serialized Certificate and PEM serialized corresponding private key
+pub struct PemCertifiedKey {
+	pub cert_pem: String,
+	pub private_key_pem: String,
+}
+
+impl PemCertifiedKey {
+	pub fn write(&self, dir: &Path, name: &str) -> Result<(), io::Error> {
+		use std::io::Write;
+		std::fs::create_dir_all(dir)?;
+
+		let key_path = dir.join(format!("{name}.key.pem"));
+		let mut key_out = File::create(key_path)?;
+		write!(key_out, "{}", &self.private_key_pem)?;
+
+		let cert_path = dir.join(format!("{name}.pem"));
+		let mut cert_out = File::create(cert_path)?;
+		write!(cert_out, "{}", &self.cert_pem)?;
+
+		Ok(())
+	}
+}
+
+/// Builder to configure TLS [CertificateParams] to be finalized
+/// into either a [Ca] or an [EndEntity].
+#[derive(Default)]
+pub struct CertificateBuilder {
+	params: CertificateParams,
+}
+
+impl CertificateBuilder {
+	/// Initialize `CertificateParams` with defaults
+	/// # Example
+	/// ```
+	/// # use rustls_cert_gen::CertificateBuilder;
+	/// let cert = CertificateBuilder::new();
+	/// ```
+	pub fn new() -> Self {
+		let mut params = CertificateParams::default();
+		// override default Common Name
+		params.distinguished_name = DistinguishedName::new();
+		Self { params }
+	}
+	/// Set signature algorithm (instead of default).
+	pub fn signature_algorithm(mut self, alg: &KeypairAlgorithm) -> anyhow::Result<Self> {
+		let keypair = alg.to_keypair()?;
+		self.params.alg = keypair.algorithm();
+		self.params.key_pair = Some(keypair);
+		Ok(self)
+	}
+	/// Set options for Ca Certificates
+	/// # Example
+	/// ```
+	/// # use rustls_cert_gen::CertificateBuilder;
+	/// let cert = CertificateBuilder::new().certificate_authority();
+	/// ```
+	pub fn certificate_authority(self) -> CaBuilder {
+		CaBuilder::new(self.params)
+	}
+	/// Set options for `EndEntity` Certificates
+	pub fn end_entity(self) -> EndEntityBuilder {
+		EndEntityBuilder::new(self.params)
+	}
+}
+
+/// [CertificateParams] from which an [Ca] [Certificate] can be built
+pub struct CaBuilder {
+	params: CertificateParams,
+}
+
+impl CaBuilder {
+	/// Initialize `CaBuilder`
+	pub fn new(mut params: CertificateParams) -> Self {
+		params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		params.key_usages.push(KeyUsagePurpose::KeyCertSign);
+		params.key_usages.push(KeyUsagePurpose::CrlSign);
+		Self { params }
+	}
+	/// Add CountryName to `distinguished_name`. Multiple calls will
+	/// replace previous value.
+	pub fn country_name(mut self, country: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::CountryName, PrintableString(country.into()));
+		self
+	}
+	/// Add OrganizationName to `distinguished_name`. Multiple calls will
+	/// replace previous value.
+	pub fn organization_name(mut self, name: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::OrganizationName, name);
+		self
+	}
+	/// build `Ca` Certificate.
+	pub fn build(self) -> Result<Ca, rcgen::Error> {
+		Ok(Ca {
+			cert: Certificate::from_params(self.params)?,
+		})
+	}
+}
+
+/// End-entity [Certificate]
+pub struct Ca {
+	cert: Certificate,
+}
+
+impl Ca {
+	/// Self-sign and serialize
+	pub fn serialize_pem(&self) -> Result<PemCertifiedKey, rcgen::Error> {
+		Ok(PemCertifiedKey {
+			cert_pem: self.cert.serialize_pem()?,
+			private_key_pem: self.cert.serialize_private_key_pem(),
+		})
+	}
+	/// Return `&Certificate`
+	pub fn cert(&self) -> &Certificate {
+		&self.cert
+	}
+}
+
+/// End-entity [Certificate]
+pub struct EndEntity {
+	cert: Certificate,
+}
+
+impl EndEntity {
+	/// Sign with `signer` and serialize.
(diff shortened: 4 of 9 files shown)


Cargo.toml
```diff
@@ -3,7 +3,11 @@ members = ["rcgen", "rustls-cert-gen"]
 resolver = "2"
 
 [workspace.dependencies]
-pem = { version = "3.0.2" }
+pem = "3.0.2"
+rand = "0.8"
+ring = "0.17"
+rsa = "0.9"
```
cpu, line 9: [My feedback](https://github.com/rustls/rcgen/pull/185/files#r1385579137) on the other PR about the `rsa` dependency was marked as resolved, but the dependency is still present. 

Do you disagree with my suggestion we consider RSA key generation out-of-scope for this tool since rcgen the library can't support it, or did something get lost in the shuffle?

rustls-cert-gen/README.md
```diff
@@ -0,0 +1,30 @@
...
+Having compiled the binary you can simply pass a path to output
+generated files.
+
+	cargo run -- -o output/dir
+
+In the output directory you will find these files:
+
+  * `cert.pem`  (end-entity's X.509 certificate, signed by `root-ca`'s key)
+  * `cert.key.pem` (end-entity's private key)
+  * `root-ca.pem` (ca's self-signed x.509 certificate)
```
cpu, line 18:
```suggestion
  * `root-ca.pem` (ca's self-signed X.509 certificate)
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+		let cert_path = dir.join(format!("{name}.pem"));
+		let mut cert_out = File::create(cert_path)?;
+		write!(cert_out, "{}", &self.cert_pem)?;
+
+		Ok(())
+	}
+}
+
+/// Builder to configure TLS [CertificateParams] to be finalized
+/// into either a Ca or an End-Entity.
```
cpu, line 32:
```suggestion
/// into either a [Ca] or an [EndEntity].
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	/// # use rustls_cert_gen::CertificateBuilder;
+	/// let cert = CertificateBuilder::new();
+	/// ```
+	pub fn new() -> Self {
+		let mut params = CertificateParams::default();
+		// override default Common Name
+		params.distinguished_name = DistinguishedName::new();
+		Self { params }
+	}
+	/// Set signature algorithm (instead of default). Returns `anyhow::Result<Self>`.
```
cpu, line 51: Nit: I think this is duplication with the fn signature.
```suggestion
	/// Set signature algorithm (instead of default).
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	pub fn organization_name(mut self, name: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::OrganizationName, name);
+		self
+	}
+	pub fn build(self) -> Result<Ca, rcgen::Error> {
+		let cert = Certificate::from_params(self.params)?;
+		let cert = Ca { cert };
+		Ok(cert)
```
cpu, line 105:
```suggestion
		Ok(Ca { cert: Certificate::from_params(self.params)? } )
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+
+impl Ca {
+	/// Self-sign and serialize
+	pub fn serialize_pem(&self) -> Result<PemCertifiedKey, rcgen::Error> {
+		let cert_pem = self.cert.serialize_pem()?;
+		let key_pem = self.cert.serialize_private_key_pem();
+		Ok(PemCertifiedKey {
+			cert_pem,
+			private_key_pem: key_pem,
+		})
```
cpu, line 121:
```suggestion
		Ok(PemCertifiedKey {
			cert_pem: self.cert.serialize_pem()?,
			private_key_pem: self.cert.serialize_private_key_pem(),
		})
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+
+impl EndEntity {
+	/// Sign with `self.signer` and serialize.
+	pub fn serialize_pem(&self, signer: &Certificate) -> Result<PemCertifiedKey, rcgen::Error> {
+		let cert_pem = self.cert.serialize_pem_with_signer(signer)?;
+		let key_pem = self.cert.serialize_private_key_pem();
+		Ok(PemCertifiedKey {
+			cert_pem,
+			private_key_pem: key_pem,
+		})
```
cpu, line 141:
```suggestion
		Ok(PemCertifiedKey {
			cert_pem: self.cert.serialize_pem_with_signer(signer)?,
			private_key_pem: self.cert.serialize_private_key_pem(),
		})
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+		params.use_authority_key_identifier_extension = true;
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		Self { params }
+	}
+	pub fn common_name(mut self, name: &str) -> Self {
+		self.params
+			.distinguished_name
+			.push(DnType::CommonName, name);
+		self
+	}
```
cpu, line 163: It looks like calling this multiple times will append multiple `CommonName` DNs. I think probably this should only set the Common Name DN once? Or it should have a docstring describing the behaviour.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	/// `SanTypes` that will be recorded as `subject_alt_names`
+	pub fn subject_alternative_names(mut self, sans: Vec<SanType>) -> Self {
+		for san in sans {
+			self.params.subject_alt_names.push(san);
+		}
+		self
```
cpu, line 169: I think this builder fn has the same property where it's appending for multiple calls. That seems OK for SANs but I think that should be called out in a doc string.

In an ideal world we would error on duplicate SAN values, but that might make more sense in `rcgen`'s handling of the parameters.

```suggestion
		self.params.subject_alt_names.extend(sans);
		self
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	}
+	/// Add ClientAuth to `extended_key_usages`.
+	pub fn client_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ClientAuth;
+		self.params.extended_key_usages.push(usage);
+		self
```
cpu, line 175:
```suggestion
self.params.extended_key_usages.push(ExtendedKeyUsagePurpose::ClientAuth);
		self
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	}
+	/// Add ServerAuth to `extended_key_usages`.
+	pub fn server_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ServerAuth;
+		self.params.extended_key_usages.push(usage);
+		self
```
cpu, line 181:
```suggestion
	self.params.extended_key_usages.push(ExtendedKeyUsagePurpose::ServerAuth);
		self
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	}
+	pub fn build(self) -> Result<EndEntity, rcgen::Error> {
+		let cert = Certificate::from_params(self.params)?;
+		let cert = EndEntity { cert };
+		Ok(cert)
```
cpu, line 186:
```suggestion
		Ok(EndEntity { cert: Certificate::from_params(self.params)? })
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	}
+}
+
+pub struct Signature<'a> {
+	pub alg: &'a rcgen::SignatureAlgorithm,
```
cpu, line 191: Except for tests, the `alg` field is never read. In `CertificateBuilder::signature_algorithm` the `key_pair` field is used to populate the `params.alg`.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+		self
+	}
+	pub fn build(self) -> Result<EndEntity, rcgen::Error> {
+		let cert = Certificate::from_params(self.params)?;
+		let cert = EndEntity { cert };
+		Ok(cert)
+	}
+}
+
+pub struct Signature<'a> {
```
cpu, line 190: I think this type would be better expressed as a an enum named something like `KeypairAlgorithm` with variants `Ed25519`, `EcdsaP256` and `EcdsaP384`.

I think we should avoid "stringly-typed" inputs since they're hard for consumers to use. The `ed25519`, `ecdsa_p256` and `ecdsa_p384` functions on this type could be reconstituted as `KeypairAlgorithm::to_keypair` that converts to `rcgen::KeyPair` as appropriate.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	pub alg: &'a rcgen::SignatureAlgorithm,
+	pub key_pair: rcgen::KeyPair,
+}
+
+impl Signature<'_> {
```
cpu, line 195: Just noting that I'm skipping over reviewing this part based on the feedback above.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+	pub fn subject_alternative_names(mut self, sans: Vec<SanType>) -> Self {
+		for san in sans {
+			self.params.subject_alt_names.push(san);
+		}
+		self
+	}
+	/// Add ClientAuth to `extended_key_usages`.
+	pub fn client_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ClientAuth;
+		self.params.extended_key_usages.push(usage);
```
cpu, line 174: This should probably only `push` if the EKU isn't already present to avoid duplicates.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,433 @@
...
+		self
+	}
+	/// Add ServerAuth to `extended_key_usages`.
+	pub fn server_auth(&mut self) -> &Self {
+		let usage = ExtendedKeyUsagePurpose::ServerAuth;
+		self.params.extended_key_usages.push(usage);
```
cpu, line 180: same feedback here w.r.t duplicates.

rustls-cert-gen/src/main.rs
```diff
@@ -1,38 +1,109 @@
...
+	hosts.into_iter().map(parse_san).collect()
+}
+
+fn parse_san(host: String) -> SanType {
+	if let Ok(ip) = host.parse::<IpAddr>() {
+		SanType::IpAddress(ip)
+	} else {
+		SanType::DnsName(host)
+	}
+}
```
cpu, line 92: WDYT about in-lining this into `parse_sans`? I don't think it has much value on its own since only `parse_sans` uses it (excluding tests which could use `parse_sans` instead).

cpu (commented):
Thanks! Here's another round of feedback

Cargo.toml
tbro: It looks like I removed it from the `rustls-cert-gen` crate but forgot to remove it here.

rustls-cert-gen/src/cert.rs
tbro: I agree. We can have the cli arg parser parse directly to an enum so that `to_keypair` can take a variant. I have implemented this in the latest commit.

rustls-cert-gen/src/cert.rs
tbro: should be addressed now.

rustls-cert-gen/src/cert.rs
tbro: ditto

rustls-cert-gen/src/main.rs
tbro: done

rustls-cert-gen/src/cert.rs
tbro: I think this is a special case. `push` is a bit misleading since `DistinquishedName` is backed by a `hashmap`. Under the hood, `entries.insert()` is called with the `DnType` as the key. So a second push will replace rather than append. So I think the behavior is ok. We can add a docstring like "Set CommonName. Multiple calls will replace previous value".

rustls-cert-gen/src/cert.rs
tbro: This change results in warnings from  `cargo doc`  causing some CI jobs to fail. The warnings are:
```
warning: public documentation for `CertificateBuilder` links to private item `EndEntity`
```
Something must be escaping my attention, but I don't quite understand the warning.

rustls-cert-gen/src/cert.rs
cpu: Ah! Great point. You're right :+1: Sorry about the confusion.

rustls-cert-gen/src/cert.rs
cpu: I think this is related to [feedback I left on the other PR](https://github.com/rustls/rcgen/pull/185#discussion_r1385550049).

If the intent is for this code to be usable as a library in addition to being useful through the CLI then I think in `rustls-cert-gen/src/lib.rs` where you have `pub use cert::CertificateBuilder` you should add the CA and EE types as `pub` exports.

tbro: Oh yes I forgot about that `lib.rs`.

tbro:
> Thanks! Here's another round of feedback

I think I have addressed everything in that round.

cpu:
Thank you! I will do another pass in the next day or two.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,423 @@
...
+	}
+}
+
+/// End-entity [Certificate]
+pub struct EndEntity {
+	cert: Certificate,
+}
+
+impl EndEntity {
+	/// Sign with `self.signer` and serialize.
```
cpu, line 130:
```suggestion
	/// Sign with `signer` and serialize.
```

rustls-cert-gen/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![warn(missing_docs)]
+//! This library wraps [rcgen] to provide a simple API to generate TLS
+//! certificate-chains. Its primary intent is to ease development of
+//! applications that verify chain of trust. It can be used for
```
cpu, line 4:
```suggestion
//! applications that verify chains of trust. It can be used for
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,423 @@
...
+
+impl EndEntityParams {
+	/// Initialize `EndEntityParams`
+	pub fn new(mut params: CertificateParams) -> Self {
+		params.is_ca = IsCa::NoCa;
+		params.use_authority_key_identifier_extension = true;
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		Self { params }
+	}
+	/// Add CommonName to `distinquished_names`. Multiple calls will
```
cpu, line 152:
```suggestion
	/// Add CommonName to `distinguished_name`. Multiple calls will
```

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,423 @@
...
+	pub fn certificate_authority(self) -> CaParams {
+		CaParams::new(self.params)
+	}
+	/// Set options for `EndEntity` Certificates
+	pub fn end_entity(self) -> EndEntityParams {
+		EndEntityParams::new(self.params)
+	}
+}
+
+pub struct CaParams {
```
cpu, line 74: Re-reviewing this, I think `CaBuilder` and `EndEntityBuilder` might be better names since these are returned from the `CertificateBuilder` and continue the building process.

rustls-cert-gen/src/lib.rs
```diff
@@ -0,0 +1,8 @@
...
+//! whatever purpose you may need a TLS certificate-chain.
+
+mod cert;
+pub use cert::{Ca, CertificateBuilder, EndEntity};
```
cpu, line 8: I think we want `EndEntityParams` and `CaParams` here too (or whatever they end up named if you agree with my rename suggestion).

Right now the return types from `CertificateBuilder::end_entity` and `CertificateBuilder::ca` are opaque out of crate.

rustls-cert-gen/src/main.rs
```diff
@@ -1,38 +1,117 @@
...
+
+	ca.serialize_pem()?
+		.write(&opts.output, &opts.ca_file_name)?;
+
 	Ok(())
 }
+
+#[derive(Clone, Debug, Bpaf)]
+#[bpaf(options)]
+/// rustls-cert-gen Tls Certificate Generator
```
cpu, line 45:
```suggestion
/// rustls-cert-gen TLS Certificate Generator
```

rustls-cert-gen/src/main.rs
```diff
@@ -1,38 +1,117 @@
...
+struct Options {
+	/// Output Directory for generated files
```
cpu, line 47:
```suggestion
	/// Output directory for generated files
```

rustls-cert-gen/src/main.rs
```diff
@@ -1,38 +1,117 @@
...
+	fn test_parse_san() {
+		let hosts = vec!["my.host.com", "localhost", "185.199.108.153"]
+			.into_iter()
+			.map(Into::into)
+			.collect();
+		let sans: Vec<SanType> = parse_sans(hosts);
+		assert_eq!(SanType::DnsName("my.host.com".into()), sans[0]);
+		assert_eq!(SanType::DnsName("localhost".into()), sans[1]);
+		assert_eq!(
+			SanType::IpAddress("185.199.108.153".parse().unwrap()),
```
cpu, line 113: I think it would be nice to have an IPv6 address testcase here too.

cpu (commented):
Sorry for the delay getting this another review. I only have small nits to flag. I think there's likely good future work (like supporting a way to override the validity period), but I'm leaning towards trying to land this ASAP and iterating from there.

tbro:
> Sorry for the delay getting this another review. I only have small nits to flag. I think there's likely good future work (like supporting a way to override the validity period), but I'm leaning towards trying to land this ASAP and iterating from there.

Yes I also thought of the validity periods. I'm sure there are a few more features that could be added. But my preference is also to finish this first bit. Then it should be much easier to review smaller future PRs that target specific features/enhancements.

tbro:
I believe latest review comments have been addressed.

est31 (approved):

rcgen/Cargo.toml
```diff
@@ -23,10 +23,10 @@ required-features = ["pem", "x509-parser"]
 
 [dependencies]
 yasna = { version = "0.5.2", features = ["time", "std"] }
-ring = "0.17"
+ring = { workspace = true }
 pem = { workspace = true, optional = true }
 time = { version = "0.3.6", default-features = false }
-x509-parser = { version = "0.15", features = ["verify"], optional = true }
+x509-parser = { workspace = true, features = ["verify"], optional = true }
```
djc, line 29: Personally I would prefer specifying all dependency versions at the workspace level, because otherwise it's not very clear when adding a dependency where to look.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,431 @@
...
+	/// # use rustls_cert_gen::CertificateBuilder;
+	/// let cert = CertificateBuilder::new();
+	/// ```
+	pub fn new() -> Self {
+		let mut params = CertificateParams::default();
+		// override default Common Name
+		params.distinguished_name = DistinguishedName::new();
+		Self { params }
+	}
+	/// Set signature algorithm (instead of default).
```
djc, line 52: Please add empty lines between all the methods.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,431 @@
...
+
+		let cert_path = dir.join(format!("{name}.pem"));
+		let mut cert_out = File::create(cert_path)?;
+		write!(cert_out, "{}", &self.cert_pem)?;
+
+		Ok(())
+	}
+}
+
+/// Builder to configure TLS [CertificateParams] to be finalized
```
djc, line 32: I think usually we link names with both backticks and brackets, like ```[`CertificateParams`]```.

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,431 @@
...
+		Self { params }
+	}
+	/// Set signature algorithm (instead of default).
+	pub fn signature_algorithm(mut self, alg: &KeypairAlgorithm) -> anyhow::Result<Self> {
+		let keypair = alg.to_keypair()?;
+		self.params.alg = keypair.algorithm();
+		self.params.key_pair = Some(keypair);
+		Ok(self)
+	}
+	/// Set options for Ca Certificates
```
djc, line 59: Should probably be ```[`Ca`]```, let's not capitalize "Certificates" here?

rustls-cert-gen/src/cert.rs
```diff
@@ -0,0 +1,431 @@
+use bpaf::Bpaf;
+use rcgen::{
+	BasicConstraints, Certificate, CertificateParams, DistinguishedName, DnType,
+	DnValue::PrintableString, ExtendedKeyUsagePurpose, IsCa, KeyUsagePurpose, SanType,
+};
+use std::{fmt, fs::File, io, path::Path};
+
+#[derive(Debug, Clone)]
+/// PEM serialized Certificate and PEM serialized corresponding private key
```
djc, line 9: In terms of item ordering I think it would make more sense to keep `PemCertifiedKey` closer to `EndEntity` and `Certificate`, also keeping `EndEntityBuilder` and `CaBuilder` together.

djc (commented):
I have a few minor nits (hope these are not contradicting any earlier feedback), this looks good to me!

djc (approved):

cpu:
> exclude rustls-cert-gen from semver check

Why is this necessary? Can you add context motivating the change to the commit message?

tbro:
> > exclude rustls-cert-gen from semver check
> 
> Why is this necessary - because there's no initial published verison? Can you add context motivating the change to the commit message?

Yes I'm working around CI failure due to there no published version of `rustls-cert-gen`. I have updated commit message as requested.

cpu:
Makes sense, thank you!

tbro:
Hello everyone. Since we have 3 approvals can we go ahead and merge this? Pretty much every merge into `main` is causing conflicts requiring manual conflict resolution. In some cases changes to code. I have left these resolutions as separate commits above with clear messages, so they can easily be reviewed.  I am happy to squashed them all together given your go ahead.  I fear if this PR is left sitting around the burden will become too much to keep up with. I understand that there are a few comments remaining above, but they are self-described as "minor nits". I suggest we move those to a separate issue . Feel free to assign it to me.

cpu:
> Hello everyone. Since we have 3 approvals can we go ahead and merge this?

That sounds good to me, I've been waiting for the branch to build green in CI again and it looks like that's been addressed. Thanks!

@djc @est31 Any reservations about merging this?

est31:
I'm fine with merging!

est31:
(and thanks @tbro for your contribution)

cpu:
> I'm fine with merging!

Cool, thanks for chiming in. I'm going to put it in the queue. Talking to Djc OOB I don't think he'll have a chance to look at this. If he does later we can tackle feedback in follow-up PRs.

> (and thanks @tbro for your contribution)

+1!

PR #191, opened 2023-11-14
Ensure CI covers examples and unit tests, fix clippy findings

cpu:
While reviewing https://github.com/rustls/rcgen/pull/188 I wanted to confirm that example code was being built in CI. It turns out that it wasn't. Similarly we haven't been running `clippy` against test code, and so there was a number of findings to address.

This branch updates CI to:

* Remove `--all`.  This is a deprecated alias for `--workspace`, and `--workspace` is the default for a directory containing a workspace so it can be omitted.
* Use `--all-targets` whenever we run `cargo check`, `cargo test` or `cargo clippy`. This ensures coverage for both examples and unit tests.

In order for the `cargo clippy ... --all-targets` to succeed this branch addresses each of the findings that were present. I've done this with a separate commit per class of finding to make it easier to review. In one case (7bfe0ef5d87d48ae039afe332604e8858e6f1ca1) I allowed the finding instead of fixing it since it seemed like the choice of digit groupings was done intentionally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,8 +38,8 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      - run: cargo clippy --all-features
-      - run: cargo clippy --no-default-features
+      - run: cargo clippy --all-features --all-targets
+      - run: cargo clippy --no-default-features --all-targets
 
   rustdoc:
     name: Documentation
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
       - name: cargo doc (all features)
-        run: cargo doc --all --all-features --document-private-items
+        run: cargo doc --all-features --document-private-items
         env:
           RUSTDOCFLAGS: ${{ matrix.rust_channel == 'nightly' && '-Dwarnings --cfg=docsrs' || '-Dwarnings' }}
 
@@ -89,10 +89,10 @@ jobs:
       uses: dtolnay/rust-toolchain@stable
     - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
     - run: vcpkg install openssl:x64-windows-static-md
-    - name: Run cargo check --all
-      run: cargo check --all
+    - name: Run cargo check
+      run: cargo check
     - name: Run the tests
-      run: cargo test --all
+      run: cargo test
     - name: Run the tests with x509-parser enabled
       run: cargo test --verbose --features x509-parser
     - name: Run the tests with no default features enabled
@@ -128,10 +128,10 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.toolchain }}
-    - name: Run cargo check --all
-      run: cargo check --all
+    - name: Run cargo check
+      run: cargo check
     - name: Run the tests
-      run: cargo test --all
+      run: cargo test
     - name: Run the tests with x509-parser enabled
       run: cargo test --verbose --features x509-parser
     - name: Run the tests with no default features enabled
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,13 +90,13 @@ jobs:
     - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
     - run: vcpkg install openssl:x64-windows-static-md
     - name: Run cargo check
-      run: cargo check
+      run: cargo check --all-targets
     - name: Run the tests
-      run: cargo test
+      run: cargo test --all-targets
     - name: Run the tests with x509-parser enabled
-      run: cargo test --verbose --features x509-parser
+      run: cargo test --verbose --features x509-parser --all-targets
     - name: Run the tests with no default features enabled
-      run: cargo test --verbose --no-default-features
+      run: cargo test --verbose --no-default-features --all-targets
 
   build:
     strategy:
@@ -129,13 +129,13 @@ jobs:
       with:
         toolchain: ${{ matrix.toolchain }}
     - name: Run cargo check
-      run: cargo check
+      run: cargo check --all-targets
     - name: Run the tests
-      run: cargo test
+      run: cargo test --all-targets
     - name: Run the tests with x509-parser enabled
-      run: cargo test --verbose --features x509-parser
+      run: cargo test --verbose --features x509-parser --all-targets
     - name: Run the tests with no default features enabled
-      run: cargo test --verbose --no-default-features
+      run: cargo test --verbose --no-default-features --all-targets
 
   coverage:
     name: Measure coverage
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
       - name: cargo doc (all features)
-        run: cargo doc --all --all-features --document-private-items
+        run: cargo doc --workspace --all-features --document-private-items
```
djc, line 60: Seems to me that `--workspace` is also the default in directories that contain a workspace -- so we could just leave it out and clean all these up? That would be my preference.

.github/workflows/ci.yml
```diff
@@ -89,14 +89,14 @@ jobs:
       uses: dtolnay/rust-toolchain@stable
     - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
     - run: vcpkg install openssl:x64-windows-static-md
-    - name: Run cargo check --workspace
-      run: cargo check --workspace
+    - name: Run cargo check --workspace --all-targets
+      run: cargo check --workspace --all-targets
     - name: Run the tests
-      run: cargo test --workspace
+      run: cargo test --workspace --all-targets
```
djc, line 95: `test` defaults to `--all-targets` so I think we should leave it out here and below.

djc (approved):

rcgen/examples/rsa-irc-openssl.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(clippy::complexity, clippy::style, clippy::pedantic)]
```
djc, line 1: Not sure this tracks with your commit message? If new lints appear in any of these categories we're still going to warn about these, which (according to this `#![allow]`) is unwanted?

djc (approved):
Nice!

.github/workflows/ci.yml
cpu: I swore I tested this and didn't see it processing the examples directory but I might have made a mistake. I'll double check and drop it if so

rcgen/examples/rsa-irc-openssl.rs
cpu: I thought being warned for new lints that apply would be the desired behaviour.

.github/workflows/ci.yml
cpu: Yeah, I'm not sure if there's something deeper going on here, but I see a difference when `--all-targets` isn't provided and I run `cargo test` in the root of the repo:

Without:
```
[daniel@blanc:~/Code/Rust/rcgen]$ cargo test 1>/dev/null
    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running unittests src/lib.rs (target/debug/deps/rcgen-f88f7e4d6f3a7bd5)
     Running tests/botan.rs (target/debug/deps/botan-c52964ac1215ee80)
     Running tests/generic.rs (target/debug/deps/generic-19ee0dcd0b8cb3b1)
     Running tests/openssl.rs (target/debug/deps/openssl-0a1d87c1cbad2f2f)
     Running tests/util.rs (target/debug/deps/util-a1c4b7443d444db3)
     Running tests/webpki.rs (target/debug/deps/webpki-94dd7b7973978afc)
     Running unittests src/main.rs (target/debug/deps/rustls_cert_gen-0df5c6c468d810e0)
   Doc-tests rcgen
```

With:
```
[daniel@blanc:~/Code/Rust/rcgen]$ cargo test --all-targets 1>/dev/null
    Finished test [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (target/debug/deps/rcgen-f88f7e4d6f3a7bd5)
     Running tests/botan.rs (target/debug/deps/botan-c52964ac1215ee80)
     Running tests/generic.rs (target/debug/deps/generic-19ee0dcd0b8cb3b1)
     Running tests/openssl.rs (target/debug/deps/openssl-0a1d87c1cbad2f2f)
     Running tests/util.rs (target/debug/deps/util-a1c4b7443d444db3)
     Running tests/webpki.rs (target/debug/deps/webpki-94dd7b7973978afc)
     Running unittests examples/rsa-irc.rs (target/debug/examples/rsa_irc-1f3811a2067a25e1)
     Running unittests examples/rsa-irc-openssl.rs (target/debug/examples/rsa_irc_openssl-33576615a0a33f63)
     Running unittests src/main.rs (target/debug/deps/rustls_cert_gen-0df5c6c468d810e0)
```

Notable the `--all-targets` version is the only one that outputs lines for `examples/...`

.github/workflows/ci.yml
cpu: Fixed to remove `--all` throughout, using the implicit `--workspace` default.

rcgen/examples/rsa-irc-openssl.rs
cpu: I reworked the commit message.

.github/workflows/ci.yml
djc: Well, I don't think we need `cargo test` to cover examples as long as we have `cargo check --all-targets` covering examples?

rcgen/examples/rsa-irc-openssl.rs
djc: I read this `#![allow]` as wanting to not see warnings for new lints from these particular categories.

.github/workflows/ci.yml
cpu: Is there harm in having `cargo test` use `--all-targets`? I admit it seems unlikely there will be unit tests added to an example but it seems better to be in a state where if that does happen we don't need to remember to make any changes to CI.

I don't feel strongly but I also don't see the downside.

djc: Oh, so this is about unit tests in examples... Sure, I guess it's fine.

rcgen/examples/rsa-irc-openssl.rs
cpu: I guess I'm not sure why you wouldn't want to see those warnings. Fixing them doesn't seem like a tremendous burden and will keep the example code more idiomatic.

If it's a point of contention I can drop this change but it seems like an unnecessary concern.

.github/workflows/ci.yml
cpu: Yup: my primary goal is to ensure that all code is built in CI, and all tests are run, no matter where they're defined.

rcgen/examples/rsa-irc-openssl.rs
djc: (I'm not saying having this would be my choice... just guessing at the original intention.)

cpu: Let's wait and see what est31 thinks. I'll restore the `allow`'s if requested, but if that's the direction we go it should be done consistently. `sign-leaf-with-ca.rs` is missing the `allow` and should be fixed.

rcgen/tests/openssl.rs
```diff
@@ -79,7 +79,7 @@ impl Read for PipeEnd {
 	fn read(&mut self, mut buf: &mut [u8]) -> ioResult<usize> {
 		let inner = self.inner.borrow_mut();
 		let r_sl = &inner.0[1 - self.end_idx][self.read_pos..];
-		if r_sl.len() == 0 {
+		if r_sl.is_empty() {
```
est31, line 82: this is one of the clippy lints that I don't like :(.

est31 (approved):

PR #201, opened 2023-12-13
Add support for `aws_lc_rs` as crypto backend

BiagioFesta:
Starting from [0.22.0, `rustls`](https://github.com/rustls/rustls/releases/tag/v%2F0.22.0), ecosystem has started supporting [`aws-lc-rs`](https://github.com/aws/aws-lc-rs) as crypto back-end. This latter replaces crypto functionalities provided by `ring`.

The main reason behind this is to provide [FIPS](https://en.wikipedia.org/wiki/FIPS_140-2) compliance in the future via `aws-lc`:

> Optional support for cryptography from [aws-lc-rs](https://github.com/aws/aws-lc-rs). Once the certification process completes , we will support FIPS mode using aws-lc-rs.

---

This PR is implementing the [same](https://github.com/rustls/rustls/blob/4d1b762b5328a1714862ba73ec72d5522fe0c049/rustls/Cargo.toml#L19) supporting mechanism.

In short:
 * if cfg(feature = "ring") -> ring   (this will be the default)
 * if cfg(feature = "aws-lc-rc") -> aws-lc-rs
 * if both -> ~compile_error!~ EDIT: ring
 * if none -> compile_error!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           components: clippy
       - run: cargo clippy --all-features --all-targets
-      - run: cargo clippy --no-default-features --all-targets
+      - run: cargo clippy --no-default-features --features aws_lc_rs,pem --all-targets
 
   rustdoc:
     name: Documentation
@@ -117,6 +117,8 @@ jobs:
         arch: amd64
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
+    - name: Install NASM for aws-lc-rs on Windows
+      uses: ilammy/setup-nasm@v1
     - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
     - run: vcpkg install openssl:x64-windows-static-md
     - name: Run cargo check
@@ -125,8 +127,8 @@ jobs:
       run: cargo test --all-targets
     - name: Run the tests with x509-parser enabled
       run: cargo test --verbose --features x509-parser --all-targets
-    - name: Run the tests with no default features enabled
-      run: cargo test --verbose --no-default-features --all-targets
+    - name: Run the tests with aws_lc_rs backend enabled
+      run: cargo test --verbose --no-default-features --features aws_lc_rs,pem --all-targets
 
   build:
     strategy:
@@ -164,8 +166,8 @@ jobs:
       run: cargo test --all-targets
     - name: Run the tests with x509-parser enabled
       run: cargo test --verbose --features x509-parser --all-targets
-    - name: Run the tests with no default features enabled
-      run: cargo test --verbose --no-default-features --all-targets
+    - name: Run the tests with aws_lc_rs backend enabled
+      run: cargo test --verbose --no-default-features --features aws_lc_rs,pem --all-targets
 
   coverage:
     name: Measure coverage
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -1,13 +1,16 @@
 #[cfg(feature = "pem")]
 use pem::Pem;
-use ring::rand::SystemRandom;
-use ring::signature::KeyPair as RingKeyPair;
-use ring::signature::{self, EcdsaKeyPair, Ed25519KeyPair, RsaEncoding, RsaKeyPair};
 use std::convert::TryFrom;
 use std::fmt;
 use yasna::DERWriter;
 
 use crate::error::ExternalError;
+use crate::ring_like::error as ring_error;
+use crate::ring_like::rand::SystemRandom;
+use crate::ring_like::signature::{
+	self, EcdsaKeyPair, Ed25519KeyPair, KeyPair as RingKeyPair, RsaEncoding, RsaKeyPair,
+};
+use crate::ring_like::{ecdsa_from_pkcs8, rsa_key_pair_public_modulus_len};
 use crate::sign_algo::algo::*;
 use crate::sign_algo::SignAlgo;
 #[cfg(feature = "pem")]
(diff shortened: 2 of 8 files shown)


djc:
Given that Cargo features are supposed to be additive, I'm not a fan of this:

> * if both -> compile_error!

I would suggest just picking *ring* in that case, or, if you don't like that solution, providing explicit access to a crypto provider via some kind of trait.

cpu:
I was hoping the aws-lc-rs backend would unblock RSA key generation support, but it looks as though there's upstream work required. I've filed a feature req: https://github.com/aws/aws-lc-rs/issues/296

BiagioFesta:
> Given that Cargo features are supposed to be additive, I'm not a fan of this:
> 
> > * if both -> compile_error!
> 
> I would suggest just picking _ring_ in that case, or providing explicit access to a crypto provider via some kind of trait (not sure how invasive that would be to the API).

Fixed, CI still fails also because some `--no-default-features` tests. What about that case?

est31:
compile_error is fine in that case, tests should be changed to either specify `ring` or `aws_lc_rs`. The main important property is that cargo features stay additive.

rcgen/src/lib.rs
```diff
@@ -116,6 +116,8 @@ mod csr;
 mod error;
 mod key_pair;
 mod oid;
+/// re-export of crypto functions (back-ed by `ring` or `aws-lc-rs`)
```
est31, line 119:
```suggestion
/// re-export of crypto functions (backed by `ring` or `aws-lc-rs`)
```

est31 (approved):

est31:
Maybe the name can be bikeshedded but otherwise this is fine to go.

BiagioFesta:
Codecoverage test fails but I think it is not fair :)

cpu:
> Codecoverage test fails but I think it is not fair :)

Yup! The flagged lines are missing coverage on `main`, don't worry about that one :+1:

.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
     - name: Run the tests with x509-parser enabled
       run: cargo test --verbose --features x509-parser --all-targets
     - name: Run the tests with no default features enabled
-      run: cargo test --verbose --no-default-features --all-targets
+      run: cargo test --verbose --all-targets
```
cpu, line 129: The `name` and `run` have fallen out of sync here.

.github/workflows/ci.yml
```diff
@@ -165,7 +165,7 @@ jobs:
...
```
cpu, line 168: Another spot that needs the `name` updated.

rcgen/src/key_pair.rs
```diff
@@ -1,13 +1,15 @@
...
-use ring::signature::{self, EcdsaKeyPair, Ed25519KeyPair, RsaEncoding, RsaKeyPair};
 use std::convert::TryFrom;
 use std::fmt;
 use yasna::DERWriter;
 
 use crate::error::ExternalError;
+use crate::ring_like::error as ring_error;
+use crate::ring_like::rand::SystemRandom;
+use crate::ring_like::signature::KeyPair as RingKeyPair;
+use crate::ring_like::signature::{self, EcdsaKeyPair, Ed25519KeyPair, RsaEncoding, RsaKeyPair};
```
cpu, line 11: WDYT about folding the L10 import into L11's grouping since they share namespaces.
```suggestion
use crate::ring_like::signature::{
	self, EcdsaKeyPair, Ed25519KeyPair, KeyPair as RingKeyPair, RsaEncoding, RsaKeyPair,
};
```

.github/workflows/ci.yml
cpu, line 129: It seems to me that since the choice of crypto provider is done via features, and *ring* overrides aws-lc-rs, that we need CI tasks for `--no-default-features --features=aws_lc_rs` to prevent bitrot there. The `--all-features` and default features builds will both be testing *ring*.

rcgen/src/ring_like.rs
```diff
@@ -0,0 +1,38 @@
+#[cfg(feature = "ring")]
+pub(crate) use ring::*;
+
+#[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
+pub(crate) use aws_lc_rs::*;
+
+use crate::{error::ExternalError, Error};
```
cpu, line 7: in the main Rustls crates we'd write these imports:

```suggestion
use crate::error::ExternalError;
use crate::Error;
```

Not blocking feedback here, just sharing in case you end up contributing to repos outside of this one in the Rustls org.

rcgen/src/ring_like.rs
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	#[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
+	{
+		kp.public_modulus_len()
+	}
+}
+
+#[cfg(not(any(feature = "ring", feature = "aws_lc_rs")))]
+compile_error!("At least one of the features among 'ring' and 'aws_lc_rs' must be activated");
```
cpu, line 38: I think this might be clearer written as:
```suggestion
compile_error!("At least one of the 'ring' or 'aws_lc_rs' features must be activated");
```

cpu (commented):
Thanks for taking this on :rocket: 

I think it would be interesting to consider how the `KeyPair` APIs might need to change to allow runtime selection of the backing crypto library but it might make sense to hash that out in a follow-up issue.

djc (approved):
Nothing to add to @cpu's feedback, nice work!

.github/workflows/ci.yml
BiagioFesta: `cargo test` is already another step. Probably better to have a proper test on CI with `aws_lc_rs`. I am going to change accordingly

rcgen/src/ring_like.rs
BiagioFesta: Actually I like to know more about it. I never understood if and what is the general common convention for imports.

Personally (in my projects), I like to use `imports_granularity = "Item"` ([from cargo +nightly fmt](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=import#imports_granularity)) (an example [here](https://github.com/BiagioFesta/wtransport/blob/master/wtransport/src/config.rs#L35-L53)).

But that is just driven by my *personal* feeling of not having breaking lines and strange indentations in the import sections; I found it more.. "clean". However, I understand that sometimes is a waste of lines...

cpu: There's a PR in the Rustls repo to automate the style: https://github.com/rustls/rustls/pull/1577

I don't know much about the history of how the style was arrived at but if you leave a comment there djc or ctz might have some thoughts !

cpu (commented):
Thanks for iterating on this. There's just one last thing to fix:

The [Windows CI task](https://github.com/rustls/rcgen/actions/runs/7208504335/job/19637577766?pr=201) fails right now with the `aws-lc-rs` backend feature because it needs `nasm` as a build time dependency on that OS. Could you add [something like this](https://github.com/rustls/rustls/blob/75edb20a1e6a894089516053348b6137a425b9b4/.github/workflows/build.yml#L42-L44) to the relevant bit of CI?

There's an [upstream issue](https://github.com/aws/aws-lc-rs/issues/285) that I hope will let us remove this in the future.

cpu (approved):

cpu:
Thanks!

PR #202, opened 2023-12-16
Release 0.12.0

### LINKED ISSUE #199: 0.12.0 release planning
Let's gather some context for release planning since there have been a few requests.

Some items that are close to being done that I think we should block on:

* [x] https://github.com/rustls/rcgen/pull/197
* [x] https://github.com/rustls/rcgen/pull/183
* [ ] https://github.com/rustls/rcgen/pull/190 - requires additional maintainer review.
* [ ] ~https://github.com/rustls/rcgen/pull/170 - requires update for review feedback.~ edit: consensus here is to skip this one.

Some items that will require a major version increase that are not as close to being completed:

* https://github.com/rustls/rcgen/issues/155 
* https://github.com/rustls/rcgen/issues/150 
* https://github.com/rustls/rcgen/issues/122 - the three above issues are addressed by [a WIP PR](https://github.com/rustls/rcgen/pull/164) that needs to be completed.
* https://github.com/rustls/rcgen/issues/73 - unstarted, could probably be rolled into the WIP PR mentioned above.
* https://github.com/rustls/rcgen/issues/62 - unstarted, but a significant foot-gun that should be addressed soon.
* https://github.com/rustls/rcgen/issues/181 - unstarted
* https://github.com/rustls/rcgen/issues/68 - unstarted
* https://github.com/rustls/rcgen/issues/81- unstarted
* switching to [`pki-types`](https://github.com/rustls/pki-types) and [`rustls-pemfile`](https://crates.io/crates/rustls-pemfile) - unstarted - needs discussion.

In my view there's a good amount of work left to do that will require a comparable version bump. I think this means we have a choice between:

1. Cutting a 0.12.0 release that includes only the changes that are close to being completed, and then later (_O(~weeks)_) cutting a 0.13.0 that addresses some of the remaining work.
2. Holding 0.12.0 for longer, and addressing more of these changes in one go and then trying to support 0.12.0 with fewer future API changes.

est31:
For release notes see the changelog.

fixes #199

### DIFF
--- a/rcgen/CHANGELOG.md
+++ b/rcgen/CHANGELOG.md
@@ -4,9 +4,9 @@
 ## Release 0.12.0 - December 16, 2023
 
 - Rename `RcgenError` to `Error`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
-- Lock down the `From` impls for `Error` and make it future-proof.
+- `Error::PemError` has been changed to hold a `String` value. The `From<_>` traits for external error types have been removed. This allows rcgen to update dependent crates without impacting API consumers.
 - Upgrade to `ring` `v0.17`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
-- Make dependency on `ring` optional and allow usage of `aws_lc_rs`. Contributed by [BiagioFesta](https://github.com/BiagioFesta).
+- Make dependency on `ring` optional and allow usage of `aws-lc-rs`. Ring continues to be used as the default. You may opt-in to using `aws-lc-rs` instead by disabling default features and specifying the `aws_lc_rs` feature.  Contributed by [BiagioFesta](https://github.com/BiagioFesta).
 - Add `Ia5String` support for `DistinguishedName`s.
 
 ## Release 0.11.3 - October 1, 2023
--- a/rcgen/CHANGELOG.md
+++ b/rcgen/CHANGELOG.md
@@ -1,10 +1,13 @@
 
 # Changes
 
-## Unreleased
+## Release 0.12.0 - December 16, 2023
 
-- Rename `RcgenError` to `Error` to avoid stuttering when used fully-qualified via `rcgen::`.
-- Upgrade to `ring` `v0.17`.
+- Rename `RcgenError` to `Error`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
+- Lock down the `From` impls for `Error` and make it future-proof.
+- Upgrade to `ring` `v0.17`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
+- Make dependency on `ring` optional and allow usage of `aws_lc_rs`. Contributed by [BiagioFesta](https://github.com/BiagioFesta).
+- Add `Ia5String` support for `DistinguishedName`s.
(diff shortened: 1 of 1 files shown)


rcgen/CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 
 # Changes
 
-## Unreleased
+## Release 0.12.0 - December 16, 2023
```
cpu, line 4: I think this is missing https://github.com/rustls/rcgen/pull/197 contributed by Brocar.

Proposed text:

- Add `KeyIdMethod::PreSpecified` variant to allow specifying an exact subject key identifier (SKI) instead of a method for generating it based on the subject key. `CertificateParams::from_ca_cert_pem` has been updated to use the pre-existing SKI when building params from an existing CA cert.

rcgen/CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
 
-- Rename `RcgenError` to `Error` to avoid stuttering when used fully-qualified via `rcgen::`.
-- Upgrade to `ring` `v0.17`.
+- Rename `RcgenError` to `Error`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
+- Lock down the `From` impls for `Error` and make it future-proof.
```
cpu, line 7: I think we can give more information here. I would also try to put the user-facing changes up-front and then describe the rationale. What about:

```suggestion
- `Error::PemError` has been changed to hold a `String` value. The `From<_>` traits for external error types have been removed. This allows rcgen to update dependent crates without impacting API consumers.
```

rcgen/CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
+- Upgrade to `ring` `v0.17`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
+- Make dependency on `ring` optional and allow usage of `aws_lc_rs`. Contributed by [BiagioFesta](https://github.com/BiagioFesta).
```
cpu, line 9: I think this would be most helpful if we explicitly say how to use aws_lc_rs:
```suggestion
- Make dependency on `ring` optional and allow usage of `aws-lc-rs`. Ring continues to be used as the default. You may opt-in to using `aws-lc-rs` instead by disabling default features and specifying the `aws_lc_rs` feature.  Contributed by [BiagioFesta](https://github.com/BiagioFesta).
```

cpu (commented):
Thanks for putting this together :+1:

est31:
There is no changes to the version in `Cargo.toml` in this PR. I suppose #183 had to increase it due to the breaking changes. Maybe we should switch to `0.13.0-dev` after this PR?

cpu:
As a small note, in the future I'd like a chance to approve the changelog before the release is published if I leave feedback. If I don't reply right away it seems reasonable to roll forward without waiting.

> Maybe we should switch to 0.13.0-dev after this PR?

That sounds fine to me. In the other Rustls repos we tend to use -alpha.0 but I don't feel strongly.

est31:
> I'd like a chance to approve the changelog before the release is published if I leave feedback

Sorry, noted!

PR #204, opened 2023-12-19
Allow building without the `pem` crate feature

daxpedda:
Currently trying to build like this:
```
cargo build -p rcgen --no-default-features --features ring
```
will fail because it can't find `pem::PemError` or `Error::PemError`.

Simply adding a `cfg` guard on this implementation fixes the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,13 @@ jobs:
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        features:
+          - --all-features
+          - --no-default-features --features ring
+          - --no-default-features --features aws_lc_rs
+          - --no-default-features --features aws_lc_rs,pem
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


djc:
Would you mind adding a CI job that tests for this, to help us avoid regressing this again?

.github/workflows/ci.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         with:
           components: clippy
       - run: cargo clippy --all-features --all-targets
+      - run: cargo clippy --no-default-features --features ring --all-targets
+      - run: cargo clippy --no-default-features --features aws_lc_rs --all-targets
```
daxpedda, line 43: Should I add these combinations to `cargo test` as well?

daxpedda: I went ahead and parallelized the Clippy runs as well.
The `cargo test`s seem to not be straightforward to add to a matrix, so I refrained from touching those.

djc (approved):
Thanks, looks great!

cpu (approved):
Thank you

cpu:
> Clippy Expected — Waiting for status to be reported

This is from the branch protection rules falling out of sync with the updated matrix-based task name. I think we should merge without waiting for requirements to be met and then fix those rules.

est31:
I think this warrants a 0.12.1 bugfix release.

PR #205, opened 2024-01-10
Rework Certificate issuance API, make DER/PEM serialization stable

### LINKED ISSUE #62: serialize_der() regenerates the certificate
I am using rcgen to generate a certificate, then I'm sharing the fingerprint of the certificate with a peer, and then I'm using a library to communicate with that peer. The library requires me to provide the certificate & private key in PEM format.

I've been debugging for a while why the fingerprint I generate, like this:

```rust
let fingerprint = ring::digest::digest(&ring::digest::SHA256, &cert.serialize_der()?);
```

... does not match the fingerprint the peer expects to receive, nor the fingerprint OpenSSL reports when I supply it with the PEM-encoded certificate obtained with `serialize_pem()`.

After reading the discussion in #28, and looking at the code, I think I understand what's going on — `serialize_der()` and `serialize_pem()` are not actually just serializing; they are in fact *generating* the certificate each time they are called, so the random components are different between the DER serialization and the PEM serialization.

Ideally the certificate would be generated when the `Certificate` struct is created, so that it can be repeatedly serialized — perhaps into different formats — without regenerating it each time.

If that's not possible, the functions should either be renamed or at least the mis-naming documented clearly to save the time of anyone else who encounters this.

Right now, `from_params()` is documented as "Generates a new certificate" and `serialize_der()` is documented as "Serializes the certificate to the binary DER format", when in reality `from_params()` just stores the params (and does keygen if needed), and `serialize_der()` etc actually do the certificate generation.

cpu:
Previously a `Certificate` was a container for `CertificateParams` and a `KeyPair`, most commonly created from a `CertificateParams` instance. Serializing the `Certificate` (either as self signed with `serialize_pem` or `serialize_der`, or signed by an issuer with `serialize_pem_with_signer` or `serialize_der_with_signer`) would issue a certificate and produce the serialized form in one operation. The net result is that if a user wanted both DER and PEM serializations they would likely call `serialize_der(_with_signer)` and then `serialize_pem(_with_signer)` and mistakenly end up with the encoding of two distinct certificates, not the PEM and DER encoding of the same cert. Since the `KeyPair` contains a private key this API design also meant that the `Certificate` type had to be handled with care, and `Zeroized`.

This branch reworks the issuance API and `Certificate` type to better match user expectation: `Certificate` is only public material and represents an issued certificate that can be serialized in a stable manner in DER or PEM encoding.

I recommend reviewing this commit-by-commit, but here is a summary of the most notable API changes:

* `Certificate::from_params` and `Certificate::serialize_der` and `Certificate::serialize_pem` for issuing a self-signed certificate are replaced with `Certificate::generate_self_signed()` and calling `der` or `pem` on the result.
* `Certificate::from_params` and `Certificate::serialize_der_with_signer` and `Certificate::serialize_pem_with_signer` for issuing a certificate signed by another certificate are replaced with `Certificate::generate()` and calling `der` or `pem` on the result.
* `CertificateSigningRequest::serialize_der_with_signer` and `CertificateSigningRequest::serialize_pem_with_signer` for issuing a certificate from a CSR are replaced with `Certificate::from_request` and calling `der` or `pem` on the result. The `CertificateSigningRequest` type is renamed to `CertificateSigningRequestParams` to better emphasize its role and match the other `*Params` types that already exist.
* Since we now calculate the DER encoding of the certificate at `Certificate` construction time, the `pem` and `der` fns are now infallible. 
* Since `Certificate` no longer holds `KeyPair`, the `generate` fns now expect a `&KeyPair` argument for the signer when issuing a certificate signed by another certificate.
* The generation fns now return a `CertifiedKey` that contains both a `Certificate` and a `KeyPair`. For params that specify a compatible `KeyPair` it is passed through in the `CertifiedKey` as-is. For params without a `KeyPair` a newly generated `KeyPair` is used.

In the future we should look at harmonizing the creation of `CertificateSigningRequest` and `CertificateRevocationList` to better match this updated API. Unfortunately I don't have time to handle that at the moment. Since this API surface is relatively niche compared to the `Certificate` issuance flow it felt valuable to resolve #62 without blocking on this future work.

Resolves https://github.com/rustls/rcgen/issues/62


rcgen/src/lib.rs
```diff
@@ -1593,6 +1585,24 @@ impl Certificate {
 	pub fn serialize_private_key_pem(&self) -> String {
 		self.key_pair.serialize_pem()
 	}
+
+	/// Return a subject [`KeyPair`] to use for the [`CertificateParams`].
+	///
+	/// If `params` do not specify a [`KeyPair`], a new one is generated.
+	/// If `params` specifies a [`KeyPair`], that isn't compatible with the signature
+	/// algorithm specified, an error is returned.
+	fn prepare_key_pair(
```
djc, line 1635: Maybe this could be `KeyPair::prepare()`?

rcgen/src/lib.rs
```diff
@@ -73,6 +73,7 @@ pub type RcgenError = Error;
 pub struct Certificate {
 	params: CertificateParams,
 	key_pair: KeyPair,
+	spki: Vec<u8>,
```
djc, line 75: Nit: I feel like SPKI is not a very accessible acronym, how about writing it out as `subject_public_key_info` in the field name and the type constructors at least (I probably wouldn't expand it in `pub_key_spki`)?

rcgen/src/lib.rs
```diff
@@ -1562,49 +1555,31 @@ impl Certificate {
 		pem::encode_config(&Pem::new("CERTIFICATE", self.der()), ENCODE_CONFIG)
 	}
 	/// Serializes a certificate signing request in binary DER format
-	pub fn serialize_request_der(&self) -> Result<Vec<u8>, Error> {
+	pub fn serialize_request_der(&self, subject_key: &KeyPair) -> Result<Vec<u8>, Error> {
```
djc, line 1605: Might be helpful to add some clarification to the docstrings for `serialize_request_*()` about what `subject_key` and how it's used? I can't think of a better name but also feel like this name is maybe not the most obvious.

djc (approved):
I think this makes the API much better, nice!

rcgen/src/lib.rs
```diff
@@ -69,7 +69,7 @@ pub use crate::sign_algo::SignatureAlgorithm;
 )]
 pub type RcgenError = Error;
 
-/// A self signed certificate together with signing keys
+/// A certificate together with a subject keypair.
```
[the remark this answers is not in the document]
cpu: Fixed.

rcgen/src/lib.rs
```diff
@@ -1352,21 +1339,48 @@ impl CustomExtension {
...
+	/// Pre-specified identifier. The exact given value is used as the key identifier.
 	PreSpecified(Vec<u8>),
 }
 
+impl KeyIdMethod {
+	/// Calculates a key identifier for the provided public key using the key ID method.
+	///
+	/// This key identifier is used in the SubjectKeyIdentifier and AuthorityKeyIdentifier
+	/// X.509v3 extensions.
+	pub(crate) fn calculate<K: PublicKeyData>(&self, pub_key: &K) -> Vec<u8> {
```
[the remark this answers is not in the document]
cpu: > Nits: I don't love the calculate() name for this. Maybe derive() or digest()?

I went with `derive` since the `PreSpecified` variant could be anything and isn't necessarily a digest at all. 

>  Also I'd suggest eliminating the K type param in favor of pub_key: &impl PublicKeyData.

Done, though this later falls away to `&impl AsRef<[u8]>` to accommodate the change in design where we hold the SPKI and not a `KeyPair` or something implementing `PublicKeyData`.

> Alternatively, maybe it makes sense to rename this to write(), taking a writer argument? Would avoid the to_vec() allocation, at least (and this is private API, anyway).

I tried this but unfortunately because `Certificate::get_key_identifier` exists in the public API we need something we can use without a DER writer. We can't create our own writer and then pull the written bytes back out easily because the bytes are written as a DER encoded octet sequence with a preamble. Trying to write a fn that returns a `Vec<u8>` and one that uses a writer without duplicating code wasn't working out well so I jacked out and left it as-is :laughing: I think the extra allocation will have to stand for now.

rcgen/src/lib.rs
cpu: We don't have a `KeyPair` in this case, we have an `Option<KeyPair>` that might be `None`. Most of this helper exists to deal with the `None` case, there doesn't seem to be much of value to lift into a `KeyPair` fn for the non-none case so I left this as is.

rcgen/src/lib.rs
```diff
@@ -1593,6 +1585,24 @@ impl Certificate {
...
+	///
+	/// If `params` do not specify a [`KeyPair`], a new one is generated.
+	/// If `params` specifies a [`KeyPair`], that isn't compatible with the signature
+	/// algorithm specified, an error is returned.
+	fn prepare_key_pair(
+		key_pair: &mut Option<KeyPair>,
+		sig_alg: &'static SignatureAlgorithm,
+	) -> Result<KeyPair, Error> {
+		match key_pair.take() {
+			None => KeyPair::generate(sig_alg),
```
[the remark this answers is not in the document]
cpu: Fixed.

rcgen/src/lib.rs
cpu: Good call. Fixed.

rcgen/src/lib.rs
cpu: I added more clarification here in a separate commit. I also explicitly called out that this API still has the footgun we're fixing in this branch for the `Certificate` type (e.g. calling `serialize_request_der` and then `serialize_request_pem` from the same `Certificate` isn't going to do what you expect).

rcgen/src/lib.rs
```diff
@@ -1506,7 +1505,7 @@ impl Certificate {
 	/// If you want to control the certificate subject [`KeyPair`] or the randomness used to
 	/// generate it, set the [`CertificateParams::key_pair`] field ahead of time before calling
 	/// this function.
-	pub fn generate_self_signed(mut params: CertificateParams) -> Result<Self, Error> {
+	pub fn generate_self_signed(mut params: CertificateParams) -> Result<(Self, KeyPair), Error> {
```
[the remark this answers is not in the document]
cpu: I think that makes sense. I introduced a new `CertifiedKey` type (I think this is the best name, and has some precedent in Rustls).

rcgen/src/lib.rs
```diff
@@ -1536,6 +1536,28 @@ impl Certificate {
 		)?;
 		Ok((Certificate { params, spki, der }, key_pair))
 	}
+	/// Generate a new certificate using the certificate signing request parameters, signed by
+	/// the provided issuer.
+	pub fn generate_request(
```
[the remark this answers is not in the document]
cpu: > I would propose that this becomes Certificate::from_request() instead.

Done.

> Also, we have CertificateParams and Certificate but in this case CertificateSigningRequest plays the role of the *Params. Should we append that suffix here to make this more obvious? Or another suffix, maybe Input?

That's a good point - I switched the type to `CertificateSigningRequestParams` for best alignment with `CertificateParams` and `CertificateRevocationListParams`.

rcgen/src/csr.rs
```diff
@@ -23,15 +21,15 @@ impl PublicKeyData for PublicKey {
 	}
 }
 
-/// Data for a certificate signing request
-pub struct CertificateSigningRequest {
+/// Parameters for a certificate signing request
+pub struct CertificateSigningRequestParams {
```
djc, line 25: Tiny nit: would be nicer to extract the rename into a separate commit, IMO.

djc (approved):

rcgen/src/lib.rs
djc: We're not using any data from Certificate so it seems weird for this to be part of the Certificate impl. Plus it yields a Result<KeyPair, Error> so it does feel like a kind of constructor for KeyPair except that you might inject a pre-existing one to verify its algorithm. Maybe call it KeyPair::validate_or_generate()?

cpu: You've won me over :+1: I updated the implementation to associate this with the `KeyPair` type as `KeyPair::validate_or_generate`.

rcgen/src/csr.rs
cpu: Pulled into a separate commit.

est31 (approved):
looks great!

est31:
I also really like the `CertifiedKey` struct. Maybe the name is not ideal but we can always change it in a followup PR.

I'm also thinking about a `struct CertifiedKeyRef<'a,'b> { pub cert: &'a Certificate, pub key: &'b KeyPair }`, which can be used in the cases where references for both the key and the cert are wanted, and a conversion function like `impl CertifiedKey { fn ref(&self) -> CertifiedKeyRef<'_, '_> {...}}` . I think that's also best for a followup.

PR #206, opened 2024-01-11
choose a crypto_provider for rustls_cert_gen

stormshield-gt:
From the release 0.12 of rcgen https://github.com/rustls/rcgen/pull/202 , we must now choose `ring` or `aws_lc_rc` as a feature.
Because `rustls_cert_gen` has `default-features = false`,  its build is currently broken.

~This PR activate  `aws_lc_rc` feature, because soon it will get RSA key generation support https://github.com/aws/aws-lc-rs/issues/296 , so it will provide more options.~ Edit: we switched to using ring to match the rcgen default.

But we could also change it to default to whatever `rcgen` default or use `ring` as default depending on your preferences.

### DIFF
--- a/rustls-cert-gen/Cargo.toml
+++ b/rustls-cert-gen/Cargo.toml
@@ -7,7 +7,7 @@ edition.workspace = true
 keywords.workspace = true
 
 [dependencies]
-rcgen = { path = "../rcgen", default-features = false, features = ["pem"] }
+rcgen = { path = "../rcgen", default-features = false, features = ["pem", "ring"] }
 bpaf = { version = "0.9.5", features = ["derive"] }
 pem = { workspace = true }
 ring = { workspace = true }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,21 @@ jobs:
     - name: Run the tests with aws_lc_rs backend enabled
       run: cargo test --verbose --no-default-features --features aws_lc_rs,pem --all-targets
 
+  # Build rustls-cert-gen as a standalone package, see this PR for why it's needed:
+  # https://github.com/rustls/rcgen/pull/206#pullrequestreview-1816197358
+  build-rustls-cert-gen-standalone:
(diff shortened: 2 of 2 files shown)


cpu:
Thanks for the PR!

> Because rustls_cert_gen has default-features = false, its build is currently broken.

It sounds like there's CI coverage missing for this tool that would detect this broken state :-( Would you be willing to update your PR to add that coverage as well?

stormshield-gt:
It's a feature unification problem, when we build at a workspace level, all features get unified so the `ring` feature from `rcgen` is also added to `rustls_cert_gen` that's why the CI doesn't catch it.

One solution is to add a CI job with `cargo build --package rustls-cert-gen` or to use an external tools like `cargo hack`, as shown in this [thread](https://stackoverflow.com/a/74293494)

Which solution do you prefer?

cpu:
> It's a feature unification problem

Aha, that makes sense :bulb: Thank you

> One solution is to add a CI job with cargo build --package rustls-cert-gen

That sounds good to me. We're using `cargo hack` in some other Rustls projects so I'm not opposed to that solution if another maintainer prefers it, but I think we can start with the simplest option.

rustls-cert-gen/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition.workspace = true
 keywords.workspace = true
 
 [dependencies]
-rcgen = { path = "../rcgen", default-features = false, features = ["pem"] }
+rcgen = { path = "../rcgen", default-features = false, features = ["pem", "aws_lc_rs"] }
```
cpu, line 10: I think we should use `ring` as the default for now and consider a project-wide switch at some point in the future.

stormshield-gt:
We see that's the new job failed the CI as expected, I will now push the fix

djc (approved):

cpu (approved):
Thanks!

cpu:
This feels small and obvious enough to merge without blocking on an additional review. If there's any post-merge feedback I will commit to implementing it.

PR #207, opened 2024-01-12
Support compiling without cryptography primitives

corrideat:
This makes all of the crypto libraries optional. My use case requires simply generating the binary ASN.1 structures and all cryptography is handled externally with HSMs.

This change removes unnecessary code and build complexity (for example, I'm targeting WASM, and I need to take extra steps to build the dependencies I don't need).

This change may still require some adjustments.

### DIFF
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -9,8 +9,9 @@ use crate::error::ExternalError;
 use crate::ring_like::error as ring_error;
 #[cfg(feature = "random")]
 use crate::ring_like::rand::SystemRandom;
+#[cfg(feature = "random")]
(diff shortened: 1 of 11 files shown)


rcgen/src/key_pair.rs
```diff
@@ -178,6 +193,13 @@ impl KeyPair {
 	}
 
 	/// Generate a new random key pair for the specified signature algorithm
+	#[cfg(not(feature = "crypto"))]
+	pub fn generate(alg: &'static SignatureAlgorithm) -> Result<Self, Error> {
```
djc, line 197: It doesn't seem like a great idea to generate an impl that can only error?

djc (commented):
This doesn't seem to entirely make sense to me. I don't see a clear use case for wanting `random` but not `crypto`?

PR #208, opened 2024-01-12
Support compiling without cryptography primitives

corrideat:
This makes all of the crypto libraries optional. My use case requires simply generating the binary ASN.1 structures and all cryptography is handled externally with HSMs.

This change removes unnecessary code and build complexity (for example, I'm targeting WASM, and I need to take extra steps to build the dependencies I don't need).

Note to the maintainer: this is a clean (and ideally better) rework of a previous PR, #207. Because I deleted the older and messier branch, the previous PR couldn't be updated, so I closed it.

### DIFF
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -21,6 +21,10 @@ required-features = ["pem"]
 name = "sign-leaf-with-ca"
 required-features = ["pem", "x509-parser"]
 
+[[example]]
+name = "simple"
+required-features = ["crypto"]
+
 [dependencies]
 aws-lc-rs = { version = "1.0.0", optional = true }
 yasna = { version = "0.5.2", features = ["time", "std"] }
@@ -31,9 +35,10 @@ x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
-default = ["pem", "ring"]
-aws_lc_rs = ["dep:aws-lc-rs"]
-ring = ["dep:ring"]
+default = ["crypto", "pem", "ring"]
+crypto = []
+aws_lc_rs = ["crypto", "dep:aws-lc-rs"]
+ring = ["crypto", "dep:ring"]
 
 
 [package.metadata.docs.rs]
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -1,30 +1,38 @@
 #[cfg(feature = "pem")]
 use pem::Pem;
+#[cfg(feature = "crypto")]
 use std::convert::TryFrom;
 use std::fmt;
 use yasna::DERWriter;
 
+#[cfg(any(feature = "crypto", feature = "pem"))]
 use crate::error::ExternalError;
-use crate::ring_like::error as ring_error;
-use crate::ring_like::rand::SystemRandom;
-use crate::ring_like::signature::{
-	self, EcdsaKeyPair, Ed25519KeyPair, KeyPair as RingKeyPair, RsaEncoding, RsaKeyPair,
+#[cfg(feature = "crypto")]
+use crate::ring_like::{
+	error as ring_error,
+	rand::SystemRandom,
+	signature::{
+		self, EcdsaKeyPair, Ed25519KeyPair, KeyPair as RingKeyPair, RsaEncoding, RsaKeyPair,
+	},
+	{ecdsa_from_pkcs8, rsa_key_pair_public_modulus_len},
 };
-use crate::ring_like::{ecdsa_from_pkcs8, rsa_key_pair_public_modulus_len};
-use crate::sign_algo::algo::*;
-use crate::sign_algo::SignAlgo;
+#[cfg(feature = "crypto")]
+use crate::sign_algo::{algo::*, SignAlgo};
 #[cfg(feature = "pem")]
 use crate::ENCODE_CONFIG;
-use crate::{Error, SignatureAlgorithm};
+use crate::{sign_algo::SignatureAlgorithm, Error};
 
 /// A key pair variant
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum KeyPairKind {
 	/// A Ecdsa key pair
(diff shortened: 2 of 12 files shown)


rcgen/src/key_pair.rs
```diff
@@ -211,6 +229,12 @@ impl KeyPair {
 		}
 	}
 
+	/// Generate a new random key pair for the specified signature algorithm
+	#[cfg(not(feature = "crypto"))]
+	pub fn generate(_alg: &'static SignatureAlgorithm) -> Result<Self, Error> {
```
corrideat, line 234: Original comment by @djc: https://github.com/rustls/rcgen/pull/207#discussion_r1450126844

> It doesn't seem like a great idea to generate an impl that can only error?

Would you rather I remove this entirely and the `generate` just isn't available without the `crypto` feature?

rcgen/Cargo.toml
```diff
@@ -31,7 +31,8 @@ x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
-default = ["pem", "ring"]
+default = ["crypto", "pem", "ring"]
+crypto = []
```
corrideat, line 35: Original comment by @djc: https://github.com/rustls/rcgen/pull/207#pullrequestreview-1817771475

> This doesn't seem to entirely make sense to me. I don't see a clear use case for wanting random but not crypto?

Noted. I made it only `crypto` in this case.

rcgen/src/key_pair.rs
est31: yes, that's generally preferable. Rust has strong static checks. the std implementations for wasm are panicking, and now I heard that it is considered to be a mistake.

rcgen/src/key_pair.rs
```diff
@@ -1,30 +1,41 @@
 #[cfg(feature = "pem")]
 use pem::Pem;
+#[cfg(feature = "crypto")]
 use std::convert::TryFrom;
 use std::fmt;
 use yasna::DERWriter;
 
+#[cfg(any(feature = "crypto", feature = "pem"))]
 use crate::error::ExternalError;
+#[cfg(feature = "crypto")]
```
est31, line 10: it would be better to have one use block like `use crate::{...}` instead of copy-pasting this to all instances

rcgen/src/lib.rs
```diff
@@ -1141,17 +1153,28 @@ impl CertificateParams {
 	/// This key identifier is used in the SubjectKeyIdentifier X.509v3 extension.
 	fn key_identifier<K: PublicKeyData>(&self, pub_key: &K) -> Vec<u8> {
 		// Decide which method from RFC 7093 to use
+		#[cfg_attr(
+			not(feature = "crypto"),
+			allow(irrefutable_let_patterns, unused_variables)
```
est31, line 1158: is allowing unused_variables still needed?

rcgen/src/ring_like.rs
```diff
@@ -35,5 +39,5 @@ pub(crate) fn rsa_key_pair_public_modulus_len(kp: &signature::RsaKeyPair) -> usi
 	}
 }
 
-#[cfg(not(any(feature = "ring", feature = "aws_lc_rs")))]
+#[cfg(all(feature = "crypto", not(any(feature = "ring", feature = "aws_lc_rs"))))]
 compile_error!("At least one of the 'ring' or 'aws_lc_rs' features must be activated");
```
est31, line 43: maybe the error could be improved to suggest the alternative of disabling the crypto feature?

rcgen/Cargo.toml
est31: the ring and aws-lc-rs features should probably also depend on `crypto`. It makes no sense to enable ring/aws-lc-rs and not enable `crypto`.

est31 (commented):
generally a good idea. I think we should test this in CI as it seems easy to regress to me.

rcgen/src/key_pair.rs
corrideat: I see. Well, this has already been done.

rcgen/Cargo.toml
corrideat: That's a good point. I can do that.

rcgen/src/key_pair.rs
corrideat: That's a good point too, I've now done this.

rcgen/src/lib.rs
corrideat: Yes, it's needed because of this:

```
let digest_method = match &self.key_identifier_method {
```

Note that:
1. `irrefutable_let_patterns` is needed because only one match expression is used (`KeyIdMethod::PreSpecified(b)`)
2. `unused_variables` is needed because the following block is not used, so `digest_method` is unused.

This could be changed by refactoring perhaps so that the attributes aren't needed.

cpu:
I think this needs a rebase now that the refactoring in #205 has landed. Would you mind tidying up the commit history at the same time? I think all of the commits could be squashed down into one.

corrideat:
@cpu I've now rebased; it probably needs another review just in case.

rcgen/src/key_pair.rs
cpu: Did a commit for this change not get pushed or am I confused about the outcome of the discussion? 

The state of the branch at the moment is that there is still a `generate` impl gated by `#[cfg(not(feature = "crypto"))]` that will only ever produce an error. I read the above discussion as preferring this impl not exist.

cpu (approved):
Generally looks good to me, thanks! 

I had one question/comment about the resolution of the comment thread RE: the `generate()` impl for `not(feature = "crypto")`.

rcgen/src/key_pair.rs
cpu: @corrideat I think this piece of feedback is the only merge blocker. Do you think you'll have a chance to fix it?

est31 (approved):

djc (approved):

cpu (approved):

PR #209, opened 2024-01-14
Add basic support for  Subject Alternative Name OtherName

Tudyx:
Add basic support for `OtherName` in the Subject Alternative Name extension.
`OtherName` is used by smart card certificates and also, for instance, to identify nodes in a peer-to-peer network. 

`OtherName` is defined like this in the [RFC 5280§4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#page-38):

```
OtherName ::= SEQUENCE {
        type-id    OBJECT IDENTIFIER,
        value      [0] EXPLICIT ANY DEFINED BY type-id }
```

Which is really close to `AttributeTypeAndValue` that we already support ([RFC 5280§4.1.2.4](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4))

```
 AttributeTypeAndValue ::= SEQUENCE {
     type     AttributeType,
     value    AttributeValue }

AttributeType ::= OBJECT IDENTIFIER

AttributeValue ::= ANY -- DEFINED BY AttributeType
```
So I  took the same approach while implementing it.

I don't try to map the OID with a specific ASN.1 object, as users could define their own.

This is the [same approach as  OpenSSL](https://www.openssl.org/docs/man3.2/man5/x509v3_config#Subject-Alternative-Name) where  users provide  the OID, the format and its content.


est31:
OpenSSL seems to allow general ASN.1 for OtherName while here we restrict ourselves to a bunch of string types. Should we instead just take binary encoded ASN.1 plus a custom specified oid?

Tudyx:
This is the same tradeoff that is made for [`DnValue`](https://docs.rs/rcgen/latest/rcgen/enum.DnValue.html) which could also technically be an arbitrary ASN.1 object. As `OtherName`  is used for setup a name, the vast majority of the time user will just want to use a string type, so it makes sense to facilitate that path. 
The `enum` is non-exhaustive so in the future we might add another variant that takes an arbitrary ASN.1. But by doing so we will probably re-expose a dependency into our public API so it should be done carefully.
The current proposition unlocks the vast majority of use cases while being future proof.

I don’t feel strongly about that, and will make the required changes if we think exposing an arbitrary  ASN.1 struct is the best solution.

cpu:
I haven't review the branch yet but in general I think we should address https://github.com/rustls/rcgen/issues/181 before making the problem worse by introducing more types that carry tagged string types that don't enforce that the content matches the type's limitations.

Tudyx:
I've rebased on the latest commits to be able to use the new string types (https://github.com/rustls/rcgen/pull/214).

What I propose in this MR is the same tradeoff that the one we are making for [`DnValue`](https://docs.rs/rcgen/latest/rcgen/enum.DnValue.html)  `enum`.  
For instance,  [`DnValue`](https://docs.rs/rcgen/latest/rcgen/enum.DnValue.html) is used to represent the **issuer name** which is defined by the following ASN.1 in the [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4):

```
Name ::= CHOICE { -- only one possibility for now --
     rdnSequence  RDNSequence }

   RDNSequence ::= SEQUENCE OF RelativeDistinguishedName

   RelativeDistinguishedName ::=
     SET SIZE (1..MAX) OF AttributeTypeAndValue

   AttributeTypeAndValue ::= SEQUENCE {
     type     AttributeType,
     value    AttributeValue }

   AttributeType ::= OBJECT IDENTIFIER

   AttributeValue ::= ANY -- DEFINED BY AttributeType
```
Technically,  [`DnValue`](https://docs.rs/rcgen/latest/rcgen/enum.DnValue.html) is incorrect here because the issuer value can be any ASN.1 type. But has [stated by the RFC](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4):

> The type of
   the component AttributeValue is determined by the AttributeType; in
   general it will be a DirectoryString.

  > The DirectoryString type is defined as a choice of PrintableString,
   TeletexString, BMPString, UTF8String, and UniversalString.

Most of the time the issuer value will just be a string, so that makes sense we make variants for that,  it will cover most of the user's use cases.  [`DnValue`](https://docs.rs/rcgen/latest/rcgen/enum.DnValue.html) is non-exhaustive, so, if in the future we want to add the possibility to provide an arbitrary ASN.1 type, we can do it without breaking users' code.

That's exactly what I propose for `subjectAltName`, I also think strings will cover most of the use cases. 

For my use cases (p2p), we need an UTF-8 string. For  smart cards, this the same thing (extracted from [Microsoft doc](https://learn.microsoft.com/en-us/troubleshoot/windows-server/windows-security/enabling-smart-card-logon-third-party-certification-authorities#configuration-instructions)):

> Subject Alternative Name = Other Name: Principal Name= (UPN). For instance:
UPN = user1@name.com
The UPN OtherName OID is: "1.3.6.1.4.1.311.20.2.3"
The UPN OtherName value: Must be ASN1-encoded UTF8 string

Note that this OID is also used even for non-Windows OS, see [this tutorial from Red Hat](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_smart_card_authentication/configuring-and-importing-local-certificates-to-a-smart-card_managing-smart-card-authentication).

> **Note:** after reflection I just kept the UTF-8 variant for the `OtherNameValue`, as I'm not aware of any uses cases for the other ASN.1 strings.

rcgen/src/lib.rs
```diff
@@ -172,7 +190,25 @@ fn ip_addr_from_octets(octets: &[u8]) -> Result<IpAddr, Error> {
 impl SanType {
 	#[cfg(feature = "x509-parser")]
 	fn try_from_general(name: &x509_parser::extensions::GeneralName<'_>) -> Result<Self, Error> {
+		use x509_parser::der_parser::asn1_rs::{self, FromDer, Tag, TaggedExplicit};
 		Ok(match name {
+			x509_parser::extensions::GeneralName::OtherName(oid, value) => {
```
djc, line 195: Nit: let's move this arm last, since it's long and probably rarer than the others.

rcgen/src/lib.rs
```diff
@@ -856,6 +894,22 @@ impl CertificateParams {
 					writer.next().write_tagged_implicit(
 						Tag::context(san.tag()),
 						|writer| match san {
+							SanType::OtherName((oid, value)) => {
```
djc, line 897: Nit: move this arm last.

rcgen/src/lib.rs
```diff
@@ -151,13 +151,31 @@ const ENCODE_CONFIG: pem::EncodeConfig = {
...
 pub enum SanType {
+	OtherName((Vec<u64>, OtherNameValue)),
 	/// Also known as E-Mail address
 	Rfc822Name(Ia5String),
 	DnsName(Ia5String),
 	URI(Ia5String),
 	IpAddress(IpAddr),
 }
 
+/// An other name entry
```
djc, line 162: Add some context here from your PR description maybe? Like links to the RFC and context on the choice of the supported type(s).

rcgen/src/lib.rs
```diff
@@ -151,13 +151,31 @@ const ENCODE_CONFIG: pem::EncodeConfig = {
 #[non_exhaustive]
 /// The type of subject alt name
 pub enum SanType {
+	OtherName((Vec<u64>, OtherNameValue)),
```
djc, line 154: Move this last, too.

djc (approved):
This seems reasonable to me.

rcgen/src/lib.rs
Tudyx: Good idea, I've added some context and links

cpu:
Just leaving a note that I haven't forgotten about this PR. I will aim to review it this week. Thanks!

rcgen/src/lib.rs
```diff
@@ -156,6 +156,29 @@ pub enum SanType {
...
+
+/// An OtherName value, defined in [RFC 5280§4.1.2.4].
+///
+/// Technically, this could be any ASN.1 types but for now we cover only UTF-8 strings
+/// as it will cover most of use cases, for instance smart cards.
+///
+/// [RFC 5280§4.1.2.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+#[non_exhaustive]
+pub enum OtherNameValue {
```
est31, line 170: The RFC calls this `AnotherName`. Maybe we should use that instead?

est31 (approved):

rcgen/src/lib.rs
Tudyx: Actually, they use both. In the main RFC ( [RFC 5280§4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#page-38)) they use `OtherName` and in the [appendix](https://datatracker.ietf.org/doc/html/rfc5280#page-128) they use `AnotherName`, they also justify this in the appendix:

> -- AnotherName replaces OTHER-NAME ::= TYPE-IDENTIFIER, as
-- TYPE-IDENTIFIER is not supported in the '88 ASN.1 syntax

In [OpenSSL](https://www.openssl.org/docs/man1.0.2/man5/x509v3_config.html) they use `OtherName`, in our dependencies (for instance [x509-parser](https://docs.rs/x509-parser/latest/x509_parser/extensions/enum.GeneralName.html#variant.OtherName)), they also use it. I think we should stick with `OtherName`.

est31: Fair points! :+1:

rcgen/src/lib.rs
```diff
@@ -156,6 +156,29 @@ pub enum SanType {
 	DnsName(Ia5String),
 	URI(Ia5String),
 	IpAddress(IpAddr),
+	OtherName((Vec<u64>, OtherNameValue)),
+}
+
+/// An OtherName value, defined in [RFC 5280§4.1.2.4].
+///
+/// Technically, this could be any ASN.1 types but for now we cover only UTF-8 strings
+/// as it will cover most of use cases, for instance smart cards.
```
cpu, line 165: I have some editorial suggestions for this comment:

* I think 'OtherName' should be in backticks as a technical reference.
* I think we can rephrase why the value is limited to UTF-8 here, and give more information about the use case you referenced.

What do you think about this wording:

```suggestion
/// An `OtherName` value, defined in [RFC 5280§4.1.2.4].
///
/// While the standard specifies this could be any ASN.1 type rcgen limits
/// the value to a UTF-8 encoded string as this will cover the most common
/// use cases, for instance smart card user principal names (UPN).
///
/// [RFC 5280§4.1.2.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4
```

rcgen/src/lib.rs
```diff
@@ -183,13 +207,31 @@ impl SanType {
...
+					.map_err(|_| Error::CouldNotParseCertificate)?;
+				let other_name = other_name.into_inner();
+
+				let data = other_name.data;
+				let try_str =
+					|data| std::str::from_utf8(data).map_err(|_| Error::CouldNotParseCertificate);
+				let other_name_value = match other_name.tag() {
+					Tag::Utf8String => OtherNameValue::Utf8String(try_str(data)?.to_owned()),
+					_ => return Err(Error::CouldNotParseCertificate),
+				};
```
cpu, line 223: nit: I think there are some intermediate bindings (and the `try_str` closure) that could be inlined to make this more succint:

```suggestion
				let other_name_value = match other_name.tag() {
					Tag::Utf8String => OtherNameValue::Utf8String(
						std::str::from_utf8(other_name.data)
							.map_err(|_| Error::CouldNotParseCertificate)?
							.to_owned(),
					),
					_ => return Err(Error::CouldNotParseCertificate),
				};
```

rcgen/src/lib.rs
```diff
@@ -865,6 +908,22 @@ impl CertificateParams {
...
+									writer.next().write_tagged(
+										Tag::context(0),
+										|writer| match value {
+											OtherNameValue::Utf8String(s) => {
+												writer.write_utf8_string(s)
+											},
+										},
+									);
+								});
+							},
```
cpu, line 926: A couple suggestions here:

* let's inline the `oid` binding - it doesn't get referenced outside of the one call to `write_oid`.

* I think to make it easier to extend the `OtherNameValue` enum it might make sense to pull out the code for serializing the value into a `write_der(&self, writer: DERWriter)` fn on `OtherNameValue`. We've been doing that with other types (e.g. `CrlDistributionPoint::write_der`, `PublicKeyData::serialize_public_key_der`) and I think it makes for easier to follow code.

Something like:

```
impl OtherNameValue {
	fn write_der(&self, writer: DERWriter) {
		writer.write_tagged(Tag::context(0), |writer| match self {
			OtherNameValue::Utf8String(s) => writer.write_utf8_string(s),
		});
	}
}
``` 

which would let us write this section as:

```suggestion
								// otherName SEQUENCE { OID, [0] explicit any defined by oid }
								// https://datatracker.ietf.org/doc/html/rfc5280#page-38
								writer.write_sequence(|writer| {
									writer.next().write_oid(&ObjectIdentifier::from_slice(&oid));
									value.write_der(writer.next());
								});
```

It's a bit unfortunate that we define `TAG_OTHER_NAME` as part of `SanType::tag()` but can't use it in the serialization of the other name value (instead it's hardcoded as `0`), but I think addressing that will require a larger refactor that pulls out the san serialization into a fn on `SanType` and that feels like too much to foist on you here.

cpu (commented):
Thanks! I had a few comments but nothing that I think will be a blocker.

rcgen/src/lib.rs
Tudyx: Thanks that's clearer this way!

rcgen/src/lib.rs
Tudyx: Thanks that's cleaner this way

rcgen/src/lib.rs
Tudyx: Done! ;)

cpu (approved):
Thank you :-)

PR #210, opened 2024-01-14
add support for other oid in the extended key usage

Tudyx:
As stated in [RFC 5280§4.2.1.12](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12) in

> Key purposes may be defined by any organization with a need.  Object
   identifiers used to identify key purposes MUST be assigned in
   accordance with IANA or ITU-T Recommendation X.660 [[X.660](https://datatracker.ietf.org/doc/html/rfc5280#ref-X.660)].

This corresponds to the other variant of the  [x509_parser::extensions::ExtendedKeyUsage](https://docs.rs/x509-parser/latest/x509_parser/extensions/struct.ExtendedKeyUsage.html#structfield.other)

This PR add support for others oid in the extended key usage.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -1282,10 +1282,12 @@ pub enum ExtendedKeyUsagePurpose {
 	TimeStamping,
 	/// id-kp-OCSPSigning
 	OcspSigning,
+	/// A custom purpose not from the pre-specified list of purposes
+	Other(Vec<u64>),
 }
 
 impl ExtendedKeyUsagePurpose {
-	fn oid(&self) -> &'static [u64] {
+	fn oid(&self) -> &[u64] {
 		use ExtendedKeyUsagePurpose::*;
 		match self {
 			// anyExtendedKeyUsage
@@ -1297,6 +1299,7 @@ impl ExtendedKeyUsagePurpose {
 			EmailProtection => &[1, 3, 6, 1, 5, 5, 7, 3, 4],
 			TimeStamping => &[1, 3, 6, 1, 5, 5, 7, 3, 8],
 			OcspSigning => &[1, 3, 6, 1, 5, 5, 7, 3, 9],
+			Other(oid) => &oid,
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


rcgen/src/lib.rs
```diff
@@ -1278,10 +1278,12 @@ pub enum ExtendedKeyUsagePurpose {
 	TimeStamping,
 	/// id-kp-OCSPSigning
 	OcspSigning,
+	/// Defined by an organization with a need
```
est31, line 1281:
```suggestion
	/// A custom purpose not from the pre-specified list of purposes
```

est31 (approved):
ok except the comment

est31 (approved):

djc (approved):
LGTM!

rcgen/src/lib.rs
```diff
@@ -1851,6 +1854,36 @@ mod tests {
 		assert!(extension.value.any);
 	}
 
+	#[test]
+	fn test_with_extended_key_usages_other() {
```
cpu, line 1858: I'm surprised this test doesn't need a `cfg` gate on the `x509-parser` feature.

cpu (approved):

rcgen/src/lib.rs
stormshield-gt: It's because `x509-parser` [is also a dev dependency](https://github.com/rustls/rcgen/blob/a3831c9c7714684946276c6b10c981280e6a8297/rcgen/Cargo.toml#L50) besides being under a feature flag

cpu: Ah! That makes sense and I see existing tests that are already relying on this fact. :+1:

stormshield-gt:
Are there any blockers for this PR to be merged?

cpu:
> Are there any blockers for this PR to be merged?

It needs to be rebased on `main` and then I think the "Check semver compatibility" CI task failure will resolve and it could be merged.

stormshield-gt:
I think the "Check semver compatibility" job won't resolve because this PR indeed add a variant to a non-exhaustive `enum`. As we have already breaking change in the main branch, I think it's fine.

cpu:
I believe it will resolve because `main` incremented the crate version to 0.13.0 alongside introducing the breaking changes. Since that version isn't released, adding more breaking changes won't trip the semver checks.

cpu (dismissed):

cpu (approved):

PR #212, opened 2024-01-17
Support for borrowed `RemoteKeyPair`

corrideat:
This implements https://github.com/rustls/rcgen/issues/211

PR #213, opened 2024-01-17
External keys

djc:
As an alternative to #212, this goes further in the direction of #205, removing `alg` and `key_pair` from `CertificateParams` and requiring the caller to pass in a reference to them when signing. This seems to have a number of nice properties:

* `alg` is derived from the passed in `KeyPair`, so key algorithm mismatch errors can no longer occur
* No need for passing in a signing algorithm when parsing from pre-existing parameters or key pairs
* Should make it easy to support long-lived (remote) key pairs

The main downside as far as I can see is that the top-level API gets a bit more complicated, because generating a `KeyPair` must now be done by the caller, and for now we force them to pick a signing algorithm. I think we might mitigate this by adding a no-argument constructor like `generate_default()` (or use `generate()` for the no-argument variant and `generate_for()` for the variant that requires an argument).

Generally, this feels like a clear improvement in the API's design to me.


rcgen/src/lib.rs
```diff
@@ -619,17 +611,10 @@ impl CertificateParams {
 	/// for the presence of the `BasicConstraints` extension, or perform any other
 	/// validation.
 	#[cfg(feature = "x509-parser")]
-	pub fn from_ca_cert_der(ca_cert: &[u8], key_pair: KeyPair) -> Result<Self, Error> {
+	pub fn from_ca_cert_der(ca_cert: &[u8]) -> Result<Self, Error> {
```
cpu, line 614: I think the rustdoc comments on this fn need to be tweaked for the change in return type. Presently it says:
> This function is only of use if you have an existing CA certificate you would like to use to sign a certificate generated by rcgen. By providing the returned [Certificate](https://github.com/rustls/rcgen/pull/213/struct.Certificate.html) and the **passed-through [KeyPair](https://github.com/rustls/rcgen/pull/213/struct.KeyPair.html)**

cpu (approved):
Thanks, I think this is the right direction to be headed :+1: 

> I think we might mitigate this by adding a no-argument constructor like generate_default() (or use generate() for the no-argument variant and generate_for() for the variant that requires an argument).

Maybe we could leave `generate` as the argument required option and implement `Default` for providing something that can work without argument to give you back a `KeyPair` without needing to care about the alg? That feels small enough to tack onto this branch if you were game.

djc:
> Maybe we could leave `generate` as the argument required option and implement `Default` for providing something that can work without argument to give you back a `KeyPair` without needing to care about the alg? That feels small enough to tack onto this branch if you were game.

I did consider implementing `Default` like this, but it doesn't feel like a good fit? Like, the docs for `Default::default()` say "Returns the “default value” for a type.". IMO generating a random `KeyPair` is quite different from yielding "**the** “default value”.

I'm generally happy to tack something onto this branch but figured we should discuss the design first.

cpu:
> IMO generating a random KeyPair is quite different from yielding "the “default value”.

That's a fair point. I agree that it's probably not a good fit with that in mind.

> I'm generally happy to tack something onto this branch but figured we should discuss the design first.

`generate` for the default and `generate_for` for a specific alg seem OK to me :+1:

rcgen/src/lib.rs
cpu: > I didn't find the notion of "passed-through" very clear. 

Yeah, that (awkward) phrasing was added by me in #212 but looking at the diff again I think it was a mistake. The `CertificateParms::from_ca_cert_der` fn only returns `CertificateParams` and didn't pass through the keypair. I was probably thinking of the `Certificate::generate*` APIs that return both a `Certificate` and `KeyPair`, and for the case where you specify a `KeyPair` in the `CertificateParams` it is "passed through" to the caller in the return args (vs generating one from scratch). All of that falls away with this changeset so we can lose the awkward "pass through" notion entirely.

I think the right fix for this branch is to replace this text:
>  By providing the returned [`Certificate`] and the passed-through [`KeyPair`] to [`Certificate::generate`] or  [`Certificate::from_request`] functions you can generate a new certificate from [`CertificateParams`], signed by the pre-existing CA.

with something like:
> By providing the constructed [`CertificateParams`] and the [`KeyPair`] associated with your existing `ca_cert` you can use [`Certificate::generate`] or [`Certificate::from_request`] to issue new certificates using the CA cert.

> There doesn't seem to be a lot of context in https://github.com/rustls/rcgen/pull/12.

I'm also a bit skeptical of this API surface. I feel like there's some lurking danger in the way rcgen tries to reconstitute `CertificateParams` from the DER, only puling out a handful of parameters deemed needed for signing :-/

djc: I've adopted your language proposal. 👍

djc:
> > I'm generally happy to tack something onto this branch but figured we should discuss the design first.
> 
> `generate` for the default and `generate_for` for a specific alg seem OK to me 👍

Added some commits for this.

I also added a commit that removed the `alg` field from `CertificateRevocationList` in favor of relying on the `issuer.alg`.

djc:
Rebased on top of #215.

cpu (approved):

rcgen/src/key_pair.rs
```diff
@@ -56,6 +56,47 @@ pub struct KeyPair {
 }
 
 impl KeyPair {
+	/// Generate a new random PKCS_ECDSA_P256_SHA256 key pair
```
est31, line 59:
```suggestion
	/// Generate a new random [`PKCS_ECDSA_P256_SHA256`] key pair
```

rcgen/src/key_pair.rs
```diff
@@ -56,6 +56,47 @@ pub struct KeyPair {
...
+				let key_pair_serialized = key_pair_doc.as_ref().to_vec();
+
+				let key_pair = Ed25519KeyPair::from_pkcs8(&&key_pair_doc.as_ref()).unwrap();
+				Ok(KeyPair {
+					kind: KeyPairKind::Ed(key_pair),
+					alg,
+					serialized_der: key_pair_serialized,
+				})
+			},
+			// Ring doesn't have RSA key generation yet:
```
est31, line 93: Maybe also link https://github.com/aws/aws-lc-rs/issues/296 ?

rcgen/src/lib.rs
```diff
@@ -114,8 +110,9 @@ println!("{}", key_pair.serialize_pem());
 )]
 pub fn generate_simple_self_signed(
 	subject_alt_names: impl Into<Vec<String>>,
-) -> Result<CertifiedKey, Error> {
-	Certificate::generate_self_signed(CertificateParams::new(subject_alt_names))
+	key_pair: &KeyPair,
+) -> Result<Certificate, Error> {
```
est31, line 114: If we eliminate the `CertifiedKey` type, maybe it would be better to return a tuple here? Then most people would save one line (plus they won't have to import the `KeyPair` type).

est31 (approved):
Mostly good, mainly concerned about the `generate_simple_self_signed` function. Ideally we make its API as simple as possible, requiring as little parameters as possible.

rcgen/src/key_pair.rs
djc: Fixed this.

rcgen/src/key_pair.rs
djc: Note that this is a pre-existing issue. Added a commit that fixes it anyway.

djc:
In response to your feedback, I've changed the initial commit to retain `CertifiedKey` and kept the top-level `rcgen::generate_simple_self_signed()` returning `CertifiedKey` (that is, it generates a `KeyPair` for the caller and returns the `CertifiedKey` including the certificate and key pair). I've kept `Certificate::generate_self_signed()` taking the `key_pair` as an argument. I feel like this strikes a nice balance of a very approachable high-level API with more composable building blocks.

I'll leave this open for a bit but I'm assuming this is in line with the feedback I've received so far.

est31 (approved):
Great PR, really like it!

djc:
I noticed that the merge queue strategy is currently set to "Squash and merge". Is that an intentional change from before? Seems unfortunate to squash clean commit histories like this before merging.

cpu:
I would also prefer to rebase merge.

est31:
Yeah it was an intentional change (by me) because I felt it was hard to associate the commits on main with the PR they came from. I also don't like having to ask contributors to create a clean history, because not all know how to do that, and some simply don't care. The choices are either not having their contributions, fixing it yourself by pushing to the branch, or having the "bad" history in main (say with merge commits from upstream/main or stuff). The first and last option aren't really great, and the second is not good either (for the maintainers).

I see that the option to rebase merge at the time you put the PR into the merge queue is enabled again. I think that's good.

est31:
That is, I think we can merge PR with a clean history to main in rebase style, as long as the history is kept clean.

est31:
> I think the meta-point here is that I would have expected you to communicate this change prior to making it, since we're collaborating on maintenance of this crate. That you did this silently (and after we discussed that I and @cpu preferred rebasing in https://github.com/rustls/rcgen/issues/137) is disappointing.

I've expressed my preference in #137 for squashes, and was quite surprised that after the merge queue got enabled, only rebase merges were implemented. From my point of view that was silent as well, as in I only noticed that a while after the fact, without there having been explicit communication about it. So after I noticed that, I went and changed the setting back to squashes. In retrospect, I guess that I should have communicated that explicitly back when I did the change, which I think should be a lesson for all of us here.

I don't hold any grudges over this, I just wanted to give you my point of things so that you hopefully understand my point of view and don't develop any grudges either.

> looks like the merge queue is completely disabled now?

It wasn't me, I didn't touch any setting in the last 2+ weeks at least.

> In my experience (across many different projects) most contributors are open to this.

In my experience the contributors whose changes need input the most are those who are also the least able to do it. Someone who is doing their first steps with git, etc. I feel empathy with them and think the bar for contributions should be low so that people can grow. I suppose it depends on the type of project you have though, as some projects attract that group of people more than people who do something for their job (but that is no guarantee, many folks know git only from the GUI). With rcgen I do admit I have seen this more rarely.

Sometimes it's also due to different styles. Some people mainly work on projects that don't allow force pushes or history edits, while other projects explicitly want those.

djc:
> I've expressed my preference in #137 for squashes, and was quite surprised that after the merge queue got enabled, only rebase merges were implemented. From my point of view that was silent as well, as in I only noticed that a while after the fact, without there having been explicit communication about it. So after I noticed that, I went and changed the setting back to squashes. In retrospect, I guess that I should have communicated that explicitly back when I did the change, which I think should be a lesson for all of us here.
> 
> I don't hold any grudges over this, I just wanted to give you my point of things so that you hopefully understand my point of view and don't develop any grudges either.

That's fair, we should have made that explicit.

I'm fine with keeping the default setting to squash, and then we can just bypass the queue if/when we want to rebase?

@cpu any thoughts?

cpu:
> @cpu any thoughts?

In an ideal world I think GitHub would let you set a default strategy for the merge queue but override it at submit time. Unfortunately since that feature doesn't exist I think we're stuck evaluating workarounds. I think it's largely a question of what the default should be.

My intuition is that for this repository the type of contributor who might struggle with maintaining a clean commit history is less common than a more git-savvy contributor. A quick informal survey of the past ~9 PRs by external contributors shows that most are either 1 commit, or have a history I'd consider clean enough for a rebase.

What about flipping the proposal and making the default merge queue strategy rebase-merge but leaving the door open for administrative squash-merge that bypasses the queue when the PR merits it? 

I do think it's helpful to consider a variety of contributor styles and experience but I also don't like having to babysit CI tasks to wait to administratively merge work in the common case where it would be suitable for a rebase. The administrative merge route is a big hammer because if you don't wait for those CI tasks to complete you could easily merge code that fails in CI.

est31:
> What about flipping the proposal and making the default merge queue strategy rebase-merge but leaving the door open for administrative squash-merge

I've written about my preferences above, but reading your points of views and statements I'm more open to trying out rebase merges by default.

cpu:
> I'll change the merge queue setting to use rebase merge.

Done:

rebase

PR #214, opened 2024-01-17
Tighten up string type representations to prevent illegal values

Tudyx:
Fix https://github.com/rustls/rcgen/issues/181

Mainly inspired by the [der](https://docs.rs/der/0.7.8/der/asn1/index.html) crate.


djc:
This looks okay so far. One thought is that I think maybe the `TryFrom` impls should be for `String` instead of `&str`, so as to make it obvious an owned type is required (giving callers the opportunity to move a value rather than effectively having to clone it).

Tudyx:
Maybe we could provide both? I think the `TryFrom<&str>` is quite nice for users because it allows moving from (old API)

```rust
PrintableString("BR".into());
```
to

```rust
PrintableString("BR".try_into()?)
```

If we only provide `TryFrom<String>`, users would be obligated to do

```rust
PrintableString(String::from("BR").try_into()?)
```

Tudyx:
> giving callers the opportunity to move a value rather than effectively having to clone it)

Having both would also solve this one.

> so as to make it obvious an owned type is required

I see your point but I think it's okay because users are used to the standard `String`  type which has a similar API (`"foo".into()` or `String::from("foo")`)

rcgen/src/lib.rs
```diff
@@ -1465,18 +1467,23 @@ fn write_distinguished_name(writer: DERWriter, dn: &DistinguishedName) {
 					match content {
 						DnValue::BmpString(s) => writer
 							.next()
-							.write_tagged_implicit(TAG_BMPSTRING, |writer| writer.write_bytes(s)),
-						DnValue::Ia5String(s) => writer.next().write_ia5_string(s),
-						DnValue::PrintableString(s) => writer.next().write_printable_string(s),
+							.write_tagged_implicit(TAG_BMPSTRING, |writer| {
```
cpu, line 1470: I think this could be switched to use Yasna's [`write_bmp_string`](https://qnighy.github.io/yasna.rs/yasna/struct.DERWriter.html#method.write_bmp_string) method to avoid managing the tag ourselves and to match the IA5 String and PrintableString match arms.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,317 @@
+use crate::Error;
+use std::str::FromStr;
+
+/// ASN.1 `PrintableString` type.
+///
+/// Supports a subset the ASCII character set (described below).
```
cpu, line 6: small typo:

```suggestion
/// Supports a subset of the ASCII character set (described below).
```

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,317 @@
...
+		s.try_into()
+	}
+}
+
+/// ASN.1 `TeletexString` type.
+///
+/// # Supported characters
+///
+/// The standard defines a complex character set allowed in this type. However, quoting the ASN.1
+/// mailing list, "a sizable volume of software in the world treats TeletexString (T61String) as a
```
cpu, line 137: Do you have a reference for the ASN.1 mailing list post? Seems worth linking as a footnote.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,317 @@
...
+pub struct TeletexString(String);
+
+impl TeletexString {
+	/// Extracts a string slice containing the entire `TeletexString`.
+	pub fn as_str(&self) -> &str {
+		&self.0
+	}
+
+	/// Extracts a string slice containing the entire `PrintableString`.
+	pub fn as_bytes(&self) -> &[u8] {
```
cpu, line 154: I think this comment isn't accurate: it's not a string slice and it's not a `PrintableString` :-)

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,317 @@
...
+
+impl FromStr for BmpString {
+	type Err = Error;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		s.try_into()
+	}
+}
+
+/// ASN.1 `BMPString` type.
```
cpu, line 256: I think this comment headline mentions the wrong type:

```suggestion
/// ASN.1 `UniversalString` type.
```

cpu (commented):
Here's a quick review pass. So far it's looking good. Thanks for picking this up

rcgen/src/string_types.rs
Tudyx: I added it it the new commit ;)

rcgen/src/lib.rs
Tudyx: As the inner type is a `Vec<u8>` encoded as UTF-16 BE it will force us to allocate a new string to be able to call this. 

One solution is to change the inner type to a `String`  but if one day we move from `yasna` we don' t have the guarantee that the new one will provide a helper to handle the UTF-8 to UTF-16 conversion.

Besides there is also no `write_universal_string` so for the `UniversalString` type, it makes no sense to have a `String` as an inner type for this one. So I would prefer keeping a `Vec<u8>` representation for both ` BmpString` and `UniversalString`

Tudyx:
I've updated the PR with the missing methods, and integrate it with subject alternative names.
Some notes: 
- All the types can be created with an UTF-8 encoded string (` &str`/ `String` ). For instance, `BmpString::try_from("❤πü2?")?` . A check is performed to see if all input's characters are in the valid charset.
- Instead of providing ` TryFrom<&[u8]>`  I prefer to be explicit and use  ` from_<encoding>` like [`String::from_utf8`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.from_utf8)

The PR is still a draft because I would like to add more tests but I haven't found a good way yet.

rcgen/src/error.rs
```diff
@@ -13,6 +13,16 @@ pub enum Error {
 	#[cfg(feature = "x509-parser")]
 	/// Invalid subject alternative name type
 	InvalidNameType,
+	/// Invalid PrintableString type
+	InvalidPrintableString,
```
est31, line 17: I think this should be just one variant, optionally with an enum to distinguish the string type.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,490 @@
...
+/// # Examples
+///
+/// You can create a `PrintableString` from [a literal string][`&str`] with [`PrintableString::try_from`]:
+///
+/// ```
+/// use rcgen::PrintableString;
+/// let hello = PrintableString::try_from("hello").unwrap();
+/// ```
+///
+/// # Supported characters
```
est31, line 20: Maybe it would make sense to link to the [PrintableString](https://en.wikipedia.org/wiki/PrintableString) wikipedia page here and also say that this is a true subset of the ascii printable characters? Say @ is a printable charater as per ASCII but can't be part of ASN.1's `PrintableString`.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,490 @@
...
+///
+/// ```
+/// use rcgen::Ia5String;
+/// let hello = Ia5String::try_from("hello").unwrap();
+/// ```
+///
+/// # Supported characters
+///
+/// Supports the [International Alphabet No. 5 (IA5)] character encoding, i.e.
+/// the lower 128 characters of the ASCII alphabet. (Note: IA5 is now
```
est31, line 125: ASCII only has 128 characters, so not sure what you mean by lower 128 characters? There is multiple extensions on top of ASCII, like UTF-8 or [ISO/IEC 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1).

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,490 @@
...
+/// technically known as the International Reference Alphabet or IRA as
+/// specified in the ITU-T's T.50 recommendation).
+///
+/// For UTF-8, use [`String`][`std::string::String`].
+///
+/// [International Alphabet No. 5 (IA5)]: https://en.wikipedia.org/wiki/T.50_%28standard%2
```
est31, line 131: Does this link work? I tried to copy it, it doesn't work for me.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,490 @@
...
+
+	/// Converts a [`String`][`std::string::String`] into a [`TeletexString`]
+	///
+	/// Any character not in the [`TeletexString`] charset will be rejected.
+	/// See [`TeletexString`] documentation for more information.
+	///
+	/// This conversion does not allocate or copy memory.
+	fn try_from(input: String) -> Result<Self, Error> {
+		// Check all bytes are visible
+		if !input.as_bytes().iter().all(|b| 0x20 <= *b && *b <= 0x7f) {
```
est31, line 250:
```suggestion
		if !input.as_bytes().iter().all(|b| (0x20..=0x7f).contains(b)) {
```

est31 (changes requested):

rcgen/src/error.rs
Tudyx: I've made it into one variant with an enum to distinguish the string type. I've also added some context to the error

rcgen/src/string_types.rs
Tudyx: Yep my bad, I fix the comment

rcgen/src/string_types.rs
Tudyx: I fix the link ;)

rcgen/src/string_types.rs
Tudyx: That's much clearer, thank you! I've updated the doc

rcgen/src/error.rs
```diff
@@ -98,6 +101,36 @@ impl fmt::Display for Error {
...
+	Ia5String(String),
+	/// Invalid TeletexString type
+	TeletexString(String),
+	/// Invalid BmpString type
+	BmpString(String),
+}
+
+impl fmt::Display for InvalidAsn1String {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		use InvalidAsn1String as E;
```
djc, line 122: Let's just do `use InvalidAsn1String::*;` here.

rcgen/src/string_types.rs
```diff
@@ -0,0 +1,665 @@
...
+
+impl TryFrom<String> for Ia5String {
+	type Error = Error;
+
+	/// Converts a [`String`][`std::string::String`] into a [`Ia5String`]
+	///
+	/// Any character not in the [`Ia5String`] charset will be rejected.
+	/// See [`Ia5String`] documentation for more information.
+	///
+	/// This conversion does not allocate or copy memory.
```
djc, line 204: I feel like we should just leave this out.

rcgen/src/error.rs
Tudyx: Using the variants name with a wildcard can lead to difficult diag bug. If we remove a variant of the `enum`, then in the match the old variant become a variable that will be a catch-all pattern. The warning that will be emitted will be  a "non snake case name" that's difficult to diag.
This [video](https://youtu.be/8j_FbjiowvE?feature=shared&t=97) explains this better than I do.

I wanted to open a PR after this one to change the code when we use this pattern.

Tudyx: The only lint I found is [one which prevent this pattern for stylistic purpose](https://rust-lang.github.io/rust-clippy/master/index.html#/enum_glob_use). Please do you have a link?

I don't feel strongly about it and I can use the variant with a wildcard as the rest of the crate does. If we want to tackle this one day, maybe it's better to do all this in one separate PR

est31 (approved):

Tudyx:
I rebased on top of https://github.com/rustls/rcgen/pull/213. I've also cleanup the git history a bit

rustls-cert-gen/src/main.rs
```diff
@@ -82,15 +82,12 @@ struct Options {
...
+fn parse_sans(hosts: Vec<String>) -> Result<Vec<SanType>, Error> {
 	hosts
 		.into_iter()
-		.map(|host| {
-			if let Ok(ip) = host.parse::<IpAddr>() {
-				SanType::IpAddress(ip)
-			} else {
-				SanType::DnsName(host)
-			}
+		.map(|s| match s.parse() {
```
djc, line 88: Style nit: I'd write this as

```rust
Ok(match IpAddr::from_str(ip) {
    Ok(ip) => SanType::IpAddress(ip),
    Err(_) => SanType::DnsName(s.try_into()?),
})
```

djc (approved):

rustls-cert-gen/src/main.rs
Tudyx: I've changed this in the latest push ;)

rcgen/src/lib.rs
cpu: > As the inner type is a Vec<u8> encoded as UTF-16 BE it will force us to allocate a new string to be able to call this.

Good point :+1: 

> So I would prefer keeping a Vec<u8> representation for both  BmpString and UniversalString

That sounds reasonable to me. Thanks for considering the feedback.

rcgen/src/error.rs
```diff
@@ -90,6 +93,36 @@ impl fmt::Display for Error {
 
 impl std::error::Error for Error {}
 
+/// Invalid ASN.1 string type
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum InvalidAsn1String {
+	/// Invalid PrintableString type
+	PrintableString(String),
+	/// Invalide UniversalString type
```
cpu, line 102: small typo here:

```suggestion
	/// Invalid UniversalString type
```

cpu (approved):
Looks good to me. Thank you

cpu:
@Tudyx If you rebase I think this is ready for merge. Thank you!

cpu:
Thanks again. I appreciate you wrangling this tech debt to unblock the feature work you were originally interested in doing. You went the extra mile :trophy:

PR #215, opened 2024-01-18
Upgrade webpki dev-dependency to 0.102

djc:


rcgen/Cargo.toml
```diff
@@ -47,8 +47,9 @@ allowed_external_types = [
 
 [dev-dependencies]
 openssl = "0.10"
+pki-types = { package = "rustls-pki-types", version = "1" }
```
cpu, line 50: Is there a reason you wanted to avoid using the `rustls-webpki::types` re-export?

rcgen/tests/webpki.rs
```diff
@@ -590,17 +593,9 @@ fn test_webpki_crl_parse() {
...
-
-	// Webpki represents certificate SPKIs internally without the outer SEQUENCE.
-	// We remove that here before calling verify_signature.
-	let issuer_spki = issuer_key.public_key_der();
-	let raw_spki = yasna::parse_der(&issuer_spki, |reader| reader.read_tagged_der()).unwrap();
-
-	// We should be able to verify the CRL signature with the issuer's raw SPKI.
-	webpki_crl
-		.verify_signature(&[&webpki::ECDSA_P256_SHA256], raw_spki.value())
-		.expect("failed to validate CRL signature");
```
cpu, line 603: just noting for other reviewers that this hunk of test code falls away because we removed the `verify_signature` fn of the CRL type upstream. It didn't mesh well with the existing API surface, was difficult to use (as evidenced by the SPKI SEQUENCE dance), and had no other consumers we were aware of except this unit test.

The `test_webpki_crl_revoke` test that remains uses the CRL for a revocation checking decision and the webpki crate will verify the signature as part of that work.

cpu (approved):
LGTM, thank you.

rcgen/Cargo.toml
djc: I sort of forgot it existed, but now that you brought it up, I think we decided when merging the rustls re-export that the rustls-webpki was maybe not a great fit and that we should remove it in a future release?

cpu: Ah yes, I remember that. We also talked about how rustls re-exports it as "pki-types" and webpki chose "types".

With that context in mind I think not using the webpki re-export makes sense :+1:

est31 (approved):

PR #223, opened 2024-02-11
Use pki_types to improve the interoperability with the rustls ecosystem 

Tudyx:
Use [`pki_types`] to improve the interoperability with the `rustls` ecosystem and make types pretty clear, as suggested by https://github.com/rustls/rcgen/pull/170#issuecomment-1849565550. 

To fully get rid of the `pem` dependency (by replacing it by `rustls_pemfile`), I think those further actions will be needed:
- add `CertificateSigningRequestDer` to  [`pki_types`]
- add a `Csr` variant to [rustls_pemfile::Item] which is a newtype of ` CertificateSigningRequestDer` 
- add some encoding capability to [rustls_pemfile](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/index.html#), basically it would mean adding headers ("---- BEGIN ----") and handle well line ending.

Please let me know if I miss something and if you think this goes in the right direction.
I'm open to submit following PR for this.

[rustls_pemfile::Item]: https://docs.rs/rustls-pemfile/latest/rustls_pemfile/enum.Item.html
[`CryptoProvider`]: https://docs.rs/rustls/0.22.2/rustls/crypto/struct.CryptoProvider.html
[`pki_types`]: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/index.html#

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 
 #[cfg(feature = "pem")]
 use pem::Pem;
+use pki_types::{CertificateDer, CertificateSigningRequestDer};
 use time::{Date, Month, OffsetDateTime, PrimitiveDateTime, Time};
 use yasna::models::ObjectIdentifier;
 use yasna::{DERWriter, Tag};
@@ -23,7 +24,7 @@ use crate::{
 pub struct Certificate {
 	pub(crate) params: CertificateParams,
 	pub(crate) subject_public_key_info: Vec<u8>,
-	pub(crate) der: Vec<u8>,
+	pub(crate) der: CertificateDer<'static>,
 }
 
 impl Certificate {
@@ -39,13 +40,16 @@ impl Certificate {
 			.derive(&self.subject_public_key_info)
 	}
 	/// Get the certificate in DER encoded format.
-	pub fn der(&self) -> &[u8] {
+	///
+	/// As the return type implements `Deref<Target = [u8]>`, in can easily be saved
+	/// to a file like a byte slice.
+	pub fn der(&self) -> &CertificateDer<'static> {
 		&self.der
 	}
 	/// Get the certificate in PEM encoded format.
 	#[cfg(feature = "pem")]
 	pub fn pem(&self) -> String {
-		pem::encode_config(&Pem::new("CERTIFICATE", self.der()), ENCODE_CONFIG)
+		pem::encode_config(&Pem::new("CERTIFICATE", self.der().to_vec()), ENCODE_CONFIG)
 	}
 	/// Generate and serialize a certificate signing request (CSR) in binary DER format.
 	///
@@ -57,7 +61,13 @@ impl Certificate {
 	/// should not call `serialize_request_der` and then `serialize_request_pem`. This will
 	/// result in two different CSRs. Instead call only `serialize_request_pem` and base64
 	/// decode the inner content to get the DER encoded CSR using a library like `pem`.
-	pub fn serialize_request_der(&self, subject_key: &KeyPair) -> Result<Vec<u8>, Error> {
+	///
+	/// As the return type implements `Deref<Target = [u8]>`, in can easily be saved
+	/// to a file like a byte slice.
+	pub fn serialize_request_der(
+		&self,
+		subject_key: &KeyPair,
+	) -> Result<CertificateSigningRequestDer<'static>, Error> {
 		yasna::try_construct_der(|writer| {
 			writer.write_sequence(|writer| {
 				let cert_data = yasna::try_construct_der(|writer| {
@@ -74,6 +84,7 @@ impl Certificate {
 				Ok(())
 			})
 		})
+		.map(CertificateSigningRequestDer::from)
 	}
 	/// Generate and serialize a certificate signing request (CSR) in binary DER format.
 	///
@@ -88,7 +99,7 @@ impl Certificate {
 	#[cfg(feature = "pem")]
 	pub fn serialize_request_pem(&self, subject_key: &KeyPair) -> Result<String, Error> {
 		let contents = self.serialize_request_der(subject_key)?;
-		let p = Pem::new("CERTIFICATE REQUEST", contents);
+		let p = Pem::new("CERTIFICATE REQUEST", contents.to_vec());
 		Ok(pem::encode_config(&p, ENCODE_CONFIG))
 	}
 }
@@ -218,7 +229,7 @@ impl CertificateParams {
 	#[cfg(all(feature = "pem", feature = "x509-parser"))]
 	pub fn from_ca_cert_pem(pem_str: &str) -> Result<Self, Error> {
 		let certificate = pem::parse(pem_str).or(Err(Error::CouldNotParseCertificate))?;
-		Self::from_ca_cert_der(certificate.contents())
+		Self::from_ca_cert_der(&certificate.contents().into())
 	}
 
 	/// Parses an existing ca certificate from the DER format.
@@ -236,8 +247,14 @@ impl CertificateParams {
 	/// This function assumes the provided certificate is a CA. It will not check
 	/// for the presence of the `BasicConstraints` extension, or perform any other
 	/// validation.
+	///
+	/// You can use [`rustls_pemfile::certs`] to get the `ca_cert` input. If
+	/// you have already a byte slice, just calling `into()` and taking a reference
+	/// will convert it to [`CertificateDer`].
+	///
+	/// [`rustls_pemfile::certs`]: https://docs.rs/rustls-pemfile/latest/rustls_pemfile/fn.certs.html
 	#[cfg(feature = "x509-parser")]
-	pub fn from_ca_cert_der(ca_cert: &[u8]) -> Result<Self, Error> {
+	pub fn from_ca_cert_der(ca_cert: &CertificateDer<'_>) -> Result<Self, Error> {
 		let (_remainder, x509) = x509_parser::parse_x509_certificate(ca_cert)
 			.or(Err(Error::CouldNotParseCertificate))?;
 
@@ -802,12 +819,17 @@ impl CertificateParams {
 			Ok(())
 		})
 	}
+
+	/// Generate and serialize a certificate signed by the given issuer.
+	///
+	/// As the return type implements `Deref<Target = [u8]>`, in can easily be saved
+	/// to a file like a byte slice.
 	pub(crate) fn serialize_der_with_signer<K: PublicKeyData>(
 		&self,
 		pub_key: &K,
 		issuer: &KeyPair,
 		issuer_name: &DistinguishedName,
-	) -> Result<Vec<u8>, Error> {
+	) -> Result<CertificateDer<'static>, Error> {
 		yasna::try_construct_der(|writer| {
 			writer.write_sequence(|writer| {
 				let tbs_cert_list_serialized = yasna::try_construct_der(|writer| {
@@ -826,6 +848,7 @@ impl CertificateParams {
 				Ok(())
 			})
 		})
+		.map(CertificateDer::from)
 	}
 }
 
--- a/rcgen/src/csr.rs
+++ b/rcgen/src/csr.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "x509-parser")]
+use pki_types::CertificateSigningRequestDer;
+
 use std::hash::Hash;
 
 use crate::{Certificate, CertificateParams, Error, KeyPair, PublicKeyData, SignatureAlgorithm};
@@ -36,15 +39,21 @@ impl CertificateSigningRequestParams {
 	#[cfg(all(feature = "pem", feature = "x509-parser"))]
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let csr = pem::parse(pem_str).or(Err(Error::CouldNotParseCertificationRequest))?;
-		Self::from_der(csr.contents())
+		Self::from_der(&csr.contents().into())
 	}
 
 	/// Parse a certificate signing request from DER-encoded bytes
 	///
 	/// Currently, this only supports the `Subject Alternative Name` extension.
 	/// On encountering other extensions, this function will return an error.
+	///
+	/// You can use [`rustls_pemfile::csr`] to get the `csr` input. If
+	/// you have already a byte slice, just calling `into()` and taking a reference
+	/// will convert it to [`CertificateSigningRequestDer`].
+	///
+	/// [`rustls_pemfile::csr`]: https://docs.rs/rustls-pemfile/latest/rustls_pemfile/fn.csr.html
 	#[cfg(feature = "x509-parser")]
-	pub fn from_der(csr: &[u8]) -> Result<Self, Error> {
+	pub fn from_der(csr: &CertificateSigningRequestDer<'_>) -> Result<Self, Error> {
 		use x509_parser::prelude::FromDer;
 		let csr = x509_parser::certification_request::X509CertificationRequest::from_der(csr)
 			.map_err(|_| Error::CouldNotParseCertificationRequest)?
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -1,5 +1,7 @@
 #[cfg(feature = "pem")]
 use pem::Pem;
+#[cfg(feature = "crypto")]
+use pki_types::PrivatePkcs8KeyDer;
 use std::fmt;
 use yasna::DERWriter;
 
@@ -158,8 +160,14 @@ impl KeyPair {
 	/// Parses the key pair from the DER format
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
+	///
+	/// You can use [`rustls_pemfile::private_key`] to get the `der` input. If
+	/// you have already a byte slice, just calling `into()` and taking a reference
+	/// will convert it to a [`PrivatePkcs8KeyDer`].
+	///
+	/// [`rustls_pemfile::private_key`]: https://docs.rs/rustls-pemfile/latest/rustls_pemfile/fn.private_key.html
 	#[cfg(feature = "crypto")]
-	pub fn from_der(der: &[u8]) -> Result<Self, Error> {
+	pub fn from_der(der: &PrivatePkcs8KeyDer<'_>) -> Result<Self, Error> {
 		der.try_into()
 	}
 
@@ -169,11 +177,15 @@ impl KeyPair {
 	}
 
 	/// Parses the key pair from the ASCII PEM format
+	///
+	/// The key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
+	///
+	/// Appears as "PRIVATE KEY" in PEM files.
 	#[cfg(all(feature = "pem", feature = "crypto"))]
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)._err()?;
 		let private_key_der: &[_] = private_key.contents();
-		private_key_der.try_into()
+		Self::from_der(&private_key_der.into())
 	}
 
 	/// Obtains the key pair from a raw public key and a remote private key
@@ -188,6 +200,9 @@ impl KeyPair {
 	/// Obtains the key pair from a DER formatted key
 	/// using the specified [`SignatureAlgorithm`]
 	///
+	/// The key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
+	///
+	/// Appears as "PRIVATE KEY" in PEM files
 	/// Same as [from_pem_and_sign_algo](Self::from_pem_and_sign_algo).
 	#[cfg(all(feature = "pem", feature = "crypto"))]
 	pub fn from_pem_and_sign_algo(
@@ -196,7 +211,7 @@ impl KeyPair {
 	) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)._err()?;
 		let private_key_der: &[_] = private_key.contents();
-		Self::from_der_and_sign_algo(private_key_der, alg)
+		Self::from_der_and_sign_algo(&PrivatePkcs8KeyDer::from(private_key_der), alg)
 	}
 
 	/// Obtains the key pair from a DER formatted key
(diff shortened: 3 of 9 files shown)


djc:
Thanks for working on this! I would propose doing a smaller increment which preserves the use of the pem crate for parsing (but still uses the pki-types for DER-containing types in the public API).

What do you think?

Tudyx:
Yes sure, let's do this step by step!
I've kept the dependency to the `pem` crate, and submit a PR to `pki_types`  to add the `CertificateSigningRequestDer` type (https://github.com/rustls/pki-types/pull/33)

cpu:
> ci / Validate external types appearing in public API (pull_request) Failing after 1m

To fix this task you'll need to update the cargo-check-external-types metadata to add `rustls_pki_types` to the allow-list:
https://github.com/rustls/rcgen/blob/747b5d8e72d1caadedff415445487320d899570a/rcgen/Cargo.toml#L47-L51

The intent of `pki-types` is to be a stable API and so it's also ignored for this task in the other various rustls repos.

Tudyx:
> To fix this task you'll need to update the cargo-check-external-types metadata to add rustls_pki_types to the allow-list:
I see, thank you for the tip!

rcgen/tests/botan.rs
```diff
@@ -17,12 +18,12 @@ fn default_params() -> (CertificateParams, KeyPair) {
 	(params, key_pair)
 }
 
-fn check_cert(cert_der: &[u8], cert: &Certificate) {
+fn check_cert(cert_der: &CertificateDer, cert: &Certificate) {
```
cpu, line 21: If you decide it's worth having the explicit lifetimes here and L26 could use the same treatment.

rcgen/tests/webpki.rs
```diff
@@ -54,7 +54,7 @@ fn sign_msg_rsa(key_pair: &KeyPair, msg: &[u8], encoding: &'static dyn RsaEncodi
 }
 
 fn check_cert<'a, 'b>(
-	cert_der: &[u8],
+	cert_der: &CertificateDer,
```
cpu, line 57: ditto lifetime feedback here and L71, L73

cpu (approved):
Thanks for picking this up. It'll be nice to get this in with the other breaking changes in `main` before a release :+1: 

I flagged some nits but the overall meat of the work seems great as-is.

cpu (approved):

cpu:
We're going to cut a pki-types release with the CSR type: https://github.com/rustls/pki-types/pull/34 If you want to take a Cargo patch and switch this branch over we can probably have that release published before this PR is done with review saving a follow-up for that.

rcgen/src/lib.rs
```diff
@@ -682,8 +683,8 @@ impl CertificateParams {
 	/// for the presence of the `BasicConstraints` extension, or perform any other
 	/// validation.
 	#[cfg(feature = "x509-parser")]
-	pub fn from_ca_cert_der(ca_cert: &[u8]) -> Result<Self, Error> {
-		let (_remainder, x509) = x509_parser::parse_x509_certificate(ca_cert)
+	pub fn from_ca_cert_der(ca_cert: &CertificateDer) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
Tudyx: Thank you, I also think it's better this way.
 
Maybe we could activate the [`elided-lifetimes-in-paths`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#elided-lifetimes-in-paths) `rustc`  lint in another PR to enforce the `rustls` convention also on this crate?

Tudyx:
I've used a patch while waiting for the release :+1:  

I wonder if `CeritificateSigningRequest::from_der`  should take ownership or a reference to `CeritificateSigningRequestDer`

rcgen/src/key_pair.rs
```diff
@@ -130,7 +132,7 @@ impl KeyPair {
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)._err()?;
 		let private_key_der: &[_] = private_key.contents();
-		Ok(private_key_der.try_into()?)
+		Self::from_der(PrivatePkcs8KeyDer::from(private_key_der))
```
djc, line 135: Maybe the docstring for this function should document that this function can only consume PKCS#8 keys?

rcgen/src/key_pair.rs
```diff
@@ -166,38 +171,42 @@ impl KeyPair {
...
 	#[cfg(feature = "crypto")]
-	pub fn from_der_and_sign_algo(
-		pkcs8: &[u8],
+	pub fn from_der_and_sign_algo<'a>(
+		der: impl Into<PrivateKeyDer<'a>>,
 		alg: &'static SignatureAlgorithm,
 	) -> Result<Self, Error> {
 		let rng = &SystemRandom::new();
-		let pkcs8_vec = pkcs8.to_vec();
+		let serialized_der = match &der.into() {
```
djc, line 179: IMO we should take `impl Into<PrivatePkcs8Key<'_>>` instead to make it clear at compile time we only support PKCS#8 keys.

rcgen/src/key_pair.rs
```diff
@@ -352,29 +362,18 @@ impl KeyPair {
...
-			kind,
-			alg,
-			serialized_der: pkcs8.to_vec(),
-		})
-	}
-}
-
-#[cfg(feature = "crypto")]
-impl TryFrom<Vec<u8>> for KeyPair {
+impl TryFrom<PrivateKeyDer<'_>> for KeyPair {
```
djc, line 365: PKCS#8?

rcgen/src/key_pair.rs
```diff
@@ -116,8 +118,8 @@ impl KeyPair {
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
 	#[cfg(feature = "crypto")]
-	pub fn from_der(der: &[u8]) -> Result<Self, Error> {
-		Ok(der.try_into()?)
+	pub fn from_der<'a>(der: impl Into<PrivatePkcs8KeyDer<'a>>) -> Result<Self, Error> {
```
djc, line 121: Hmm, maybe using `impl Into<>` here also kind of defeats the point?

rcgen/src/key_pair.rs
```diff
@@ -166,38 +174,37 @@ impl KeyPair {
 	/// same der key. In that instance, you can use this function to precisely
 	/// specify the `SignatureAlgorithm`.
 	#[cfg(feature = "crypto")]
-	pub fn from_der_and_sign_algo(
-		pkcs8: &[u8],
+	pub fn from_der_and_sign_algo<'a>(
+		der: impl Into<PrivatePkcs8KeyDer<'a>>,
```
djc, line 178: And same here.

rcgen/src/key_pair.rs
```diff
@@ -126,11 +128,13 @@ impl KeyPair {
 	}
 
 	/// Parses the key pair from the ASCII PEM format
+	///
+	/// The key must be in the PKCS#8 format.
```
djc, line 132: For these comments, maybe we should mention the `PRIVATE KEY` string used for the PEM delimiters, similar to what we have in rustls-pemfile?

djc (approved):

cpu:
@est31 Do you think you'll have bandwidth to review this soon? Should we continue to hold merging?

est31:
It needs a rebase, right?

rcgen/src/key_pair.rs
```diff
@@ -126,11 +128,15 @@ impl KeyPair {
...
+	///
+	/// The key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
+	///
+	/// Appears as "PRIVATE KEY" in PEM files
 	#[cfg(all(feature = "pem", feature = "crypto"))]
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)._err()?;
 		let private_key_der: &[_] = private_key.contents();
-		Ok(private_key_der.try_into()?)
+		Self::from_der(&PrivatePkcs8KeyDer::from(private_key_der))
```
est31, line 139: An into would also work here, right?

rcgen/src/key_pair.rs
```diff
@@ -116,8 +118,8 @@ impl KeyPair {
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
 	#[cfg(feature = "crypto")]
-	pub fn from_der(der: &[u8]) -> Result<Self, Error> {
-		Ok(der.try_into()?)
+	pub fn from_der(der: &PrivatePkcs8KeyDer<'_>) -> Result<Self, Error> {
```
est31, line 121: Could you also mention in the docs that one can pass a byte slice and use `into`, maybe with an example?

rcgen/src/csr.rs
```diff
@@ -36,17 +38,17 @@ impl CertificateSigningRequestParams {
...
+		Self::from_der(&csr.contents().into())
 	}
 
 	/// Parse a certificate signing request from DER-encoded bytes
 	///
 	/// Currently, this only supports the `Subject Alternative Name` extension.
 	/// On encountering other extensions, this function will return an error.
 	#[cfg(feature = "x509-parser")]
-	pub fn from_der(csr: &[u8]) -> Result<Self, Error> {
+	pub fn from_der(csr: &CertificateSigningRequestDer<'_>) -> Result<Self, Error> {
```
est31, line 49: Could you also mention in the docs that one can pass a byte slice and use `into` maybe with an example?

rcgen/src/key_pair.rs
```diff
@@ -167,37 +179,36 @@ impl KeyPair {
 	/// specify the `SignatureAlgorithm`.
 	#[cfg(feature = "crypto")]
 	pub fn from_der_and_sign_algo(
-		pkcs8: &[u8],
+		der: &PrivatePkcs8KeyDer<'_>,
```
est31, line 182: Could you also mention in the docs that one can pass a byte slice and use `into` maybe with an example?

rcgen/src/key_pair.rs
```diff
@@ -206,14 +217,15 @@ impl KeyPair {
...
 			alg,
-			serialized_der: pkcs8_vec,
+			serialized_der,
 		})
 	}
 
 	#[cfg(feature = "crypto")]
 	pub(crate) fn from_raw(
-		pkcs8: &[u8],
+		pkcs8: &PrivatePkcs8KeyDer,
```
est31, line 226: Could you also mention in the docs that one can pass a byte slice and use `into` maybe with an example?

rcgen/src/key_pair.rs
```diff
@@ -352,29 +364,29 @@ impl KeyPair {
 }
 
 #[cfg(feature = "crypto")]
-impl TryFrom<&[u8]> for KeyPair {
+impl TryFrom<&PrivatePkcs8KeyDer<'_>> for KeyPair {
```
est31, line 367: Why did you remove the old slice conversion functionality? It is easier to go through slices.

rcgen/src/key_pair.rs
```diff
@@ -352,29 +364,29 @@ impl KeyPair {
...
 			alg,
-			serialized_der: pkcs8.to_vec(),
+			serialized_der: private_key.secret_pkcs8_der().into(),
 		})
 	}
 }
 
 #[cfg(feature = "crypto")]
-impl TryFrom<Vec<u8>> for KeyPair {
+impl TryFrom<PrivatePkcs8KeyDer<'_>> for KeyPair {
```
est31, line 381: Why did you remove the old Vec conversion functionality? It is easier to go through Vecs.

rcgen/src/lib.rs
```diff
@@ -682,8 +683,8 @@ impl CertificateParams {
 	/// for the presence of the `BasicConstraints` extension, or perform any other
 	/// validation.
 	#[cfg(feature = "x509-parser")]
-	pub fn from_ca_cert_der(ca_cert: &[u8]) -> Result<Self, Error> {
-		let (_remainder, x509) = x509_parser::parse_x509_certificate(ca_cert)
+	pub fn from_ca_cert_der(ca_cert: &CertificateDer<'_>) -> Result<Self, Error> {
```
est31, line 686: Could you also mention in the docs that one can pass a byte slice and use `into` maybe with an example?

rcgen/src/lib.rs
```diff
@@ -1252,7 +1253,7 @@ impl CertificateParams {
 		pub_key: &K,
 		issuer: &KeyPair,
 		issuer_name: &DistinguishedName,
-	) -> Result<Vec<u8>, Error> {
+	) -> Result<CertificateDer<'static>, Error> {
```
est31, line 1256: Could you also mention in the docs how this can be converted into a vec or byte slice so that it can be saved into a file, ideally with an example that we test or rustdoc links so that we notice the breakage.

rcgen/src/lib.rs
```diff
@@ -1692,13 +1694,13 @@ impl Certificate {
 			.derive(&self.subject_public_key_info)
 	}
 	/// Get the certificate in DER encoded format.
-	pub fn der(&self) -> &[u8] {
+	pub fn der(&self) -> &CertificateDer<'static> {
```
est31, line 1697: Could you also mention in the docs how this can be converted into a vec or byte slice so that it can be saved into a file, ideally with an example that we test or rustdoc links so that we notice the breakage.

rcgen/src/lib.rs
```diff
@@ -1710,7 +1712,10 @@ impl Certificate {
 	/// should not call `serialize_request_der` and then `serialize_request_pem`. This will
 	/// result in two different CSRs. Instead call only `serialize_request_pem` and base64
 	/// decode the inner content to get the DER encoded CSR using a library like `pem`.
-	pub fn serialize_request_der(&self, subject_key: &KeyPair) -> Result<Vec<u8>, Error> {
+	pub fn serialize_request_der(
+		&self,
+		subject_key: &KeyPair,
+	) -> Result<CertificateSigningRequestDer<'static>, Error> {
```
est31, line 1718: Could you also mention in the docs how this can be converted into a vec or byte slice so that it can be saved into a file, ideally with an example that we test or rustdoc links so that we notice the breakage.

est31 (changes requested):
I've always felt down when using these special types with crates like ring for example (the rust crypto ecosystem also uses them heavily), because you have to import an additional type to use the API, and maybe even an additional crate. Actually you don't but you need to first look up if there is a from impl, etc. You also need to figure out how to convert it into bytes again in the output. This makes interaction with such APIs way more expensive for developers.

At the least, we should make this interaction very clear in the docs, making sure that users know how to use rcgen without using the interop crate. I've pointed out all the locations in the comments.

It might indeed improve interopability with other libraries/crates that use these types, but I think that most users just want to convert from byte slices and back, because they get something from a network, or a file, or the other way round. So while improving the interopability with a minority of use cases, it hurts the interopability with the majority (or at least I believe). Vec and slices have conversions to the `Bytes` type for example, but I couldn't find anything about `Bytes` in the docs of `PrivatePkcs8KeyDer`. So you first have to convert the `Bytes` to a Vec or a slice and then to a `PrivatePkcs8KeyDer`.

The main advantage I see in DER specific types would be that it could take care of conversion to PEM, you'd just call a function `.pem()` on it. But when I look at the rustdocs of [PrivatePkcs8KeyDer](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.PrivatePkcs8KeyDer.html#) I don't see any explanation how to convert it to pem, nor is there an easy `pem()` that returns a string (or another wrapper).

Such types do have advantages as they carry more meaning than the byte slice type, and I'm definitely not advocating for adopting C where everything is a `void*` but one should be extremely careful to not stand in the way of users, and stuff like I/O is usually just byte slices.

rcgen/src/key_pair.rs
Tudyx: Basically for the same reasons mentioned in https://github.com/rustls/rcgen/pull/170
To my knowledge, they are at least 3 standard ways to store a key. PKCS#8 (RFC 5958),  PKCS#1 (RFC 3447) and Sec1 (RFC 5915). All the 3 can be encoded as BER or DER, so they can be represented by a `Vec<u8>/&[u8]`. 
By using a specific type, we make clear to the user what format is expected, which it could be easily missed if it was only written in the doc.

Another solution is to specify the encoding in the function name and take a `Vec<u8>`, like for the `String::from_utf16be` or `String::from_utf8`. We would have the same clarity but without the interoperability with the rustls ecosystem. That's why I think the suggested solution is better.

djc:
> I've always felt down when using these special types with crates like ring for example (the rust crypto ecosystem also uses them heavily), because you have to import an additional type to use the API, and maybe even an additional crate. Actually you don't but you need to first look up if there is a from impl, etc. You also need to figure out how to convert it into bytes again in the output. This makes interaction with such APIs way more expensive for developers.

I do sympathize with these concerns, however in this case:

* I think it substantially helps users of different experience levels to be clear about PEM vs DER. Although rcgen already does a great job of making this clear in the function names, I think the type-level distinctions can help.
* It is very easy to convert a `&[u8]` or `Vec<u8>` to one of these new types.

(I myself recently got confused and passed DER into an API that expected a `Vec<u8>` containing PEM. This was an SDK wrapping tonic, which does clearly mark the underlying fields as PEM.)

> The main advantage I see in DER specific types would be that it could take care of conversion to PEM, you'd just call a function `.pem()` on it. But when I look at the rustdocs of [PrivatePkcs8KeyDer](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.PrivatePkcs8KeyDer.html#) I don't see any explanation how to convert it to pem, nor is there an easy `pem()` that returns a string (or another wrapper).

Yes, there is no PEM encoding code in the rustls project. I think we might want to change that going forward, and it would definitely make the case for this stronger.

rcgen/src/key_pair.rs
est31: Ideally for the `&[u8]/Vec<u8>` path, we'd be sniffing the format and then convert it from there.

>  We would have the same clarity but without the interoperability with the rustls ecosystem.

I'm not arguing against existence of `From<PrivatePkcs8KeyDer<'_>>` here, that one can be kept. It's not an either or. Types can have multiple `From<...>` impls (that's the difference between type arguments for traits and associated types for traits, one cannot implement `Iterator` on the same type with multiple `Item`s).

est31:
> I myself recently got confused and passed DER into an API that expected a `Vec<u8>` containing PEM. This was an SDK wrapping tonic, which does clearly mark the underlying fields as PEM.

Fair point, and I personally also had some confusing experiences where one Rust crypto library provided a key in one format and the other library was reading the key in the other format. In these instances, the chosen format has always been underdocumented (it's already over a year ago and I don't remember the exact details any more).

I think we can eventually merge this PR, if it documents the new API well, and by that I mean at each location where the new types are being used.

You are right that it's very easy to convert these types to/from byte buffers but it needs to be documented how, one shouldn't need the docs of the interface crate for that.

rcgen/src/key_pair.rs
Tudyx: In my opinion, we should limit the way a user can construct the type, to improve consistency and reduce question about which methods should be used. But I don't feel strongly about it, so I restored the `&[u8]/Vec<u8>` in the latest push

Tudyx:
I've rebased on the latest main and add some comments as suggested in the review.

rcgen/src/certificate.rs
```diff
@@ -269,8 +280,14 @@ impl CertificateParams {
 	/// This function assumes the provided certificate is a CA. It will not check
 	/// for the presence of the `BasicConstraints` extension, or perform any other
 	/// validation.
+	///
+	/// You can use [`rustls_pemfile::certs`] to get the `ca_cert` input. If
+	/// you have already a byte slice, just calling `into()` and take a reference
```
est31, line 285:
```suggestion
	/// you have already a byte slice, just calling `into()` and taking a reference
```

rcgen/src/csr.rs
```diff
@@ -36,15 +38,21 @@ impl CertificateSigningRequestParams {
...
+		Self::from_der(&csr.contents().into())
 	}
 
 	/// Parse a certificate signing request from DER-encoded bytes
 	///
 	/// Currently, this only supports the `Subject Alternative Name` extension.
 	/// On encountering other extensions, this function will return an error.
+	///
+	/// You can use [`rustls_pemfile::csr`] to get the `csr` input. If
+	/// you have already a byte slice, just calling `into()` and take a reference
```
est31, line 50:
```suggestion
	/// you have already a byte slice, just calling `into()` and taking a reference
```

rcgen/src/key_pair.rs
```diff
@@ -158,8 +160,14 @@ impl KeyPair {
 	/// Parses the key pair from the DER format
 	///
 	/// Equivalent to using the [`TryFrom`] implementation.
+	///
+	/// You can use [`rustls_pemfile::private_key`] to get the `der` input. If
+	/// you have already a byte slice, just calling `into()` and take a reference
```
est31, line 165:
```suggestion
	/// you have already a byte slice, just calling `into()` and taking a reference
```

rcgen/src/key_pair.rs
```diff
@@ -208,39 +223,45 @@ impl KeyPair {
 	/// key pair. However, sometimes multiple signature algorithms fit for the
 	/// same der key. In that instance, you can use this function to precisely
 	/// specify the `SignatureAlgorithm`.
+	///
+	/// You can use [`rustls_pemfile::private_key`] to get the `pkcs8` input. If
+	/// you have already a byte slice, just calling `into()` and take a reference
```
est31, line 228:
```suggestion
	/// you have already a byte slice, just calling `into()` and taking a reference
```

rcgen/src/key_pair.rs
```diff
@@ -249,14 +270,22 @@ impl KeyPair {
...
 			alg,
-			serialized_der: pkcs8_vec,
+			serialized_der,
 		})
 	}
 
+	/// Parses the key pair from the DER format
+	///
+	/// You can use [`rustls_pemfile::private_key`] to get the `pkcs8` input. If
+	/// you have already a byte slice, just calling `into()` and take a reference
```
est31, line 280:
```suggestion
	/// you have already a byte slice, just calling `into()` and taking a reference
```

est31 (approved):
good except the nits

djc:
Thanks for working on this, I'll just merge this now and might follow up with a PR with some nits.

cpu:
Nice work @Tudyx :clap:

PR #224, opened 2024-02-11
Add feature for ring's less-safe-getrandom-custom-or-rdrand

coder0xff:


est31:
I think it might be a better idea to suggest people to directly depend on ring? I'm not sure we should add feature flags for all of the feature flags of our dependencies, unless it's really a major feature.

coder0xff:
Perhaps. WebRTC is my true goal, and would have had to enable this flag. Maybe it will be easier to convince the maintainers not to use rcgen.

est31:
Not ethat I'm not suggesting to not use rcgen at all, just to *also* directly depend on ring and then enable the feature there. Then it will also be enabled for the ring that rcgen uses due to dependency unification.

coder0xff:
Ok, I misunderstood. Thanks for clarifying.

coder0xff:
Curious, does that mean that my project, which uses WebRTC, could do the same, directly depend on ring to set the flag, thereby bypassing the intermediate dependencies?

djc:
If you're enabling features that aren't enabled by default, yes, that should work.

PR #225, opened 2024-02-12
codecov: disable pull-request annotations

cpu:
By default Codecov leaves comments on diffs as ["annotations"](https://docs.codecov.com/docs/common-recipe-list#disable-github-check-run-annotations). These are separate from comments, which we disabled previously (https://github.com/rustls/rcgen/pull/187). This commit disables the annotations as well. Too much noise-to-signal right now.

Resolves https://github.com/rustls/rcgen/issues/186

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,3 @@
 comment: false
+github_checks:
+  annotations: false


djc (approved):

cpu:
Going to merge this with one review since it isn't a code change and moves towards something I think there's already consensus for.

est31 (approved):

PR #230, opened 2024-02-28
Add RSA key generation

### LINKED ISSUE #229: Support RSA-based sign algorithms
For now, this code generates a panic with the text `Cannot generate self-signed certificate: KeyGenerationUnavailable`:
```rust
let mut cert_params = CertificateParams::new(subject_alt_names);
cert_params.alg = &rcgen::PKCS_RSA_SHA256;
let cert = Certificate::from_params(cert_params).expect("Cannot generate self-signed certificate");
```

After searching, I found that note:
https://github.com/rustls/rcgen/blob/3c3e984e11efa4985ca596fe8ba917e6b676fe77/rcgen/src/key_pair.rs#L104-L109

As I search for now, `aws-lc-rs` has RSA signing support from version 1.6.0 via PR aws/aws-lc-rs#315. Full `aws-lc-rs` changelog can be found [here](https://github.com/aws/aws-lc-rs/releases/tag/v1.6.0).

Can you implement the RSA sign algorithm, please? I need this specific certificate generator.

est31:
Uses the recently added RSA key generation support of aws-lc-rs.

Fixes #229

### DIFF
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -5,6 +5,8 @@ use yasna::DERWriter;
 
 #[cfg(any(feature = "crypto", feature = "pem"))]
 use crate::error::ExternalError;
+#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+use crate::ring_like::rsa::KeySize;
 #[cfg(feature = "crypto")]
 use crate::ring_like::{
 	error as ring_error,
@@ -74,6 +76,9 @@ impl KeyPair {
 	/// Generate a new random key pair for the specified signature algorithm
 	///
 	/// If you're not sure which algorithm to use, [`PKCS_ECDSA_P256_SHA256`] is a good choice.
+	/// If passed an RSA signature algorithm, it depends on the backend whether we return
+	/// a generated key or an error for key generation being unavailable.
+	/// Currently, only `aws-lc-rs` supports RSA key generation.
 	#[cfg(feature = "crypto")]
 	pub fn generate_for(alg: &'static SignatureAlgorithm) -> Result<Self, Error> {
 		let rng = &SystemRandom::new();
@@ -101,13 +106,53 @@ impl KeyPair {
 					serialized_der: key_pair_serialized,
 				})
 			},
+			#[cfg(all(feature = "aws_lc_rs", not(feature = "ring")))]
+			SignAlgo::Rsa(sign_alg) => Self::generate_rsa_inner(alg, sign_alg, KeySize::Rsa2048),
 			// Ring doesn't have RSA key generation yet:
 			// https://github.com/briansmith/ring/issues/219
 			// https://github.com/briansmith/ring/pull/733
-			// Nor does aws-lc-rs:
-			// https://github.com/aws/aws-lc-rs/issues/296
-			SignAlgo::Rsa() => Err(Error::KeyGenerationUnavailable),
+			#[cfg(any(not(feature = "aws_lc_rs"), feature = "ring"))]
+			SignAlgo::Rsa(_sign_alg) => Err(Error::KeyGenerationUnavailable),
 		}
+	}
+
+	/// Generates a new random RSA key pair for the specified key size
+	///
+	/// If passed a signature algorithm that is not RSA, it will return
+	/// [`Error::KeyGenerationUnavailable`].
+	#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+	pub fn generate_rsa_for(
+		alg: &'static SignatureAlgorithm,
+		key_size: RsaKeySize,
+	) -> Result<Self, Error> {
+		match alg.sign_alg {
+			SignAlgo::Rsa(sign_alg) => {
+				let key_size = match key_size {
+					RsaKeySize::_2048 => KeySize::Rsa2048,
+					RsaKeySize::_3072 => KeySize::Rsa3072,
+					RsaKeySize::_4096 => KeySize::Rsa4096,
+				};
+				Self::generate_rsa_inner(alg, sign_alg, key_size)
+			},
+			_ => Err(Error::KeyGenerationUnavailable),
+		}
+	}
+
+	#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+	fn generate_rsa_inner(
+		alg: &'static SignatureAlgorithm,
+		sign_alg: &'static dyn RsaEncoding,
+		key_size: KeySize,
+	) -> Result<Self, Error> {
+		use aws_lc_rs::encoding::AsDer;
+		let key_pair = RsaKeyPair::generate(key_size)._err()?;
+		let key_pair_serialized = key_pair.as_der()._err()?.as_ref().to_vec();
+
+		Ok(KeyPair {
+			kind: KeyPairKind::Rsa(key_pair, sign_alg),
+			alg,
+			serialized_der: key_pair_serialized,
+		})
 	}
 
 	/// Parses the key pair from the DER format
@@ -375,6 +420,16 @@ impl TryFrom<Vec<u8>> for KeyPair {
 			serialized_der: pkcs8,
 		})
 	}
+}
+
+/// The key size used for RSA key generation
+#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum RsaKeySize {
+	_2048,
+	_3072,
+	_4096,
 }
 
 impl PublicKeyData for KeyPair {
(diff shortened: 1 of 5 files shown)


rcgen/src/sign_algo.rs
```diff
@@ -5,14 +5,14 @@ use yasna::DERWriter;
 use yasna::Tag;
 
 #[cfg(feature = "crypto")]
-use crate::ring_like::signature::{self, EcdsaSigningAlgorithm, EdDSAParameters};
+use crate::ring_like::signature::{self, EcdsaSigningAlgorithm, EdDSAParameters, RsaEncoding};
```
djc, line 8: I was confused by the lack of a `cfg` guard for this import, but I guess ring does actually have this type (it just doesn't allow signing with it).

djc (approved):
Nice!

rcgen/src/key_pair.rs
```diff
@@ -101,12 +101,25 @@ impl KeyPair {
 					serialized_der: key_pair_serialized,
 				})
 			},
+			#[cfg(all(feature = "aws_lc_rs", not(feature = "ring")))]
+			SignAlgo::Rsa(sign_alg) => {
+				use aws_lc_rs::encoding::AsDer;
+				let key_pair_doc =
```
cpu, line 107: What's the meaning of the "_doc" suffix on this var? I think it could be in-lined into L109 in either case.

rcgen/src/key_pair.rs
```diff
@@ -101,12 +101,25 @@ impl KeyPair {
...
+					RsaKeyPair::generate(crate::ring_like::rsa::KeySize::Rsa2048)._err()?;
```
cpu, line 108: Should there be a way to specify key size? I suspect there will be interest in 3072 bit keys at some point.

rcgen/src/key_pair.rs
```diff
@@ -101,12 +101,25 @@ impl KeyPair {
...
+				let key_pair_serialized = key_pair_doc.as_der()._err()?.as_ref().to_vec();
+
+				let key_pair = RsaKeyPair::from_pkcs8(&key_pair_serialized).unwrap();
```
cpu, line 111: Is there a reason that this `unwrap` is safe? probably deserves a comment,  or changing to `?` since this is already a fallible context.

cpu (commented):
Thanks for picking this up!

I think this should come with at least one unit test to prevent regressions and to demonstrate the feature works as intended.

It also looks like we could probably drop the dev dependency on the `rsa` crate now? It's only used in the `rsa-irc.rs` example for key generation and I think it would be better to show Rcgen doing this natively. The `rsa` crate has an [unresolved vulnerability](https://rustsec.org/advisories/RUSTSEC-2023-0071) that makes its removal from rcgen particularly tempting IMO.

cpu:
One more thought: It would be ideal to also see `rustls-cert-gen` updated w/ RSA keygen support alongside this work: https://github.com/rustls/rcgen/blob/3c3e984e11efa4985ca596fe8ba917e6b676fe77/rustls-cert-gen/src/cert.rs#L205-L212

rcgen/src/key_pair.rs
est31: The unwrap is safe because it parses something that's just been serialized.

rcgen/src/key_pair.rs
est31: Yeah the key size should be customizeable, but I'm not sure how to best expose that in the API. The `SignatureAlgorithm`s are independent of key size, and passing a parameter for all signature algorithms is also not great. Maybe a separate function that takes a `SignatureAlgorithm` plus a key size parameter and that panics if you pass it something non-RSA? IDK.

In general, the switch from 2048 to 4096 isn't *that* big of an improvement, it's not the square amount of computational complexity as one would think.

rcgen/src/key_pair.rs
cpu: Ahh ok, and the serializer and the parser are from the same crate, so that does seem reasonable :+1: 

I wouldn't mind a comment saying that explicitly, but also recognize it's fairly obvious in hindsight so I leave it up to you.

rcgen/src/key_pair.rs
est31: "document". I tried keeping the code analogous to the other match cases.

cpu: > I tried keeping the code analogous to the other match cases.

Sounds reasonable to me :+1:

rcgen/src/sign_algo.rs
est31: Yeah ring has this type too and it in fact requires it for signing, so we already use it where we do signing, but now the PR adds it to `SignatureAlgorithm`'s `SignAlgo::Rsa`. There is some potential followup where we do the signing to use the contained `RsaEncoding` directly instead of manually reconstructing it via an if-chain.

est31:
Thanks for the reviews, will look into addressing them tomorrow.

rcgen/src/key_pair.rs
```diff
@@ -101,15 +105,44 @@ impl KeyPair {
...
+			SignAlgo::Rsa(_sign_alg) => Err(Error::KeyGenerationUnavailable),
 		}
 	}
 
+	/// Generates a new random RSA key pair for the specified key size
+	///
+	/// If passed a signature algorithm that is not RSA, it will return
+	/// [`Error::KeyGenerationUnavailable`].
+	#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+	pub fn generate_rsa_for(alg: &'static SignatureAlgorithm) -> Result<Self, Error> {
```
djc, line 123: What is this method for? I don't think it could compile?

djc (approved):

rcgen/src/key_pair.rs
est31: Ah yeah sorry, I have pushed an unfinished state. I have fixed it now. The method is to allow customizing the key size (which defaults to 2048).

rcgen/src/key_pair.rs
est31: Removed the whole line. the construction is required for `EcdsaKeyPair` and `Ed25519KeyPair` because ring [doesn't return `Self`](https://docs.rs/ring/latest/ring/signature/struct.Ed25519KeyPair.html#method.generate_pkcs8) in the generator functions but a `Document` which needs to first be turned into a buffer that you then give to ring to parse it. But for RSA key generation, [the API simpler](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rsa/struct.KeyPair.html#method.generate), here `Self` is returned

rcgen/src/key_pair.rs
```diff
@@ -74,6 +76,8 @@ impl KeyPair {
 	/// Generate a new random key pair for the specified signature algorithm
 	///
 	/// If you're not sure which algorithm to use, [`PKCS_ECDSA_P256_SHA256`] is a good choice.
+	/// If passed an RSA signature algorithm, it depends on the backend whether we return
+	/// a generated key or an error for key generation being unavailable.
```
cpu, line 80: nit: might be helpful to mention which backend supports RSA here.

rustls-cert-gen/src/cert.rs
```diff
@@ -205,6 +205,7 @@ impl EndEntityBuilder {
 /// Supported Keypair Algorithms
 #[derive(Clone, Copy, Debug, Default, Bpaf, PartialEq)]
 pub enum KeyPairAlgorithm {
+	Rsa,
```
cpu, line 208: Should this enum variant maybe carry `RsaKeySize` now that the `rcgen` side supports specifying key size?

cpu (approved):
Thanks :+1: 

Did you have thoughts on dropping the `rsa` dev-dep for the one example? I think it's fine to omit in this branch, just curious if you're interested in the idea generally for potential follow-up.

est31:
> Did you have thoughts on dropping the `rsa` dev-dep for the one example?

Yeah we can drop it I think, but would prefer to do it in a followup.

rcgen/src/key_pair.rs
```diff
@@ -377,6 +421,16 @@ impl TryFrom<Vec<u8>> for KeyPair {
 	}
 }
 
+/// The key size used for RSA key generation
+#[cfg(all(feature = "crypto", feature = "aws_lc_rs", not(feature = "ring")))]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum RsaKeySize {
+	_2048,
```
djc, line 429: Nit: the `_` prefix looks unidiomatic to me, but not sure I have a better idea.

djc (approved):

rcgen/src/key_pair.rs
est31: Done

rcgen/src/key_pair.rs
est31: Yeah, not really perfect. `r#2048` doesn't work, but I like `_2048` more than `Rsa2048`.

rustls-cert-gen/src/cert.rs
est31: Maybe, yeah, happy to see followup PRs for it.

PR #233, opened 2024-02-29
Streamline signing API

djc:
IMO this is a more idiomatic API. I don't love the `sign_self()` name but haven't come up with anything better.

Alternatives:

* Change the names to be more passive voice (`signed_by()`?), since conceptually the params are what's being signed
* Move these methods to `KeyPair` instead -- this avoids the active/passive confusion but seems conceptually weird


cpu (commented):
> Change the names to be more passive voice (signed_by()?), since conceptually the params are what's being signed

I don't find the approach in this branch too awkward, but this option sounds a little bit better. Was there a reason you hesitated to select it up-front? I suppose it's more common for this sort of function to be named in an active tense?

rcgen/src/lib.rs
```diff
@@ -727,6 +726,48 @@ impl CertificateParams {
...
+			params: self,
+			subject_public_key_info,
+			der,
+		})
+	}
+	/// Generates a new self-signed certificate from the given parameters.
+	///
+	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
+	/// [`Certificate::pem`].
+	pub fn sign_self(self, key_pair: &KeyPair) -> Result<Certificate, Error> {
```
est31, line 762: I think `sign_self_signed` is a better name for this. IMO it's a good idea to add both a `sign` at the start and one at the end: the first one to indicate it's analogous with the `sign` function, the second because of the established `self-signed` term. The two occurrences of `sign` are weird indeed, yes, but I think it's the least weird name. But agreeing that every choice is going to be weird in some fashion.

rcgen/src/lib.rs
```diff
@@ -1602,83 +1643,6 @@ fn write_general_subtrees(writer: DERWriter, tag: u64, general_subtrees: &[Gener
 }
 
 impl Certificate {
-	/// Generates a new self-signed certificate from the given parameters.
-	///
-	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
-	/// [`Certificate::pem`].
-	pub fn generate_self_signed(
```
est31, line 1609: Is the next release planned to be a breaking one? If no, maybe we could keep these functions around but deprecate them.

djc:
Used `self_signed()` and `signed_by()` -- what do you think?

est31 (approved):
`self_signed` and `signed_by` are good names!

PR #237, opened 2024-03-07
Split certificate module out of crate root

djc:
Before this PR, `lib.rs` was 2258 lines long, and I found it quite hard to navigate. Here, I tried to split off a `certificate` module coming in at 1497 lines (1215 without tests), reducing `lib.rs` to 796 lines.

The new module also follows a top-down ordering, following the prevailing style in rustls projects. While some of the smaller modules seem to be top-down oriented, `lib.rs` so far was closer to bottom-up, but its lack of consistent ordering was definitely one of the reasons I've found it hard to navigate.

To build this, I took `Certificate` and `CertificateParams` and move them into a separate module. Any types that are mainly used via `Certificate` and `CertificateParams` are also moved. I have kept the OID definitions together (rather than moving the certificate-related ones to `certificate`) -- maybe we should move the ones that are only used in `certificate`?

At least for me, this will be much easier to navigate.


est31:
>  I have kept the OID definitions together (rather than moving the certificate-related ones to certificate) -- maybe we should move the ones that are only used in certificate?

I think this is better, let's have them all in one place, then one knows where to look it up.

rcgen/src/certificate.rs
```diff
@@ -0,0 +1,1496 @@
...
+use crate::key_pair::PublicKeyData;
+#[cfg(feature = "crypto")]
+use crate::ring_like::digest;
+#[cfg(feature = "pem")]
+use crate::ENCODE_CONFIG;
+use crate::{
+	write_distinguished_name, write_dt_utc_or_generalized, write_x509_authority_key_identifier,
+	write_x509_extension, DistinguishedName, Error, KeyIdMethod, KeyPair, KeyUsagePurpose, SanType,
+	SerialNumber, OID_BASIC_CONSTRAINTS, OID_COMMON_NAME, OID_COUNTRY_NAME,
+	OID_CRL_DISTRIBUTION_POINTS, OID_EXT_KEY_USAGE, OID_KEY_USAGE, OID_LOCALITY_NAME,
```
est31, line 21: Imports with these many lines are an anti-pattern as their interaction with rustfmt and text reflow is prone to cause merge conflicts. If you removed the `OID_` prefix from the oids and moved all the `OID_` related stuff to a separate `mod oid`, contained inside `lib.rs`, and then used the OIDs e.g. via `oid::BASIC_CONSTRAINTS`, then the number of lines of the import would be greatly reduced.

Alternatively, one could have a star import `use crate::oid::*`. I generally dislike star imports but if everything in the module is prefixed with the `OID_` name then it's pretty clear where the item is from.

est31 (approved):
ok except nit about import

rcgen/src/certificate.rs
djc: Good call! There was actually already an `oid` module, I just failed to notice it. Imported the module only in the new `certificate` module to reduce the number of imports here.

cpu:
Thank you! I've also found `lib.rs` to be unwieldy to navigate in the past. This is a nice change.

PR #238, opened 2024-03-08
Inline oid module

djc:
Following up on some suggestions from https://github.com/rustls/rcgen/pull/237#discussion_r1516701656.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -594,7 +594,7 @@ fn write_x509_authority_key_identifier(writer: DERWriter, aki: Vec<u8>) {
 	// In addition, for CRLs:
 	//    'Conforming CRL issuers MUST use the key identifier method, and MUST
 	//     include this extension in all CRLs issued.'
-	write_x509_extension(writer, OID_AUTHORITY_KEY_IDENTIFIER, false, |writer| {
+	write_x509_extension(writer, AUTHORITY_KEY_IDENTIFIER, false, |writer| {
 		writer.write_sequence(|writer| {
 			writer
 				.next()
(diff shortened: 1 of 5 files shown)


rcgen/src/lib.rs
```diff
@@ -663,6 +662,77 @@ impl AsRef<[u8]> for SerialNumber {
 	}
 }
 
+mod oid {
```
est31, line 665: I wasn't aware either that there has already been an oid module. If it's in a separate file we can keep it there (and maybe have pub(crate) instead of pub).

rcgen/src/lib.rs
```diff
@@ -32,37 +32,46 @@ println!("{}", key_pair.serialize_pem());
...
+pub use crl::{
 	CertificateRevocationList, CertificateRevocationListParams, CrlDistributionPoint,
 	CrlIssuingDistributionPoint, CrlScope, RevocationReason, RevokedCertParams,
 };
-pub use crate::csr::{CertificateSigningRequestParams, PublicKey};
-pub use crate::error::{Error, InvalidAsn1String};
-use crate::key_pair::PublicKeyData;
-pub use crate::key_pair::{KeyPair, RemoteKeyPair};
-use crate::oid::*;
+mod csr;
```
est31, line 61: I think it's nicer before the reorderings.

est31 (changes requested):

rcgen/src/lib.rs
djc: Changed this so it only moves the `mod` declarations up and removes the unnecessary `crate::` prefixes from the imports.

est31: > separating use from mod?

Yeah that separation is useful I think.

djc: So do you want to approve this in its current state?

est31 (approved):
Thanks!

rcgen/src/lib.rs
est31: Done, approved.

est31:
There is merge conflicts, needs a rebase before merging.

PR #239, opened 2024-03-08
Apply most Clippy suggestions

djc:
I temporarily disabled the `#![allow(clippy::complexity, clippy::style, clippy::pedantic)]` directive and implemented all the suggestions that I thought improved the code.


rcgen/src/certificate.rs
```diff
@@ -1111,26 +1111,6 @@ macro_rules! mask {
 }
 
 impl CidrSubnet {
-	/// Obtains the CidrSubnet from the well-known
-	/// addr/prefix notation.
-	/// ```
-	/// # use std::str::FromStr;
```
djc, line 1117: Note that the example here actually imports `FromStr` as if it was needed...

est31 (approved):
ok except for the new -> default change. new is nicer than the longer function name. Also don't like the deprecation. One can implement default though as an alternative.

PR #240, opened 2024-03-08
Serialize CRL parameters into CertificateRevocationList

djc:
Updates the CRL creation flow along the lines of #205.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -6,7 +6,7 @@ use pem::Pem;
 use pki_types::{CertificateDer, CertificateSigningRequestDer};
 use time::{Date, Month, OffsetDateTime, PrimitiveDateTime, Time};
 use yasna::models::ObjectIdentifier;
-use yasna::{DERWriter, DERWriterSeq, Tag};
+use yasna::{DERWriter, Tag};
 
 use crate::crl::CrlDistributionPoint;
 use crate::csr::CertificateSigningRequest;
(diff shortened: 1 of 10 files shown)


rcgen/src/crl.rs
```diff
@@ -196,39 +174,24 @@ pub struct CertificateRevocationListParams {
...
-		})
-	}
-	fn write_crl(
-		&self,
-		writer: DERWriter,
-		issuer: &KeyPair,
-		issuer_name: &DistinguishedName,
-	) -> Result<(), Error> {
-		writer.write_sequence(|writer| {
+	/// Serializes the certificate revocation list (CRL) in binary DER format, signed with
```
est31, line 177:
```suggestion
	/// Serializes the certificate revocation list (CRL), signed with
```

Internally we use DER format but we return a wrapper around the DER format.

djc:
Reviewed the docstrings -- should be better now -- and rebased on top of current main.

cpu (approved):
Thanks, looks nice to me :+1:

djc:
Some updates:

* Rebased
* Split up two-step CRL serialization commit into a few easier to review parts
* Made the API more consistent

rcgen/src/certificate.rs
```diff
@@ -467,20 +468,18 @@ impl CertificateParams {
...
 	/// and include the subject public key information from `subject_key`. Additionally,
 	/// the CSR will be self-signed using the subject key.
 	///
-	/// **Important**: since this function generates a new CSR for each invocation you
-	/// should not call `serialize_request_der` and then `serialize_request_pem`. This will
-	/// result in two different CSRs. Instead call only `serialize_request_pem` and base64
-	/// decode the inner content to get the DER encoded CSR using a library like `pem`.
-	pub fn serialize_request_der(
+	/// Note that subsequent invocations of `serialize_request()` will not produce exactly
+	/// the same output.
```
cpu, line 478: editorial nit: I think this will read more naturally as "will not produce the exact same output" instead of "will not produce exactly the same output"

cpu (approved):

est31 (approved):

PR #242, opened 2024-03-13
Support more private key formats when using aws_lc_rs feature

Alvenix:
I hope this change is okay. I am not sure if I need to add tests for this? Also should I rename the function in ring-like?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ resolver = "2"
 [workspace.dependencies]
 aws-lc-rs = "1.6.0"
 pem = "3.0.2"
-pki-types = { package = "rustls-pki-types", version = "1.3.0" }
+pki-types = { package = "rustls-pki-types", version = "1.4.1" }
 rand = "0.8"
 ring = "0.17"
 x509-parser = "0.16"
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -3,11 +3,13 @@ use std::fmt;
 #[cfg(feature = "pem")]
 use pem::Pem;
 #[cfg(feature = "crypto")]
-use pki_types::PrivatePkcs8KeyDer;
+use pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
 use yasna::{DERWriter, DERWriterSeq};
 
 #[cfg(any(feature = "crypto", feature = "pem"))]
 use crate::error::ExternalError;
+#[cfg(all(feature = "crypto", feature = "aws_lc_rs"))]
+use crate::ring_like::ecdsa_from_private_key_der;
 #[cfg(all(feature = "crypto", feature = "aws_lc_rs"))]
 use crate::ring_like::rsa::KeySize;
 #[cfg(feature = "crypto")]
@@ -165,13 +167,15 @@ impl KeyPair {
 
 	/// Parses the key pair from the ASCII PEM format
 	///
-	/// The key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
+	/// If `aws_lc_rs` feature is used, then the key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958, SEC1/RFC 5915, or PKCS#1/RFC 3447;
+	/// Appears as "PRIVATE KEY", "RSA PRIVATE KEY", or "EC PRIVATE KEY" in PEM files.
 	///
+	/// Otherwise if the `ring` feature is used, then the key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
 	/// Appears as "PRIVATE KEY" in PEM files.
 	#[cfg(all(feature = "pem", feature = "crypto"))]
 	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
 		let private_key = pem::parse(pem_str)._err()?;
-		Self::try_from(&PrivatePkcs8KeyDer::from(private_key.contents()))
+		Self::try_from(private_key.contents())
 	}
 
 	/// Obtains the key pair from a raw public key and a remote private key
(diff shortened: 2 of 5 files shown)


rcgen/src/ring_like.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) fn ecdsa_from_pkcs8(
 
 	#[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
 	{
-		Ok(signature::EcdsaKeyPair::from_pkcs8(alg, pkcs8)._err()?)
+		Ok(signature::EcdsaKeyPair::from_private_key_der(alg, pkcs8)._err()?)
```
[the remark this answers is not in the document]
djc: I don't think we should add a function with whose semantics we can't clearly document because we don't understand them ourselves, so I think you should figure out what the exact format is that OpenSSL generated here, and add API for that -- ideally we'd get aws-lc-rs to more explicitly document what their function support, too.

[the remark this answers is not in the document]
djc: We already support PKCS#8 throughout, but I think it would be okay to add a function that takes a `PrivateSec1KeyDer` and funnels that into this aws-lc-rs API.

djc:
> > > Is the direction I am going with this PR okay?
> > 
> > I think names like `_with_more_formats()` do not match with what I suggested in [#242 (comment)](https://github.com/rustls/rcgen/pull/242#discussion_r1524575334), and I would probably not want to add such API.
> 
> Yes I understand but I want to find a way that make it easiest to the user to parse a private key of either, sec1, pkcs1, pkcs8. I feel if each format has an API specific to them. Many (at least me) would have to write a wrapper that tries them all.
> 
> What about changing the original APIs to specify pkcs8 in it is name, and keeping the generic one without? Similar to the following:
> 
> ```
> #![cfg(feature = "aws_lc_rs")]
> fn from_der_and_sign_algo(key: &PrivateKeyDer<'_>)
> 
> fn from_der_and_sign_algo_pkcs8(key: &PrivatePkcs8KeyDer<'_>)
> ```
> 
> We could also not export these APIs but use them inside from_raw.

If we can add tests that show that the aws-lc-rs API accepts all of PKCS#8, PKCS#1 and SEC1, I think something like this could make sense? I would move the `pkcs8_` before `der` to make the name `from_pkcs8_der_and_sign_algo()`.

djc:
Look at the `From` impls for `PrivateKeyDer`?

cpu:
>  I would move the pkcs8_ before der to make the name from_pkcs8_der_and_sign_algo().

I think it makes sense to make this change ahead of the next release so we can land this branch as a follow-up without needing any breaking API changes: https://github.com/rustls/rcgen/pull/249

djc:
We discussed this on Discord and we think it might make sense to have `TryFrom<&[u8]>` and `TryFrom<Vec<u8>>` implementations for `PrivateKeyDer` (in rustls-pki-types) that do some minimal DER parsing to determine whether the passed in DER is consistent with PKCS#1, PKCS#8 or SEC1 or error out otherwise. Would you be willing to take that on?

djc (approved):
This is looking great modulo some nits!

cpu (approved):
This looks good to me, thank you. I think once we sort out the pki-types side (https://github.com/rustls/pki-types/pull/40) and publish a release this would be ready to go.

cpu:
> I think once we sort out the pki-types side (https://github.com/rustls/pki-types/pull/40) and publish a release

This was done and [pki-types 1.4.0](https://github.com/rustls/pki-types/releases/tag/v%2F1.4.0) has the required change.

@Alvenix When you have a chance can you update your branch to take that release and drop the patch? Thanks!

cpu:
Thanks!

> PR https://github.com/rustls/rcgen/pull/252 should be merged before. So I can rebase on main.

Agreed, I think we're just waiting on est31 to +1 it.

djc:
#252 has been merged. If you can rebase soon this can be released at the same time.

djc:
@est31 any thoughts?

rcgen/src/key_pair.rs
```diff
@@ -319,6 +325,138 @@ impl KeyPair {
...
+		Ok(KeyPair {
+			kind,
+			alg,
+			serialized_der,
+		})
+	}
+
+	/// Parses the key pair from the DER format
+	///
+	/// You can use [`rustls_pemfile::private_key`] to get the `key` input. If
```
est31, line 417: I don't think we need rustdoc for this internal function outside of the first line.

est31:
Can we maybe offer the new name everywhere, on ring-only builds being a wrapper to the pkcs8 version?

est31 (approved):

rcgen/src/key_pair.rs
```diff
@@ -319,6 +323,167 @@ impl KeyPair {
...
+	/// you have already a byte slice, just calling `try_into()` will convert it to a [`PrivateKeyDer`].
+	///
+	/// [`rustls_pemfile::private_key`]: https://docs.rs/rustls-pemfile/latest/rustls_pemfile/fn.private_key.html
+	#[cfg(feature = "crypto")]
+	pub(crate) fn from_raw(
```
djc, line 438: One suggestion: if possible I would move this code into a `impl TryFrom<&PrivateKeyDer> for KeyPair`, skipping the `from_raw()` indirection.

djc: I mean pretty much moving all of it, so removing `from_raw()` (which we definitely don't want to duplicate). Is that not feasible?

rcgen/src/key_pair.rs
```diff
@@ -659,12 +542,9 @@ impl TryFrom<&PrivatePkcs8KeyDer<'_>> for KeyPair {
 	type Error = Error;
 
 	fn try_from(pkcs8: &PrivatePkcs8KeyDer) -> Result<KeyPair, Error> {
-		let (kind, alg) = KeyPair::from_pkcs8_raw(pkcs8)?;
-		Ok(KeyPair {
-			kind,
-			alg,
-			serialized_der: pkcs8.secret_pkcs8_der().into(),
-		})
+		let key = PrivateKeyDer::from(pkcs8);
```
djc, line 545: Nit: just skip the binding, `PrivateKeyDer::from(pkcs8).try_into()` should do it?

djc (approved):
LGTM, great work!

djc:
My other question is whether it still makes sense to expose `from_pkcs8_der_and_sign_algo()` now that we have the more general method?

rcgen/src/key_pair.rs
```diff
@@ -673,7 +551,73 @@ impl TryFrom<&PrivateKeyDer<'_>> for KeyPair {
 	type Error = Error;
 
 	fn try_from(key: &PrivateKeyDer) -> Result<KeyPair, Error> {
-		let (kind, alg) = KeyPair::from_raw(key)?;
+		#[cfg(all(feature = "ring", not(feature = "aws_lc_rs")))]
+		let (kind, alg) = {
+			if let PrivateKeyDer::Pkcs8(pkcs8) = key {
```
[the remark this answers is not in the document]
djc: Definitely!

Cargo.toml
```diff
@@ -5,11 +5,15 @@ resolver = "2"
 [workspace.dependencies]
 aws-lc-rs = "1.6.0"
 pem = "3.0.2"
-pki-types = { package = "rustls-pki-types", version = "1.3.0" }
+pki-types = { package = "rustls-pki-types", version = "1.4.0" }
```
cpu, line 8: If you bump this to [1.4.1](https://github.com/rustls/pki-types/releases/tag/v%2F1.4.1) we can drop the patch below.

rcgen/src/key_pair.rs
```diff
@@ -165,13 +167,15 @@ impl KeyPair {
 
 	/// Parses the key pair from the ASCII PEM format
 	///
-	/// The key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958;
+	/// If `aws_lc_rs` feature is used, then the key must be a DER-encoded plaintext private key; as specified in PKCS#1/RFC 3447, RFC 5915, or PKCS #8/RFC 5958;
```
cpu, line 170: two small suggestions:

1. Let's order this PKCS8, SEC1, PKCS1 - I think that's ordering it most common to least common. (Especially since only PKCS8 was supported before).
2. Let's mention "SEC1" alongside RFC 5915 since I think people will be pattern matching on the first more than reading the linked RFC.

```suggestion
	/// If `aws_lc_rs` feature is used, then the key must be a DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958, SEC1/RFC 5915, or PKCS#1/RFC 3447;
```

cpu (approved):
Thanks!

est31:
> whether it still makes sense to expose from_pkcs8_der_and_sign_algo() now that we have the more general method?

There might be instances where you only want to support pkcs8 because some standard or so requires it to be pkcs8. Validation plus loading with something more permissive is prone to exploitable bugs where you trick the validator then get to the loading stage and get treated differently.

I'm not sure it's enough justification to have all parsing functions be present both for pkcs8 and general formats though, maybe we just have the lowest level

djc:
> There might be instances where you only want to support pkcs8 because some standard or so requires it to be pkcs8. Validation plus loading with something more permissive is prone to exploitable bugs where you trick the validator then get to the loading stage and get treated differently.

If you really want PKCS#8 you can `try_from()` `PrivateKeyDer::Pkcs8(_)`, right? I think this API still facilitates that.

PR #244, opened 2024-03-14
Unbreak doctests

djc:
#233 regressed a doctest for CRL signing. It turns out CI was using `test --all-targets` which is documented as omitting doctests:

```
djc-2021 ci-doctests rcgen $ cargo test -h
Execute all unit and integration tests and build examples of a local package

[..]

Target Selection:
      [..]
      --all-targets       Test all targets (does not include doctests)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
       - name: cargo doc (all features)
         run: cargo doc --all-features --document-private-items
         env:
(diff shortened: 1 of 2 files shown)


est31:
at least according to my [4 year old and probably outdated table](https://github.com/rust-lang/cargo/pull/8437#issuecomment-653842297) there is no single cargo command that tests everything, `cargo test` misses benches for example. But that's fine as we don't have benches (but we do have doctests so this is an improvement).

est31 (approved):

PR #245, opened 2024-03-14
lib: export key_pair::RsaKeySize

cpu:
The new `KeyPair::generate_rsa_for` fn from https://github.com/rustls/rcgen/pull/230 is only usable outside the crate if the `key_pair::RsaKeySize` type is exported from `lib.rs` since `key_pair` is a private module.


djc (approved):

cpu:
This feels small & obvious enough to merge with one review.

est31:
Thanks, I've missed this one. Good catch!

PR #247, opened 2024-03-18
Add rsa keypair generation using the rsa crate.

uglyoldbob:
This pull request add capability of generating a 4096 bit keypair using the rsa crate. I wasn't sure how to get the number of bits for the keypair so I assumed 4096.

### DIFF
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -26,14 +26,17 @@ aws-lc-rs = { version = "1.0.0", optional = true }
 yasna = { version = "0.5.2", features = ["time", "std"] }
 ring = { version = "0.17", optional = true }
 pem = { workspace = true, optional = true }
+rand = { version = "0.8.5", optional = true }
+rsa = { version = "0.9.6", optional = true }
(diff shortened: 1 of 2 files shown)


cpu:
Hi there,

We recently added (_but haven't released_) support for RSA key generation using aws-lc-rs: https://github.com/rustls/rcgen/pull/230 Does this meet your need? 

I suspect there's not much interest in having two different dependencies for this feature, and the rsa crate has at least one [unresolved vulnerability](https://rustsec.org/advisories/RUSTSEC-2023-0071.html) at the present time.

est31:
I'm not sure they want to fix that vulnerability any time soon, and it's also in a code path that we'd use. I think it's okay to have multiple backends, but the vulnerability is a bit off-putting from integrating the `rsa` crate directly into rcgen. The rsa-irc example shows how one can use the rsa crate as a rcgen user to generate key pairs for rcgen. It's not *that* inconvenient.

est31:
Thanks for the PR. Closing, will reconsider once https://github.com/RustCrypto/RSA/issues/19 is fixed and released.

PR #241, opened 2024-03-13
Support ECDSA_P521_SHA512 when using aws_lc_rs feature

Alvenix:
Hope I didn't miss anything.

PR #248, opened 2024-03-19
Fix unused import warning

Alvenix:
This PR is remove a warning as requested on [241](https://github.com/rustls/rcgen/pull/241#issuecomment-2006224451).

Alternatively I could use `use super::super::*` in relevant tests and get rid of the conditional compilation. I feel that way could be easier to maintain.


djc (approved):

est31 (approved):

cpu:
Thanks for fixing this.

Issue #236, opened 2024-03-07
0.13.0 release planning


sigoden:
There are two updates that are very important to me, I hope to see the new release soon.

1. Use `ring@0.17`. `rcgen@0.12.1` still has a dependency on `ring@0.16`
```
ring v0.16.20
└── x509-parser v0.15.1
     └── rcgen v0.12.1
```
2. RSA support

Thank you all.

djc:
I think it would be great to release -- I think it would be nice to get #233 and #223 in, though.

est31:
FTR the way x509-parser 0.15.1 depends on old ring is [through a direct dependency](https://github.com/rusticata/x509-parser/blob/x509-parser-0.15.1/Cargo.toml), not because it depends on rcgen. That would be a circular dependency which Cargo doesn't allow (outside of dev dependencies but this isn't the case here).

> I think it would be great to release -- I think it would be nice to get https://github.com/rustls/rcgen/pull/233 and https://github.com/rustls/rcgen/pull/223 in, though.

I agree, IMO it's a good moment to release, we've had multiple larger changes and fully moving to newest ring is a good motivation.

cpu:
I think once https://github.com/rustls/rcgen/pull/241 lands we should consider cutting the release. https://github.com/rustls/rcgen/pull/242/files looks like it could use more time (not yet approved, no tests).

djc:
Makes sense. Maybe we can do some renaming in advance of #242 so that #242 will be non-breaking?

cpu:
> Maybe we can do some renaming in advance of https://github.com/rustls/rcgen/pull/242 so that https://github.com/rustls/rcgen/pull/242 will be non-breaking?

I'm supportive of the idea but I'm not sure I know what you have in mind. #242 doesn't seem to change any existing fns, but introduces new ones (`KeyPair::from_pem_and_sign_algo_with_more_formats`, `KeyPair::from_der_and_sign_algo_with_more_formats`). It doesn't seem breaking to me, or I'm overlooking where some staged renaming will help.

djc:
I didn't like those APIs so there is a different proposal in the comments.

cpu:
> I didn't like those APIs so there is a different proposal in the comments.

Aha! Thanks, I understand the proposal now.

djc:
I think we'd like to get #252 in, but after that I guess we're ready.

cpu:
[...]

PR #249, opened 2024-03-19
key_pair: emphasize PKCS8 input requirement in constructor fn names

cpu:
This branch pulls out the [suggested rename](https://github.com/rustls/rcgen/pull/242#issuecomment-2006944259) from https://github.com/rustls/rcgen/pull/242 in order to unblock release planning (https://github.com/rustls/rcgen/issues/236) while the more generic key pair constructor is worked out.


est31 (approved):
very good change

djc (approved):

PR #250, opened 2024-03-19
Add ability to get a PublicKey from a KeyPair.

uglyoldbob:
This patch adds the ability to get a PublicKey struct directly from a KeyPair. I used this method to generate a csr and sign it with another certificate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,7 +693,7 @@ dependencies = [
 
 [[package]]
 name = "rcgen"
-version = "0.12.0"
+version = "0.12.1"
 dependencies = [
  "aws-lc-rs",
  "botan",
--- a/rcgen/CHANGELOG.md
+++ b/rcgen/CHANGELOG.md
@@ -1,6 +1,15 @@
 
 # Changes
 
+## Release 0.12.1 - January 25th, 2024
+
+- RFC 5280 specifies that a serial number must not be larger than 20 octets in
+  length. Prior to this release an unintended interaction between rcgen and its
+  underlying DER encoding library could result in 21 octet serials. This has now
+  been fixed.
+- A regression that caused build errors when the optional `pem` feature was
+  omitted has been fixed.
+
 ## Release 0.12.0 - December 16, 2023
 
 - Rename `RcgenError` to `Error`. Contributed by [thomaseizinger](https://github.com/thomaseizinger).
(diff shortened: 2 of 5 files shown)


djc:
This looks sensible to me, provided you can fix up the failing CI jobs?

est31 (approved):

cpu:
>  I used this method to generate a csr and sign it with another certificate.

@uglyoldbob Can you expand on what you were trying to do here? 

I think the existing API would support the use-case you describe as follows:

1. Constructing a `KeyPair` somehow
2. Constructing `CertificateParams`.
3. Serializing a `CertificateSigningRequest` for the `KeyPair` with `CertificateParams::serialize_request(key_pair)`

I think you only need a way to construct a `PublicKey` from a `KeyPair` if you wanted to instantiate ` CertificateSigningRequest` yourself instead of using `serialize_request` and I'm not sure that's something we should aim to support. The `PublicKey` type is pretty useless outside of the crate since the `alg` and `raw` fields, and the `PublicKeyData` trait, are both crate internal.

uglyoldbob:
Here is what I am doing that uses the public key method. [code here](https://github.com/uglyoldbob/iot/blob/master/src/ca.rs#L388)
I don't see CertificateParams::serialize_request in the latest release (0.12.1).

cpu:
> I don't see CertificateParams::serialize_request in the latest release (0.12.1).

Indeed - we have a number of API changes queued in `main` that aren't in a release yet.

With 0.12.1's api you would:

* Construct a `KeyPair`, `key_pair`
* Construct `CertificateParams`, `params`.
* Set the `params.key_pair` to `Some(key_pair)`
* Construct a self-signed certificate with `Certificate::from_params(params)`, `cert`
* Serialize a CSR from the `Certificate` with [`cert.serialize_request_der`](https://docs.rs/rcgen/latest/rcgen/struct.Certificate.html#method.serialize_request_der), `csr_der`
* Create a `CertificateSigningRequest` with `CertificateSigningRequest::from_der(csr_der)`, `csr`
* Issue a cert from the CSR with `csr.serialize_der_with_signer(some_issuer_cert)`

It's certainly more steps in the 0.21.1 API, but still appears possible without modification to rcgen. 

I think with the API in `main` this change isn't required. Do you agree?

cpu:
I'm going to go ahead and close this out. I think if we hadn't done the API rework this could have been a nice way to avoid a cumbersome step where you need to issue a self-signed certificate but the new API provides an ergonomic way to make a CSR from just params and a key pair.

Thanks again for the contribution!

PR #251, opened 2024-03-20
Streamline CI jobs

djc:
I've felt that CI in this project is relatively slow. Try to make it a bit faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,23 +11,8 @@ env:
   RUSTFLAGS: -D warnings
 
 jobs:
-  rustfmt:
-    name: Format
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout sources
-        uses: actions/checkout@v4
-        with:
-          persist-credentials: false
-      - name: Install rust toolchain
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          components: rustfmt
-      - name: Check formatting
-        run: cargo fmt -- --check
-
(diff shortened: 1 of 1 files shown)


djc:
Hmm, it seems like the Windows and macOS builds are the long pole here. It looks like we have some caching in place for the Windows build at least, but it doesn't seem to be particularly effective?

cpu:
>  It looks like we have some caching in place for the Windows build at least, but it doesn't seem to be particularly effective?

Indeed, I tried to cache the `vcpkg` stuff that seemed to be the main source of slowness (IIRC it's mainly fetching botan build deps?), but I'm not confident it actually worked the way I intended.

I wonder if it would be workable to limit the botan and openssl testing that require more native dependencies to fewer OS targets and assume the upstream projects are covering the platform specific aspects?

cpu (approved):
These seem like uncontroversial improvements to me :+1:

djc:
> I wonder if it would be workable to limit the botan and openssl testing that require more native dependencies to fewer OS targets and assume the upstream projects are covering the platform specific aspects?

Seems pretty reasonable to me -- after all, I don't think we have any platform-specific code in this project?

est31 (approved):
good changes

cpu:
@djc I think you might need to administratively merge this, or adjust the branch protection rules. The auto-merge is waiting on removed CI tasks to complete.

djc:
Adjusted the branch protection rules to remove the rules that this PR got rid of.

PR #252, opened 2024-03-20
Prefer aws-lc-rs over ring if both are enabled

### LINKED ISSUE #243: figure out something for aws_lc_rs's superset of features
`aws_lc_rs` has a superset of the features of *ring*, but right now we default to using ring's types when both are specified. This was fine until #230, but that PR made it a problem. It hurts the additive nature of cargo's features, ideally the superset of both backends should work.

I see two possible solutions:

1. use aws_lc_rs if both ring and aws_lc_rs are specified. this does *not* mean that aws_lc_rs should be turned on by default, that is not a good idea. this approach works until ring gets a feature aws_lc_rs doesn't provide. I don't think this is something we need to worry about though (it's unlikely). 
2. using both ring and aws_lc_rs if both are specified, ring in the cases where only ring related features are used, aws_lc_rs otherwise. This is manageable for rsa key generation and  #241 , but becomes more complex with more complex features, like entire key pair kinds.

Personally I prefer 1 but would be ok with both. Opinions @djc @cpu?

djc:
Fixes #243.

### DIFF
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -23,7 +23,7 @@ required-features = ["pem", "x509-parser"]
 
 [[example]]
 name = "simple"
-required-features = ["crypto"]
+required-features = ["crypto", "pem"]
 
 [dependencies]
 aws-lc-rs = { workspace = true, optional = true }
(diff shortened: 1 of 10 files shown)


cpu (approved):

djc:
@est31 friendly ping? I think this is the last blocker for the release.

est31 (approved):
LGTM, thanks!

est31:
Merged this via a squash merge because some CI job was never successful.

PR #254, opened 2024-03-25
docs: update CHANGELOG for 0.13.0

cpu:
I've included substantially more detail in this CHANGELOG update compared to previous. To try and keep the top level content manageable, while also balancing making updates as friction-less as possible, I've split off some separate migration help to its own document.

I've done my best to flag all of the important changes but would appreciate more eyes to make sure I haven't missed or misrepresented anything. Like previous updates I've called out contributions by external folks and not those by core maintainers.

### DIFF
--- a/rcgen/CHANGELOG.md
+++ b/rcgen/CHANGELOG.md
@@ -1,6 +1,75 @@
 
 # Changes
 
+## Release 0.13.0 - March XX, 2024
+
+Breaking changes:
+
+- The API used to create/issue key pairs, certificates, certificate signing
+  requests (CSRs), and certificate revocation lists (CRLs) has been
+  restructured to emphasize consistency and avoid common errors with
+  serialization.
+
+  For each concrete type (cert, CSR, CRL) the process is now the same:
+
+  0. generate or load a key pair and any information about issuers required.
+  1. create parameters, customizing as appropriate.
+  2. call a generation `fn` on the parameters, providing subject key pair and
+     issuer information and as appropriate.
+  3. call serialization `fn`s on the finalized type, obtaining DER or PEM.
+
+  For more information, see [rcgen/docs/0.12-to-0.13.md].
+
+- Throughout the API DER inputs are now represented using types from the Rustls
+  `rustls-pki-types` crate, e.g. `PrivateKeyDer`, `CertificateDer`,
+  `CertificateSigningRequestDer`. Contributed by
+  [Tudyx](https://github.com/tudyx).
+
+- String types used in `SanType` and `DnValue` enums for non-UTF8 string types
+  have been replaced with more specific types that prevent representation of
+  illegal values. E.g. `Ia5String`, `BmpString`, `PrintableString`,
+  `TeletexString`, and `UniversalString`. Contributed by
+  [Tudyx](https://github.com/tudyx).
+
+- Method names starting with `get_` have been renamed to match Rust convention:
+  `CertificateRevocationList::get_params()` -> `params()`
+  `Certificate::get_params()` -> `params()`
+  `Certificate::get_key_identifier()` -> `Certificate::key_identifier()`
+  `Certificate::get_times()` -> `Certificate::times()`
+
+Added:
+
+- RSA key generation support has been added. This support requires using the
+  `aws-lc-rs` feature. By default using `KeyPair::generate()` with
+  an RSA `SignatureAlgorithm` will generate an RSA 2048 keypair. See
+  `KeyPair::generate_rsa_for()` for support for RSA 2048, 3072 and 4096 key sizes.
+
+- Support for ECDSA P521 signatures and key generation has been added when using
+  the `aws-lc-rs` feature. Contributed by [Alvenix](https://github.com/alvenix).
+
+- Support for loading private keys that may be PKCS8, PKCS1, or SEC1 has been
+  added when using the `aws-lc-rs` feature. Without this feature private keys
+  must be PKCS8. See `KeyPair::from_pem_and_sign_algo()` and
+  `KeyPair::from_der_and_sign_algo()` for more information. Contributed by
+  [Alvenix](https://github.com/alvenix).
+
+- Support has been added for Subject Alternative Name (SAN) names of type
+  `OtherName`. Contributed by [Tudyx](https://github.com/tudyx).
+
+- Support has been added for specifying custom "other" OIDs in extended key
+  usage. Contributed by [Tudyx](https://github.com/tudyx).
+
+- Support has been added for building rcgen _without_ cryptography by omitting
+  the new (default-enabled) `crypto` feature flag. Contributed by
+  [corrideat](https://github.com/corrideat).
+
+- Support for using `aws-lc-rs` in `fips` mode can now be activated by using the
+  `fips` feature in combination with the `aws-lc-rs` feature. Contributed by
+  [BiagioFesta](https://github.com/biagiofesta).
+
+- A small command-line tool for certificate generation (`rustls-cert-gen`) was
+  added. Contributed by [tbro](https://github.com/tbro).
+
 ## Release 0.12.1 - January 25th, 2024
 
 - RFC 5280 specifies that a serial number must not be larger than 20 octets in
--- a/rcgen/docs/0.12-to-0.13.md
+++ b/rcgen/docs/0.12-to-0.13.md
@@ -1,0 +1,90 @@
+# Rcgen 0.12 to 0.13 Migration Guide
+
+This document is a meant to be a helpful guide for some of the API changes made
+between rcgen 0.12 and 0.13. For information on other changes in 0.13 see
+[rcgen/CHANGELOG.md].
+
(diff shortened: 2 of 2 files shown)


djc:
I would reorder these with breaking changes before added and with both lists ordered from most to least impact.

cpu:
> I would reorder these with breaking changes before added and with both lists ordered from most to least impact.

Good idea. I've done a reordering pass.

rcgen/CHANGELOG.md
```diff
@@ -1,6 +1,101 @@
...
+  serialized the `Certificate` with `Certificate::serialize_der`,
+  `Certificate::serialize_der_with_signer`, `Certificate::serialize_pem` and
+  `Certificate::serialize_pem_with_signer`. These functions have been removed.
+
+  With the updated API you create a `Certificate` directly from
+  `CertificateParams` using `CertificateParams::self_signed` or
+  `CertificateParams::signed_by`. You can serialize the `Certificate` to PEM or
+  DER using `Certificate::der()` or `Certificate::pem`.
+
+- `KeyPair` generation has been separted from certificate issuance. The
```
djc, line 21: "separted"

rcgen/CHANGELOG.md
```diff
@@ -1,6 +1,101 @@
...
+  `Certificate::get_key_pair`, `Certificate::serialize_private_key_der` and
+  `Certificate::serialize_private_key_pem` fns have been removed. The
+  `Certificate` and `CertificateParams` types now only holds public data.
```
djc, line 24: "holds" > "hold"

rcgen/CHANGELOG.md
```diff
@@ -1,6 +1,101 @@
...
+  Issuance APIs now expect to be given both a `Certificate` and a signing
+  `KeyPair`, and the caller must generate their own `KeyPair` up-front using
+  `KeyPair::generate`, `KeyPair::generate_for`, or `KeyPair::generate_rsa_for`.
+
+- Certificate signing request creation and issuance has been reworked for
```
djc, line 29: I think we need a similar section for CRL signing?

rcgen/CHANGELOG.md
```diff
@@ -1,6 +1,101 @@
...
+
+  Previously issuance of a certificate from a CSR could be done using
+  `CertificateSigningRequest::serialize_der_with_signer` or
+  `CertificateSigningRequest::serialize_pem_with_signer`. This is now done using
+  `CertificateSigningRequestParams::signed_by`, and serialization to DER/PEM can
+  be done by calling `Certificate::pem` or `Certificate::der` on the resulting
+  `Certificate`.
+
+- Throughout the API DER inputs are now represented using types from the Rustls
+  `pki-types` crate, e.g. `PrivateKeyDer`, `CertificateDer`,
```
djc, line 46: I think we should call it by its crates.io name, rustls-pki-types.

djc (commented):
Order looks great!

I think it's great to spend more words on the larger changes in the signing API, but (especially when also adding CRLs) I wonder if it could be done more concisely by explaining the general principles and abstracting as much as possible over certificate/CSR/CRL. I've tried hard to make the APIs as similar as possible for each of them.

I like to retain the `()` suffix on methods/functions because it makes it more obvious to distinguish these from types/variables.

cpu:
>  I wonder if it could be done more concisely by explaining the general principles and abstracting as much as possible over certificate/CSR/CRL.

Yeah, I think that's good feedback :+1: I will try and rework along these lines tomorrow.

rcgen/CHANGELOG.md
cpu: Fixed.

rcgen/CHANGELOG.md
cpu: Fixed.

rcgen/CHANGELOG.md
cpu: Added coverage for CRLs.

rcgen/CHANGELOG.md
cpu: Fixed.

cpu:
> I think it's great to spend more words on the larger changes in the signing API, but (especially when also adding CRLs) I wonder if it could be done more concisely by explaining the general principles and abstracting as much as possible over certificate/CSR/CRL. I've tried hard to make the APIs as similar as possible for each of them.

I went back and replaced the `CHANGELOG.md` coverage with a section aiming to explain the general principles without explicitly listing the mechanical changes. I think that's a good level of detail for a CHANGELOG and fits better with the existing practices.

But... I think the Rustls releases & feedback have shown me that there's an audience for material that is much more explicit about what changed and what should be used instead. I've split off some of that content into a separate `rcgen/docs/0.12-to-0.13.md` document. It's probably overkill to do that for every release but I felt like it was worthwhile for this one. I'd like to try and make sure the process is as painless as possible for consumers.

> I like to retain the () suffix on methods/functions because it makes it more obvious to distinguish these from types/variables.

Makes sense. I've done a pass adding them in both for the CHANGELOG and migration guide document.

rcgen/CHANGELOG.md
```diff
@@ -1,6 +1,78 @@
...
+  [Tudyx](https://github.com/tudyx).
+
+- Method names starting with `get_` have been renamed to match Rust convention:
+  `CertificateRevocationList::get_params()` -> `params()`
+  `Certificate::get_params()` -> `params()`
+  `Certificate::get_key_identifier()` -> `Certificate::key_identifier()`
+  `Certificate::get_times()` -> `Certificate::times()`
+
+- `KeyPair::from_pem_and_sign_algo()` has been renamed to
+  `KeyPair::from_pkcs8_pem_and_sign_algo()` to emphasize the input DER should be
```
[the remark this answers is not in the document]
cpu: Good call. I will catch up with the state of #242.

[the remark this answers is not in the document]
cpu: Updated.

rcgen/docs/0.12-to-0.13.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Certificates
+
+* For quick-and-easy self-signed certificate issuance,
+  `generate_simple_self_signed` now returns a `CertifiedKey` in the success case
+  instead of a `Certificate`. The self-signed `Certificate` can be accessed in
+  the `cert` field of `CertifiedKey`, and the generated subject key pair in
+  `key_pair`.
+
+* Customized self-signed certificate issuance was previously done by
```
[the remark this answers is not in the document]
cpu: Fixed both.

rcgen/docs/0.12-to-0.13.md
```diff
@@ -0,0 +1,90 @@
...
+  serialization by calling `CertificateParams::signed_by()` and providing
+  a subject `KeyPair` as well as an issuer `Certificate` and `KeyPair`.
+
+* Previously certificate serialization was done by calling
+  `Certificate::serialize_der()`, `Certificate::serialize_pem()`,
+  `Certificate::serialize_der_with_signer()` or
+  `Certificate::serialize_pem_with_signer()`. Each time a serialization fn was
+  called a new certificate was issued, leading to confusion when it was desired
+  to serialize the same certificate in two formats. In the new API issuance is
+  handled by `CertificateParams` fns and a `Certificate` is stable. You can
```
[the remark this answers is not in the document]
cpu: > or explicitly say "and the generated Certificate does not change"?

I went with a small tweak on this language saying "the generated `Certificate` will not change when serialized". Good suggestion, thanks.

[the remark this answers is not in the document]
djc: It kind of still feels like a silly thing to say because the `der()` method doesn't even really do serialization and the new API just intrinsically makes it impossible for the previous pitfall to happen... not really sure how best to explain this.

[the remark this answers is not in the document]
cpu: > It kind of still feels like a silly thing to say because the der() method doesn't even really do serialization and the new API just intrinsically makes it impossible for the previous pitfall to happen.

I think the reason it feels silly is because the old API was surprising and the new behaviour seems intuitively like what most folks would expect. 

I still think the text is better explicitly saying both what the old API did and what the new API doesn't do. I suspect many readers won't have realized what the old API was doing until this moment and might want to check their own usage for bugs.

> the der() method doesn't even really do serialization

No argument there. I changed the last sentence to say "You can serialize it to PEM by calling `Certificate::pem()`, or access the DER encoding by calling `Certificate::der()`.

cpu:
@est31 I think this is ready for review now and the last thing blocking a 0.13 release. If you're comfortable approving the branch can I go ahead and publish the release afterwards? It would be great if we could do this before you're availability is reduced :-)

est31 (approved):
LGTM, let's release 0.13

djc:
* Created `v0.13.0` tag
* Published rcgen v0.13.0 at registry `crates-io`

cpu:
Thanks for finishing this up djc :rocket:

PR #255, opened 2024-03-25
lib: export csr::CertificateSigningRequest

cpu:
This type was introduced in 790a45ea0b7543fa5d4afd7f5a26377582f86e7c and is meant to be part of the public API, but the `csr` module is not exported. Therefore we have to re-export the type in `lib.rs`.

Noticed while preparing https://github.com/rustls/rcgen/pull/254


cpu:
> ci / Documentation (stable) (pull_request) Failing after 1m

Ah, need to document it now :laughing:

djc (approved):

est31 (approved):

PR #256, opened 2024-03-26
Updated x509-parser to 0.16 for 0.12.+ versions

ErikEverson:
Reason for change:
- I am working on adding tvOS support to various libraries and would like to be able to use the 0.12.+ versions of rcgen as there is a breaking change with the 0.13 release that is coming up soon.

What changed:
- I updated the x509-parser dependency to 0.16 as that version no longer has a dependency on an old version of the Ring dependency which does not compile for tvOS.

How was this tested:
- cargo test with all passing
- cargo +nightly build -Z build-std --target aarch64-apple-tvos(-sim) and x86

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,46 +13,18 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f6fd5ddaf0351dff5b8da21b2fb4ff8e08ddd02857f0bf69c47639106c0fff0"
-dependencies = [
- "asn1-rs-derive 0.4.0",
- "asn1-rs-impl 0.1.0",
- "displaydoc",
- "nom",
- "num-traits",
- "rusticata-macros",
- "thiserror",
- "time",
-]
-
-[[package]]
-name = "asn1-rs"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


cpu:
This seems reasonable-ish to me since x509-parser's types aren't part of the rcgen API.

I think you'll need to backport some of the clippy fixes from `main` to get this branch building cleanly.

> would like to be able to use the 0.12.+ versions of rcgen as there is a breaking change with the 0.13 release that is coming up soon.

If I have any hesitation it's just setting a precedent that we support older releases for some time window when I don't think any policy has been sorted out in that department. Can you speak more to the issues you expect to have adapting to the upcoming release?

cpu:
> If we can't make a release within those two weeks, then I think it's fair to merge this and push out a 0.12 release.

That sounds reasonable to me :+1: 

> (Note to @djc and @cpu: I will have low availability during the period April 6-21).

Noted! Thanks.

djc:
FWIW, I was hoping we could get 0.13 out this week, maybe as early as tomorrow -- but depends, largely, on est31's review capacity.

cpu:
[rcgen 0.13](https://crates.io/crates/rcgen/0.13.0) was published this morning, and uses x509-parser 0.16.

Without more input on why updating to 0.13 is unfeasible I think we should close this PR.

cpu:
Thanks again for your contribution (even if not merged!). If you have feedback about the 0.12 -> 0.13 upgrade process we can chat more here or in a separate issue.

cpu:
Great, thanks for following up. :+1:

PR #258, opened 2024-03-28
Fix reference in changelog about RSA key generation

djc:
Noticed by a colleague of mine.


cpu (approved):
Good catch, thank you.

est31 (approved):

PR #262, opened 2024-04-03
cert: use pre-specified SKI of issuer for AKI

cpu:
Previously when issuing a certificate with an authority key identifier (AKI) extension that's signed by an issuer certificate we had a small bug where we used the to-be-issued certificate's param's `key_identifier_method` to derive the key identifier of the issuing certificate to use for the issued certificate's AKI. Instead we should be using the issuer certificate's param's `key_identifier_method`, taking care to mind the pre-specified variant.

We missed this with our unit testing of the pre-specified key identifier method because we only issued a self-signed test certificate, never issuing a certificate signed by the CA that has the customization. We principally exercised that the SKI of the self-signed cert matched the pre-specified value, but never tested that an issued cert's AKI matches the pre-specified SKI of the issuer.

This branch fixes the bug and extends test coverage to prevent further regression.

Resolves https://github.com/rustls/rcgen/issues/261


djc (approved):
Nice work! Maybe throw in a proactive version bump?

cpu:
> Nice work! Maybe throw in a proactive version bump?

Good call. Added one and proactively updated the changelog.

I also double checked this was a regression (though I didn't bother digging up which commit was the breaker). Backporting the unit test from this branch to 0.12.1 shows it passing.

est31 (approved):
Thanks for the quick fix :heart:

djc:
* Published rcgen v0.13.1 at registry `crates-io`
* [new tag]         v0.13.1 -> v0.13.1

PR #264, opened 2024-04-07
Pass extended key usage parameters when importing CertificateSigningRequestParams 

uglyoldbob:
This pull requests adds parsing of most of the extended key usage parameters, except for Other.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -454,6 +454,21 @@ impl CertificateParams {
 		}
 		Ok(result)
 	}
+
+	fn write_extended_key_usage(&self, writer: DERWriter) {
+		if !self.extended_key_usages.is_empty() {
+			write_x509_extension(writer, oid::EXT_KEY_USAGE, false, |writer| {
+				writer.write_sequence(|writer| {
+					for usage in &self.extended_key_usages {
+						writer
+							.next()
+							.write_oid(&ObjectIdentifier::from_slice(usage.oid()));
+					}
+				});
+			});
+		}
+	}
+
 	fn write_subject_alt_names(&self, writer: DERWriter) {
 		write_x509_extension(writer, oid::SUBJECT_ALT_NAME, false, |writer| {
 			writer.write_sequence(|writer| {
@@ -516,15 +531,20 @@ impl CertificateParams {
 		} = self;
 		// - alg and key_pair will be used by the caller
 		// - not_before and not_after cannot be put in a CSR
+		// - key_identifier_method is here because self.write_extended_key_usage uses it
 		// - There might be a use case for specifying the key identifier
 		// in the CSR, but in the current API it can't be distinguished
 		// from the defaults so this is left for a later version if
 		// needed.
-		let _ = (not_before, not_after, key_identifier_method);
+		let _ = (
+			not_before,
+			not_after,
+			key_identifier_method,
+			extended_key_usages,
+		);
 		if serial_number.is_some()
 			|| *is_ca != IsCa::NoCa
 			|| !key_usages.is_empty()
-			|| !extended_key_usages.is_empty()
 			|| name_constraints.is_some()
 			|| !crl_distribution_points.is_empty()
 			|| *use_authority_key_identifier_extension
@@ -550,6 +570,7 @@ impl CertificateParams {
 							writer.next().write_sequence(|writer| {
 								// Write subject_alt_names
 								self.write_subject_alt_names(writer.next());
+								self.write_extended_key_usage(writer.next());
 
 								// Write custom extensions
 								for ext in custom_extensions {
@@ -807,6 +828,13 @@ impl CertificateParams {
 		})?;
 
 		Ok(der.into())
+	}
+
+	/// Insert an extended key usage (EKU) into the parameters if it does not already exist
+	pub fn insert_extended_key_usage(&mut self, eku: ExtendedKeyUsagePurpose) {
+		if !self.extended_key_usages.contains(&eku) {
+			self.extended_key_usages.push(eku);
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


rcgen/src/csr.rs
```diff
@@ -115,6 +115,78 @@ impl CertificateSigningRequestParams {
 								.push(SanType::try_from_general(name)?);
 						}
 					},
+					x509_parser::extensions::ParsedExtension::ExtendedKeyUsage(eku) => {
+						if eku.any {
```
djc, line 119: This is pretty verbose. Suggest adding a private method `fn insert_extended_key_usage(&mut self, eku: ExtendedKeyUsagePurpose)` to `CertificateParams` (at the bottom of the `impl`) that does the `contains()`/`push()` dance.

djc (commented):
Would be good to support this, thanks for the PR! Suggesting a more concise way of doing it.

est31 (approved):
great idea, but I share the repetition concerns. Approved if @djc 's suggestion is implemented.

est31:
Also, please squash after the fixes. I see a merge commit in the PR :).

uglyoldbob:
Looks like making a second version of the test_certificate_from_csr test or modifying that test to include the extended key usage would work. Any preference as to which one?

cpu:
> modifying that test to include the extended key usage would work. Any preference as to which one?

That sounds reasonable to me :+1:

uglyoldbob:
Support for the extended key usage probably needs to be added to CertificateParams::serialize_request.

djc (approved):

djc:
@uglyoldbob thanks, this is looking good!

@cpu, want to have a look before merging this? Already has @est31 approval.

(Let's bypass the merge queue to squash?)

cpu:
> @cpu, want to have a look before merging this?

:+1: I will take a look today

rcgen/src/csr.rs
```diff
@@ -115,6 +115,41 @@ impl CertificateSigningRequestParams {
...
+						if eku.time_stamping {
+							params.insert_extended_key_usage(
+								crate::ExtendedKeyUsagePurpose::TimeStamping,
+							);
+						}
+						if eku.ocsp_signing {
+							params.insert_extended_key_usage(
+								crate::ExtendedKeyUsagePurpose::OcspSigning,
+							);
+						}
```
cpu, line 151: I think this will silently drop instances of `x509_parser::ExtendedKeyUsage.other` holding `Vec<Oid<'a>>`.

Since there's no equivalent for an arbitrary EKU in rcgen I think it would be prudent to return `Err(Error::UnsupportedExtension)` if `!eku.other.is_empty()`

rcgen/tests/webpki.rs
```diff
@@ -491,15 +491,38 @@ fn test_certificate_from_csr() {
...
 		.distinguished_name
 		.push(DnType::CommonName, "Dev domain");
+
 	let cert_key = KeyPair::generate().unwrap();
 	let csr = params.serialize_request(&cert_key).unwrap();
 	let csr = CertificateSigningRequestParams::from_der(csr.der()).unwrap();
 
 	let (mut params, ca_key) = util::default_params();
 	params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+	let eku_test = vec![
```
cpu, line 501: I'm a bit confused by this test. We put EKUs into parameters, check the expected EKUs are in the parameters, create a self-signed CA certificate from the parameters, and then verify the CA certificate's parameters contain the EKUs.

It doesn't seem like anything is testing that a CSR that has EKUs is loaded with those EKUs, or that a CSR created from params with EKUs preserve the EKUs. I think those are the important properties we want to test right?

uglyoldbob: I was a bit stumped as how to fully write a test here since some of the testing I had thought of required for  CertificateParams::serialize_request to implement the extended key usage, but thought adding that might be better for a different pull request.

rcgen/src/csr.rs
uglyoldbob: Oh that makes sense

rcgen/tests/webpki.rs
cpu: > adding that might be better for a different pull request.

I would be OK with it being part of this PR!

rcgen/Cargo.toml
```diff
@@ -36,7 +36,7 @@ x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
-default = ["crypto", "pem", "ring"]
+default = ["crypto", "pem", "ring", "x509-parser"]
```
djc, line 39: I don't think this what we want? It looks like this might be only for (some of) the tests? If so, we can make x509-parser a required dev-dependency.

rcgen/src/certificate.rs
```diff
@@ -794,7 +814,7 @@ impl CertificateParams {
 		Ok(der.into())
 	}
 
-	/// Insert an extended key usage into the parameters if it does not already exist
+	/// Insert an extended key usage (EKU) into the parameters if it does not already exist
```
djc, line 817: Please move this change into the earlier commit.

rcgen/Cargo.toml
uglyoldbob: Oh I think I accidentally commited that part. It was so my ide would actually be helpful for the section I was working on.

djc: I suggest running your IDE with --all-features (that's what I do).

cpu (approved):
Thanks! Had a couple small nits to mention, but I don't think they're blockers if you feel strongly.

cpu (approved):
Thanks! Would you mind squashing the two commits into one? It would save me having to admin merge this to do the squashing myself.

PR #267, opened 2024-04-08
Split KeyPairAlgorithm, make non-CLI part public

### LINKED ISSUE #266: rustls-cert-gen: `KeyPairAlgorithm` should be publicly exported
I tried to use `rustls-cert-gen` as a library and ran into an issue. Both `CaBuilder` and `EndEntityBuilder` require a `KeyPairAlgorithm` argument, but this is not publicly exported out of the `cert` module. I see that both `main.rs` and `lib.rs` declare `mod cert;`, so that's how the command line tool happens to avoid this restriction.

divergentdave:
This addresses #266 by splitting `KeyPairAlgorithm` into two enums, a public one that is now re-exported from the library crate, and one that we use the `Bpaf` derive macro on. Conversions are defined between the two. The binary crate now gets `KeyPairAlgorithm` and `CertificateBuilder` from the library crate, rather than by including `cert.rs`.

Closes #266.

### DIFF
--- a/rustls-cert-gen/src/cert.rs
+++ b/rustls-cert-gen/src/cert.rs
@@ -227,11 +227,11 @@ pub enum KeyPairAlgorithm {
 	EcdsaP256,
 	/// ECDSA with the P-384 curve
 	///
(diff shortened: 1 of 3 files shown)


djc:
(See comments in #266.)

PR #269, opened 2024-04-18
Clarify internal data dependencies for signing APIs

djc:
Without changing the public API, prepare for splitting out a type that is capable of signing.


cpu (approved):
Looks good :-)

djc:
@est31 any opinions?

est31:
Could you talk more about the end goal? I'm not sure I like this refactor from its own merit: the type doesn't pull its weight IMO.

est31:
Oh to prevent misuse of that API as some kind of "certificate editing" facility? I see, that makes sense. Approving.

est31 (approved):

PR #272, opened 2024-05-08
Expose PKCS_RSA_PSS_SHA256 for CSR generation

tgonzalezorlandoarm:
Make PKCS_RSA_PSS_SHA256 a publicly accessible algorithm so that CSRs can be created for RSA PSS.

This has been tested with [https://github.com/parallaxsecond/parsec](parsec) on this [parsec-tool PR with the patched rcgen crate](https://github.com/parallaxsecond/parsec-tool/pull/128). Please check the results of openssl CSR verification in the [CI run](https://github.com/parallaxsecond/parsec-tool/actions/runs/8989184593/job/24691715875#step:4:1116)

For this to work with Parsec, this needs to be backported to v0.9.x
The tests are not enable as parsec is not used during testing.


rcgen/src/sign_algo.rs
```diff
@@ -145,24 +145,22 @@ pub(crate) mod algo {
 		params: SignatureAlgorithmParams::Null,
 	};
 
-	// TODO: not really sure whether the certs we generate actually work.
-	// Both openssl and webpki reject them. It *might* be possible that openssl
-	// accepts the certificate if the key is a proper RSA-PSS key, but ring doesn't
-	// support those: https://github.com/briansmith/ring/issues/1353
+	// openssl accepts the certificate if the key is a proper RSA-PSS key, but
```
tgonzalezorlandoarm, line 148: Happy to change this comment to whatever suits the project better :)

tgonzalezorlandoarm:
@cpu @djc this is a Draft PR just because it needs some discussion first

djc:
Why doesn't the parsec tool update to rcgen 0.13? I don't know that we'd want to backport to 0.9.

cpu:
Just leaving a note that I'll be travelling for the next week and probably won't have a chance to review this before then.

tgonzalezorlandoarm:
I understand why you wouldn't want to backport, we'll have the parsec-tool update to rcgen 0.13.x.

rcgen/src/sign_algo.rs
```diff
@@ -145,19 +145,17 @@ pub(crate) mod algo {
...
+	// ring doesn't support those: https://github.com/briansmith/ring/issues/1353
```
cpu, line 149: This comment seems good as-is, but I think it would be prudent to rejig this so that the mention of there being no support in `*ring*` is part of the public rustdoc comment that makes it into documentation. Maybe something like:

```rust
	/// RSA signing with PKCS#1 2.1 RSASSA-PSS padding and SHA-256 hashing as per [RFC 4055](https://tools.ietf.org/html/rfc4055)
    ///
	/// Note: `*ring*` does not support this signature algorithm, and so it can not be used with the `crypto` feature
	/// of `rcgen` when verifying signatures using the `ring` backend.
	pub static PKCS_RSA_PSS_SHA256: SignatureAlgorithm = SignatureAlgorithm {
```

It looks like `aws-lc-rs` _does_ support these signature algorithms.

rcgen/src/sign_algo.rs
```diff
@@ -145,19 +145,17 @@ pub(crate) mod algo {
...
 	//
 	/// RSA signing with PKCS#1 2.1 RSASSA-PSS padding and SHA-256 hashing as per [RFC 4055](https://tools.ietf.org/html/rfc4055)
-	pub(crate) static PKCS_RSA_PSS_SHA256: SignatureAlgorithm = SignatureAlgorithm {
+	pub static PKCS_RSA_PSS_SHA256: SignatureAlgorithm = SignatureAlgorithm {
```
cpu, line 152: Should we add the `_SHA384` and `_SHA512` variants as well for parity with `PKCS_RSA_XXX`?

rcgen/src/sign_algo.rs
```diff
@@ -160,7 +160,7 @@ pub(crate) mod algo {
 		params: SignatureAlgorithmParams::RsaPss {
 			// id-sha256 in https://datatracker.ietf.org/doc/html/rfc4055#section-2.1
 			hash_algorithm: &[2, 16, 840, 1, 101, 3, 4, 2, 1],
-			salt_length: 20,
+			salt_length: 32,
```
cpu, line 163: I think it might merit a comment here that's its conventional to use a salt length equal to the size of the hash algorithm's digest (in this case, 32 bytes for the 256 bit digest produced by SHA256). 

I don't think this is a _requirement_ with any backing in specification text so it seems overly aggressive for OpenSSL to be rejecting other salt lengths, but I also don't see any reason to avoid doing what seems to be most conventional.

cpu (commented):
Generally looks good to me. I think it would be OK to squash your two commits into one, and to move this out of draft status.

To be appropriate to merge it would also be nice to have test coverage. I would expect that it's possible to write a test using `openssl` in `rcgen/tests/openssl.rs`, and with some feature gating, with `aws-lc-rs` as the backing provider for `rcgen`'s `crypto` feature. If you're especially keen it would be interesting to see if `botan-rs` could support this as well.

rcgen/src/sign_algo.rs
tgonzalezorlandoarm: Done

tgonzalezorlandoarm:
Hi!

I have:

1. Squashed the commits as @cpu requested.
2. Added a commit for taking into account the rest of the variants (SHA384 and SHA512).
3. Tried adding a test with `aws-lc-rs`. Unfortunately, as shown in the CI, [this fails with an "unable to get certs public key" error](https://github.com/rustls/rcgen/actions/runs/9350682411/job/25734671696?pr=272#step:8:507) Taking a look in the aws-lc-rs tests, the [generated signature value is not being tested](https://github.com/aws/aws-lc-rs/blob/ebe9b6ea8b5ffb0859c8c9c601c4653ae1db693f/aws-lc-rs/tests/rsa_test.rs#L129), and taking a look at their code, they are [setting the salt length following the convention](https://github.com/aws/aws-lc-rs/blob/ebe9b6ea8b5ffb0859c8c9c601c4653ae1db693f/aws-lc-rs/src/signature.rs#L89) mentioned by @cpu in this PR (same as we are doing in here).

I apologize for the delay, I have been busy with the recent release of our parsec openssl provider, with which we are able to perform a TLS handshake, using a CSR created with rcgen (with a ParsecRemoteKeyPair) for an RSA key with PSS signing.

rcgen/src/sign_algo.rs
tgonzalezorlandoarm: Done

rcgen/src/sign_algo.rs
tgonzalezorlandoarm: The certificate being generated is failing here

cpu: Have you considered making a minimal reproducer and opening an issue with aws-lc-rs? I've had good experience interacting with the maintainers.

PR #273, opened 2024-05-10
fix missing export of `KeyPairAlgorithm`

roberts-pumpurs:
This was preventing me from using `rustls-cert-gen` as a library.
Also re-export `rcgen` crate to access shared types.

### DIFF
--- a/rustls-cert-gen/src/lib.rs
+++ b/rustls-cert-gen/src/lib.rs
@@ -5,4 +5,5 @@
 //! whatever purpose you may need a TLS certificate-chain.
 
 mod cert;
-pub use cert::{Ca, CaBuilder, CertificateBuilder, EndEntity, EndEntityBuilder};
+pub use cert::{Ca, CaBuilder, CertificateBuilder, EndEntity, EndEntityBuilder, KeyPairAlgorithm};
+pub use rcgen;


djc:
rustls-cert-gen is not intended for use as a library. You should use rcgen instead.

cpu:
I'm going to close this for now based on our preference to discourage library use of the CLI crate.

> rustls-cert-gen is not intended for use as a library.

It looks like we fell into a weird in-between here where the CLI tool is structured around a `lib.rs`, and _some_ elements are `pub`, yet we recommend not using it as a library.

I think that recommendation is correct (_it's a shame to drag in the extra deps like `bpaf` and `anyhow` for a library use-case_) but perhaps we should take an action to make it seem less appealing?

@roberts-pumpurs Can you speak to whether there's something missing from `rcgen`-proper that made you reach for `rustls-cert-gen` instead?

roberts-pumpurs:
No, `rcgen` is fine. I just liked the builder pattern on the CLI — nothing else, really. And given that it was published on crates.io and had a lib.rs that (at first glance) exported everything that I'd need, it made me want to reach for that rather than the more "low-level" `regen` crate.

cpu:
Makes sense, I'm also a fan of the builder pattern :+1: Thanks for the reply.

Issue #274, opened 2024-05-17
Add `Certificate::from_der` + `Certificate::from_pem`


oscartbeaumont:
I'm in the process of upgrading rcgen and I have to say the new API is much easier to understand, great job!

In reading some of the #62 I can see the idea of adding `Certificate::from_der` and `Certificate::from_pem` came up but they don't seem to have made it into the final release.

I wanted to know if there is a way to load an existing CA certificate to sign some more certificates without triggering signing.

Right now i'm doing the following:
```rs
let identity_key = todo!();
let cert: Vec<u8> = todo!();

let cert = CertificateParams::from_ca_cert_der(&cert.try_into().unwrap())
    .unwrap()
    .self_signed(&identity_key)
    .unwrap();
```

However, my presumption is that this will sign a new CA certificate on each startup of my application and I would rather just import the existing CA certificate and use it.

I tried to do a PR to add the `Certificate::from_der` + `Certificate::from_pem` but i'm kinda lost where the `subject_public_key_info` would come from. I've included what I had below:
```rs
impl Certificate {
	/// TODO
	pub fn from_der(der: CertificateDer<'static>) -> Result<Self, Error> {
		let params = CertificateParams::from_ca_cert_der(&der)?;
		Ok(Self {
			params,
			subject_public_key_info: todo!(),
			der,
		})
	}
}
```

djc:
Yeah, this API isn't great right now. I've described a plan to improve it in https://github.com/rustls/rcgen/pull/269#issuecomment-2079282218. Unfortunately with the current API I don't think there's a way around doing a fake self-signing of the CA certificate before using it to sign dependent certificates.

acul009:
This would be great to have. It's the reason I can't upgrade to 0.13 right now.

I only notized after converting most of my code, which is unfortunate, but I had the opportunity to see the new 0.13 in action.

I actually really like the new params/signing API - it's a lot nicer to use than the manual `serialize_der_with_signer`
A few of the types are also pretty close to the wrappers I wrote. For the Keypair I was actually able to remove large parts of my own logic.

djc:
> I came across this trying to write a certificate signing library. From what I'm seeing from testing, the certificate generated from the "fake self-signing" workaround seems to be semantically equivalent to the original, with the only differences (as far as I can tell) being the ordering of extensions and the lack of AKI auto-detection. Are there any caveats or footguns I should watch out for when signing certificates using this workaround?

I think that's about it.

PR #275, opened 2024-05-21
`Issuer` API

oscartbeaumont:
Motivation #274, based off https://github.com/rustls/rcgen/pull/269#issuecomment-2079282218

Still required before merge:
 - [ ] Finalise the API requirements
 - [ ] Implement it all
 - [ ] Documentation comments
 - [ ] Changelog - This will require a major release!

### `rustls_cert_gen::Ca`?

We could either keep it as a wrapper type of `rcgen::Issuer` or phase it out completely.

If we were gonna phase it out I would do it with the following changes:
 - [ ] Drop `rustls_cert_gen::Ca` in favor of `rcgen::Issuer`
 - [ ] Add a conversion from `rcgen::Issuer` to `rcgen::CertifiedKey` using `From` or `to_certified_key` method.
 - [ ] Move `rcgen::PemCertifiedKey::write` to `rcgen::CertifiedKey::write_pem` (replacing `Ca::serialize_pem().write` with `Issuer::to_certified_key().write_pem()`, alternatively we could also just have `Issuer::write_pem`)
 - [ ] Drop `rustls_cert_gen::PemCertifiedKey` in favor of `rcgen::CertifiedKey` (which is fine because the method is now `write_pem` not `write`)
 - [ ] Remove `rustls_cert_gen::EndEntity::serialize_pem` and provide conversion from `rustls_cert_gen::EndEntity` to `rcgen::CertifiedKey` using `From` or `to_certified_key` method.
 - [ ] Should `rustls_cert_gen::EndEntity` just be `rcgen::CertifiedKey` or is that too far?

### `KeyPair` and `Clone`?

I have removed the lifetime from `Issuer` so it holds owned values, as I personally think this is the best call and it was implied by @djc's message linked above.

This does raise an interesting concern, how should `CertificateParams::self_signed` construct an `rcgen::Issuer` as it takes in `&KeyPair` and we would require an owned `KeyPair`.

Options:
 - [ ] Implement `Clone` publically for `KeyPair` and make `CertificateParams::self_signed` take `KeyPair`
 - [ ] Implement `Clone` privately for `KeyPair` and use it
 - [ ] Make `Issuer` hold a `Cow<'a, KeyPair>` and provide an `Issuer::to_owned` method which allocates the data and returns `Issuer<'static>`.
 - [ ] Make `Issuer` hold a lifetime to the `KeyPair`

Option 4 has the major downside that it makes parsing around an `Issuer` really difficult due to it always having a lifetime to some buffer. I know in my personal use of `rcgen` I load the issuer on startup and give it around to Tokio tasks, and it would be nice to avoid leaking memory as would be required with this option.

### `rcgen::Issuer` vs `rcgen::CertifiedKey`?

Aren't these basically the same type.

`rustls_cert_gen::Ca` being either a wrapper or replaced by `rcgen::Issuer` means we need to provide a way to go from an `Issuer` to `Certificate` which we will use to replace/reimplement the existing `rustls_cert_gen::Ca::cert` method.

Given this, I think it seems logical that in code `Issuer` could be:
```rs
pub struct CertifiedKey {
    pub cert: Certificate,
    pub key_pair: KeyPair,
}
```

which is `rcgen::CertifiedKey`.

I could be missing something but if they are the same type I feel like unifying them into one makes sense and simplifies the public API.

### TODO

> The from_ca_cert_*() constructors from CertificateParams move to Issuer, and we see if we can address the documented caveats on those methods.

Gotta do this as @djc noted in the linked message.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -160,19 +160,7 @@ impl CertificateParams {
 	///
 	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
 	/// [`Certificate::pem`].
-	pub fn signed_by(
-		self,
-		key_pair: &KeyPair,
-		issuer: &Certificate,
-		issuer_key: &KeyPair,
-	) -> Result<Certificate, Error> {
-		let issuer = Issuer {
-			distinguished_name: &issuer.params.distinguished_name,
-			key_identifier_method: &issuer.params.key_identifier_method,
-			key_usages: &issuer.params.key_usages,
-			key_pair: issuer_key,
-		};
-
+	pub fn signed_by(self, key_pair: &KeyPair, issuer: &Issuer) -> Result<Certificate, Error> {
 		let subject_public_key_info = key_pair.public_key_der();
 		let der = self.serialize_der_with_signer(key_pair, issuer)?;
 		Ok(Certificate {
@@ -188,14 +176,14 @@ impl CertificateParams {
 	/// [`Certificate::pem`].
 	pub fn self_signed(self, key_pair: &KeyPair) -> Result<Certificate, Error> {
 		let issuer = Issuer {
-			distinguished_name: &self.distinguished_name,
-			key_identifier_method: &self.key_identifier_method,
-			key_usages: &self.key_usages,
-			key_pair,
+			distinguished_name: self.distinguished_name,
+			key_identifier_method: self.key_identifier_method,
+			key_usages: self.key_usages,
+			key_pair: todo!(), // TODO: Do we `KeyPair::Clone` or do we make this field a `Cow`?
 		};
 
 		let subject_public_key_info = key_pair.public_key_der();
-		let der = self.serialize_der_with_signer(key_pair, issuer)?;
+		let der = self.serialize_der_with_signer(key_pair, &issuer)?;
 		Ok(Certificate {
 			params: self,
 			subject_public_key_info,
@@ -610,7 +598,7 @@ impl CertificateParams {
 	pub(crate) fn serialize_der_with_signer<K: PublicKeyData>(
 		&self,
 		pub_key: &K,
-		issuer: Issuer<'_>,
+		issuer: &Issuer,
 	) -> Result<CertificateDer<'static>, Error> {
 		let der = issuer.key_pair.sign_der(|writer| {
 			let pub_key_spki =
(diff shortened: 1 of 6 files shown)


djc:
> ## Questions
> ### `rustls_cert_gen::Ca`?
> 
> We could either keep it as a wrapper type of `rcgen::Issuer` or phase it out completely.
> 
> If we were gonna phase it out I would do it with the following changes:
> 
>     * [ ]  Drop `rustls_cert_gen::Ca` in favor of `rcgen::Issuer`
> 
>     * [ ]  Add a conversion from `rcgen::Issuer` to `rcgen::CertifiedKey` using `From` or `to_certified_key` method.
> 
>     * [ ]  Move `rcgen::PemCertifiedKey::write` to `rcgen::CertifiedKey::write_pem` (replacing `Ca::serialize_pem().write` with `Issuer::to_certified_key().write_pem()`, alternatively we could also just have `Issuer::write_pem`)
> 
>     * [ ]  Drop `rustls_cert_gen::PemCertifiedKey` in favor of `rcgen::CertifiedKey` (which is fine because the method is now `write_pem` not `write`)
> 
>     * [ ]  Remove `rustls_cert_gen::EndEntity::serialize_pem` and provide conversion from `rustls_cert_gen::EndEntity` to `rcgen::CertifiedKey` using `From` or `to_certified_key` method.
> 
>     * [ ]  Should `rustls_cert_gen::EndEntity` just be `rcgen::CertifiedKey` or is that too far?

I generally think we should focus on getting the rcgen API first before we think much about the rustls-cert-gen API -- in the end for rustls-cert-gen, it's really only the CLI that matters. That said, it probably makes sense to replace `Ca` with `Issuer`.

> ### `KeyPair` and `Clone`?
> 
> I have removed the lifetime from `Issuer` so it holds owned values, as I personally think this is the best call and it was implied by @djc's message linked above.
> 
> This does raise an interesting concern, how should `CertificateParams::self_signed` construct an `rcgen::Issuer` as it takes in `&KeyPair` and we would require an owned `KeyPair`.
> 
> Options:
> 
>     * [ ]  Implement `Clone` publically for `KeyPair` and make `CertificateParams::self_signed` take `KeyPair`
> 
>     * [ ]  Implement `Clone` privately for `KeyPair` and use it
> 
>     * [ ]  Make `Issuer` hold a `Cow<'a, KeyPair>` and provide an `Issuer::to_owned` method which allocates the data and returns `Issuer<'static>`.
> 
>     * [ ]  Make `Issuer` hold a lifetime to the `KeyPair`

I think we should start to have a type system-level difference between the notion of self-signed certificates in general and CA-level certificates (which have some extra metadata). As such, I think perhaps `CertificateParams::self_signed()` is the exception and it should not take an `Issuer`, instead only taking the `&KeyPair` it takes today.

> ### TODO
> 
> > The from_ca_cert_*() constructors from CertificateParams move to Issuer, and we see if we can address the documented caveats on those methods.
> 
> Gotta do this as @djc noted in the linked message.

Moving the `from_ca_cert_*()` constructors is important. I also think we'll want to stop storing the `params` in `Certificate` and take `&self` as a reference in `signed_by()` methods.

Issue #280, opened 2024-06-27
Expose `alg` on PublicKey


rickvanprim:
Presently the `PublicKey` field exposed in `CertificateSigningRequestParams` doesn't allow you to do anything with it aside from comparing it to another instance of `PublicKey`.  It would be nice to be able to check which algorithm was used to generate the `PublicKey`.

In my use case, I control both the client side generating the CSR and the server side signing it, so I think it makes sense for the server to check that the algorithm used in the CSR's `PublicKey` matches what is expected.

PR #281, opened 2024-06-27
Expose `algorithm` field on `PublicKey`

rickvanprim:
Fixes https://github.com/rustls/rcgen/issues/280.

### DIFF
--- a/rcgen/src/csr.rs
+++ b/rcgen/src/csr.rs
@@ -19,6 +19,13 @@ pub struct PublicKey {
 	alg: &'static SignatureAlgorithm,
 }
 
+impl PublicKey {
+	/// The algorithm used to generate the public key and sign the CSR.
+	pub fn algorithm(&self) -> &SignatureAlgorithm {
+		self.alg
+	}
+}
+
 impl PublicKeyData for PublicKey {
 	fn alg(&self) -> &SignatureAlgorithm {
 		self.alg
--- a/rcgen/tests/generic.rs
+++ b/rcgen/tests/generic.rs
@@ -364,13 +364,15 @@ mod test_csr {
 
 	#[test]
 	fn test_csr_roundtrip() {
+		let key_pair = KeyPair::generate().unwrap();
+
 		// We should be able to serialize a CSR, and then parse the CSR.
-		_ = CertificateSigningRequestParams::from_der(
-			CertificateParams::default()
-				.serialize_request(&KeyPair::generate().unwrap())
-				.unwrap()
-				.der(),
-		)
-		.unwrap();
+		let csr = CertificateParams::default()
+			.serialize_request(&key_pair)
+			.unwrap();
+		let csrp = CertificateSigningRequestParams::from_der(csr.der()).unwrap();
+
+		// Ensure algorithms match.
+		assert_eq!(key_pair.algorithm(), csrp.public_key.algorithm());
 	}
 }


rcgen/src/csr.rs
```diff
@@ -16,7 +16,8 @@ use crate::{DistinguishedName, SanType};
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct PublicKey {
 	raw: Vec<u8>,
-	alg: &'static SignatureAlgorithm,
+	/// The algorithm used to generate the public key.
```
cpu, line 19:
```suggestion
	/// The algorithm used to generate the public key and sign the CSR.
```

cpu (approved):
This seems reasonable to me. AIUI this is a capability callers had before  https://github.com/rustls/rcgen/pull/205

rcgen/src/csr.rs
rickvanprim: Done

cpu: Thanks!

rickvanprim:
> This seems reasonable to me. AIUI this is a capability callers had before #205

My code on the older version of `rcgen` did do the algorithm check as it was exposed as part of `CertificateParams` 🙂

djc (approved):

est31 (changes requested):
I think it would be better to add it as a function on `PublicKey` instead, then `mut` access doesn't allow you to generate invalid configurations.

Also, I wonder whether it makes sense to additionally expose the `PublicKeyData` trait to the public (I don't like APIs that require you to import a trait, ideally the functions are available both through traits and inherent functions, unless it is a lot of API surface that's covered). Maybe that's a separate discussion though.

cpu:
> I think it would be better to add it as a function on PublicKey instead, then mut access doesn't allow you to generate invalid configurations.

Sounds reasonable to me.

> Also, I wonder whether it makes sense to additionally expose the PublicKeyData trait to the public

I like this direction less personally. I think this trait is a bit awkward (but have a hard time articulating why). I think exposing it in the existing API will make it harder to smooth out any rough edges.

rickvanprim:
@est31 / @cpu, I changed to a method on `PublicKey`, which is a copy paste of the method for `PublicKeyData`.  I could make the `PublicKeyData` method call the method on `PublicKey` as well if that's preferred, but given how they're both trivial accessors, I'm not sure that serves much point.  Please let me know if any further changes are desired 🙂

One small discrepancy I just noticed while adding this to `test_csr_roundtrip()` is `KeyPair::algorithm()` vs `PublicKeyData::alg()`.

est31 (approved):
Thanks!

rcgen/src/csr.rs
```diff
@@ -19,6 +19,13 @@ pub struct PublicKey {
 	alg: &'static SignatureAlgorithm,
 }
 
+impl PublicKey {
+	/// The algorithm used to generate the public key and sign the CSR.
+	pub fn alg(&self) -> &SignatureAlgorithm {
```
cpu, line 24: For a new fn where we have some flexibility I think `algorithm` would be a better name and will resolve the mismatch you called out in your unit test w.r.t the `assert_eq!`. Let's do that?

rickvanprim: Done 🙂

cpu (approved):
Thanks!

PR #284, opened 2024-07-06
crl: avoid markdown footnotes

cpu:
Nightly rustdoc flags these as non-standard markdown, and they don't render as nicely as just using a link. See also https://github.com/rustls/rustls/pull/2033, https://github.com/rustls/webpki/pull/268

Fixes [build errors on main](https://github.com/rustls/rcgen/actions/runs/9821174655/job/27116566286).


djc (approved):

cpu:
Since this is docs only I'm going to merge with one positive review. Happy to iterate if there's any additional feedback.

PR #286, opened 2024-07-30
Disable default features for `aws-lc-rs`

daxpedda:
Rcgen doesn't require `aws-lc-rs`'s default features, so disabling them removes some dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["rcgen", "rustls-cert-gen"]
 resolver = "2"
 
 [workspace.dependencies]
-aws-lc-rs = "1.6.0"
+aws-lc-rs = { version = "1.6.0", default-features = false }
 pem = "3.0.2"
 pki-types = { package = "rustls-pki-types", version = "1.4.1" }
 rand = "0.8"
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -38,9 +38,9 @@ zeroize = { version = "1.2", optional = true }
 [features]
 default = ["crypto", "pem", "ring"]
 crypto = []
-aws_lc_rs = ["crypto", "dep:aws-lc-rs"]
+aws_lc_rs = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys"]
 ring = ["crypto", "dep:ring"]
-fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
+fips = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/fips"]
 
 [package.metadata.docs.rs]
 features = ["x509-parser"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,7 +107,6 @@ dependencies = [
  "aws-lc-sys",
  "mirai-annotations",
  "paste",
- "untrusted 0.7.1",
  "zeroize",
 ]
 
@@ -529,9 +528,9 @@ checksum = "49f1f14873335454500d59611f1cf4a4b0f786f9ac11f4312a78e4cf2566695b"
 
 [[package]]
 name = "jobserver"
-version = "0.1.31"
+version = "0.1.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2b099aaa34a9751c5bf0878add70444e1ed2dd73f347be99003d4577277de6e"
+checksum = "48d1dbcbbeb6a7fec7e059840aa538bd62aaccf972c7346c4d9d2059312853d0"
 dependencies = [
  "libc",
 ]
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -3,7 +3,7 @@ members = ["rcgen", "rustls-cert-gen"]
 resolver = "2"
 
 [workspace.dependencies]
-aws-lc-rs = "1.6.0"
+aws-lc-rs = { version = "1.6.0", default-features = false }
```
cpu, line 6: I think this wants `features = ["aws-lc-sys"]`, like in [rustls](https://github.com/rustls/rustls/blob/2615a09d6d3e8ecff841306ed43069071369d1fe/rustls/Cargo.toml#L20C77-L20C103), to fix the CI failure.

cpu (approved):
LGTM mod fixing the build. Thank you!

rcgen/Cargo.toml
```diff
@@ -38,9 +38,9 @@ zeroize = { version = "1.2", optional = true }
 [features]
 default = ["crypto", "pem", "ring"]
 crypto = []
-aws_lc_rs = ["crypto", "dep:aws-lc-rs"]
+aws_lc_rs = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys"]
 ring = ["crypto", "dep:ring"]
-fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
+fips = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/fips"]
```
daxpedda, line 43: So I solved it a bit differently, otherwise `fips` would pull in `aws-lc-sys` as well, let me know if this is fine.

rcgen/Cargo.toml
```diff
@@ -38,9 +38,9 @@ zeroize = { version = "1.2", optional = true }
 [features]
 default = ["crypto", "pem", "ring"]
 crypto = []
-aws_lc_rs = ["crypto", "dep:aws-lc-rs"]
+aws_lc_rs = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys"]
 ring = ["crypto", "dep:ring"]
-fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
```
daxpedda, line 43: I removed the `?` because it didn't make sense, or was it there for a reason?

Cargo.toml
daxpedda: Continuation in https://github.com/rustls/rcgen/pull/286#discussion_r1697580115.

rcgen/Cargo.toml
cpu: Makes sense, I forgot there was FIPS support wired up here :+1:

rcgen/Cargo.toml
cpu: I think you're right that in this context it doesn't make sense and should be removed. My understanding of the `?` is that it:
> "will only enable the given feature if something else enables the optional dependency."

But in this case, the `fips` feature is activating `aws_lc_rs` right alongside `aws-lc-rs/fips`. 

@djc Does that match your understanding? The usage came along with https://github.com/rustls/rcgen/pull/234 and wasn't discussed explicitly as far as I can tell.

djc: IIRC there was a bug in older Cargo versions that needed the ? to be there. If it works without it on the project's MSRV, I'm fine with this.

djc:
Seems surprising that support for -sys is optional in aws-lc-rs?

rcgen/Cargo.toml
daxpedda: I tried it locally with Rust v1.66 and it compiles for me.

daxpedda:
> Seems surprising that support for -sys is optional in aws-lc-rs?

AFAIU the choice is between `aws-lc-sys` or `aws-lc-fips-sys` (exposed as the crate feature `fips`).
Compilation will fail if you don't choose at least one.

est31 (approved):

rcgen/Cargo.toml
est31: `"aws_lc_rs", "aws-lc-rs?/fips"` is the new way of writing `"aws-lc-rs/fips"` and makes it explicit that the enabling the `aws-lc-rs` feature is vital for the `fips` feature.

djc: In this case I don't think it makes sense to preserve the `?`, since we unconditionally enable the optional dependency anyway when the `fips` feature is enabled.

cpu: > In this case I don't think it makes sense to preserve the ?

I agree. Since @est31 didn't request any changes and approved the PR I think it's fair to merge with the removal?

cpu (approved):

Issue #285, opened 2024-07-25
Expand CSR generation support


lvkv:
Hi! I'm looking for an alternative to rust-openssl's [`X509ReqBuilder`](https://docs.rs/openssl/latest/openssl/x509/struct.X509ReqBuilder.html) for building and signing certificate signing requests. Ideally an alternative that supports:

- Custom EKUs (supported by OpenSSL)
- Custom certificate extensions (supported by OpenSSL)
- Custom ASN.1/DER-encoded CSR attributes in the PKCS#9/10 sense (not supported in any _sane_ way by OpenSSL AFAIK, but this would be amazing)

I've had a great experience with other Rustls projects, so I checked here first and found `rcgen`. Of course, I understand if a CSR builder is out of scope here.

Motivation: I'm implementing a subset of the MS-WCCE certificate enrollment protocol, which involves crafting CSRs with many interesting extensions and attributes. I'm also open to suggestions for other crates that might be able to help out with this!

djc:
@lvkv we already have support for building CSRs, which currently works by building from a `CertificateParams`. Would be happy to review PRs that add the stuff you need to `CertificateParams`. Rather than supporting a random collection of custom stuff, I think we should probably go the way of defining some API that fits with the MS-WCCE standard.

lvkv:
> we already have support for building CSRs, which currently works by building from a CertificateParams. 

Oh interesting! I didn't catch that the way to create a `CertificateSigningRequest` is to first create a `CertificateParams` and then to `serialize_request`. FYI this feels non-obvious, especially given that `CertificateSigningRequestParams` also exists, but isn't used for that purpose.

I tried creating a dummy CSR using this method, but ran into an `UnsupportedInCsr` error because I tried to add key usages to the request. Looks like EKU support was recently merged, so I assume this is all a work in progress still?

https://github.com/rustls/rcgen/blob/746da00ef695873b322823ae9ff7a1a3151578ae/rcgen/src/certificate.rs#L545-L553

> Given the scope here, it probably makes sense to make an exhaustive list of the exact bits you need and come up with a tentative API design that doesn't just expose a collection of customizations (extensions, EKUs and attributes) but tries to define comprehensive support for what MS-WCCE needs.

[...]

PR #287, opened 2024-08-07
Add KeyUsage support to CSR generation

lvkv:
I've added KeyUsage support to CSR generation, as well as a number of improvements to the parsing and writing of DER-encoded key usages.

These commits can be reviewed in order!

(Coming from https://github.com/rustls/rcgen/issues/285)

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -324,38 +324,9 @@ impl CertificateParams {
 			.key_usage()
 			.or(Err(Error::CouldNotParseCertificate))?
 			.map(|ext| ext.value);
-
-		let mut key_usages = Vec::new();
-		if let Some(key_usage) = key_usage {
-			if key_usage.digital_signature() {
-				key_usages.push(KeyUsagePurpose::DigitalSignature);
-			}
-			if key_usage.non_repudiation() {
-				key_usages.push(KeyUsagePurpose::ContentCommitment);
-			}
-			if key_usage.key_encipherment() {
-				key_usages.push(KeyUsagePurpose::KeyEncipherment);
-			}
-			if key_usage.data_encipherment() {
-				key_usages.push(KeyUsagePurpose::DataEncipherment);
-			}
-			if key_usage.key_agreement() {
-				key_usages.push(KeyUsagePurpose::KeyAgreement);
-			}
-			if key_usage.key_cert_sign() {
-				key_usages.push(KeyUsagePurpose::KeyCertSign);
-			}
-			if key_usage.crl_sign() {
-				key_usages.push(KeyUsagePurpose::CrlSign);
-			}
-			if key_usage.encipher_only() {
-				key_usages.push(KeyUsagePurpose::EncipherOnly);
-			}
-			if key_usage.decipher_only() {
-				key_usages.push(KeyUsagePurpose::DecipherOnly);
-			}
-		}
-		Ok(key_usages)
+		// This x509 parser stores flags in reversed bit BIT STRING order
+		let flags = key_usage.map_or(0u16, |k| k.flags).reverse_bits();
+		Ok(KeyUsagePurpose::from_u16(flags))
 	}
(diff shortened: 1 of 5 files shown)


lvkv:
I took a stab at simplifying the existing KeyUsage serialization logic—AFAICT, there seems to be no downside (aside from an extra byte if you don't use decipherOnly) to unconditionally encoding KeyUsage with 9 bits. I also took the opportunity to use the same logic for both certificates as well as CSRs.

lvkv:
I'll also add tests for this...

rcgen/src/lib.rs
```diff
@@ -435,6 +435,25 @@ pub enum KeyUsagePurpose {
 	DecipherOnly,
 }
 
+/// Encode a key usage as the value of a BIT STRING as defined by RFC 5280.
+/// [`u16`] is sufficient to encode the largest possible key usage value (two bytes).
+impl From<&KeyUsagePurpose> for u16 {
```
djc, line 440: A disadvantage of using a `From` impl for this is that it instantly becomes public. As such and since we have no motivation (as far as I can tell) to expose this as public API, I'd suggest making this a `to_u16()` inherent method instead.

rcgen/src/certificate.rs
```diff
@@ -455,6 +455,24 @@ impl CertificateParams {
 		Ok(result)
 	}
 
+	/// Write a certificate's KeyUsage as defined in RFC 5280.
+	fn write_key_usage(&self, writer: DERWriter) {
+		// RFC 5280 defines 9 key usages, which we detail in our key usage enum
+		// We could use std::mem::variant_count here, but it's experimental
+		const KEY_USAGE_BITS: usize = 9;
+
+		if !self.key_usages.is_empty() {
```
djc, line 464: Let's invert this condition and do an early return instead?

djc (commented):
I'd suggest reorienting the commit history as such:

* Introduce the separate `KeyUsagePurpose` to `u16` conversion method, and start using it in the same commit in place (and also other changes to always use 16 bits)
* Extract the `write_key_usage()` method from its current context
* Add writing key usages in CSR serialization

rcgen/src/certificate.rs
lvkv: Done!

lvkv:
A few notes:

- I've simplified the parsing of `KeyUsagePurpose`s with a `from_u16` function (complementing `to_u16`) that parses the BIT STRING value into a vector of key usages. It seems that the previous implementation was written so as to not depend on the bit order of x509-parser's flags, though it's a bit... if-fy (ha). What I've written is aware of the bit order, but is also faster and has tests added so that if x509 parser changes its bit order we'll know. I think it's better, but I'm fine with changing it if anyone feels strongly.
- Looks like the build CI steps (macOs, ubuntu, etc.) don't run with the x509-feature enabled. Should they?

rcgen/src/csr.rs
```diff
@@ -117,6 +117,11 @@ impl CertificateSigningRequestParams {
 		if let Some(extensions) = csr.requested_extensions() {
 			for ext in extensions {
 				match ext {
+					x509_parser::extensions::ParsedExtension::KeyUsage(key_usage) => {
+						// This x509 parser stores flags in reversed bit BIT STRING order
+						params.key_usages =
+							crate::KeyUsagePurpose::from_u16(key_usage.flags.reverse_bits());
```
djc, line 123: Nit: import `KeyUsagePurpose`?

djc (approved):
This makes sense to me, thanks!

djc:
> * Looks like the build CI steps (macOs, ubuntu, etc.) don't run with the x509-feature enabled. Should they?

It looks like the coverage job enables `--all-features`. I guess that might be enough?

rcgen/src/csr.rs
lvkv: Done!

est31 (approved):

Issue #282, opened 2024-06-30
CSR with PublicKey only


Virviil:
Usecase:

* Client generates keypair
* Sends public key to server (in form of CSR, or CSR is generated on server itself)
* Server signs him a client cert with CA existing only on server
* Returns this cert back to client

Mb I'm missing something, but didn't find a way to do it. The only possible flow now is:

* Server generates both private and public key for client (`KeyPair::generate`)
* Signs a cert with existing CA
* Returns both cert and private key to a client.

djc:
If I understand correctly, a CSR must be signed by the subject's key, which requires the private key.

Virviil:
@djc Yes, for sure. 

Actually i've found the possible solution: it's possible to construct [rcgen](https://docs.rs/rcgen/0.13.1/rcgen/index.html)::[CertificateSigningRequestParams](https://docs.rs/rcgen/0.13.1/rcgen/struct.CertificateSigningRequestParams.html#) with only params and public key, and then use it to sign a client cert by [CertificateSigningRequestParams::signed_by](https://docs.rs/rcgen/0.13.1/rcgen/struct.CertificateSigningRequestParams.html#method.signed_by)

Maybe it's worth adding this to the examples.

est31:
Hmmm I wonder whether we could make `CertificateParams::signed_by` take `K: PublicKeyData` instead? The private key is never used.

djc:
Huh, yeah -- I totally missed that, sorry.

cpu:
Let me start by sketching out the workflow I'd expect to see here.

Client side:

* Client generates key pair w/ `KeyPair::generate()` or another fn as appropriate.
* Client generates `rcgen::CertificateParams`, customizing as required to describe attributes of the certificate it desires.
* Client generates `rcgen::CertificateSigningRequest` from the `rcgen::CertificateParams` by calling `rcgen::CertificateParams::serialize_request(subject_key)` where `subject_key` is the public half of the client's `KeyPair`.
* Client serializes the `rcgen::CertificateSigningRequest` to PEM or DER w/ `rcgen::CertificateSigningRequest::pem()/der()` and transmits this to the CA over a trusted channel.

Server side:

* Server receives PEM/DER CSR from a trusted client.
* Server creates `rcgen::CertificateSigningRequestParams::from_pem()|from_der()`, providing the CSR data. Importantly this verifies the client's signature over the CSR and proves control of the private key corresponding to the public key that is the CSR subject.
[...]

PR #288, opened 2024-08-21
make `CertificateParams::signed_by` accept `impl PublicKeyData` rather than `KeyPair` for the key being signed

kwantam:
`CertificateParams::signed_by` does not need the secret corresponding to the public key being certified (though of course it needs the secret for the CA key signing the certificate!). More discussion about this in #282.

This PR:
- changes the first arg of `signed_by` to `impl PublicKeyData` (which maintains compatibility because `KeyPair` implements this trait)
- adds a new type, `SubjectPublicKey`, that can be parsed from a PEM or DER output by this crate (or, e.g., by openssl) and also impls `PublicKeyData.` This type is only available when the `x509-parser` feature is enabled (because that's what is required to parse).

I added a quick test that the `SubjectPublicKey` type's `PublicKeyData` impl produces the same output as `KeyPair` would for the same key, at least for the key types that can be generated using the crate. (So this means, no test right now for RSA keys, since those seemed to give an error when I tried to generate a keypair.)

### DIFF
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -36,7 +36,7 @@ x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
-default = ["crypto", "pem", "ring"]
+default = ["crypto", "pem", "ring", "x509-parser"]
 crypto = []
 aws_lc_rs = ["crypto", "dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys"]
 ring = ["crypto", "dep:ring"]
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -147,9 +147,9 @@ impl CertificateParams {
 	///
 	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
 	/// [`Certificate::pem`].
-	pub fn signed_by(
+	pub fn signed_by<K: PublicKeyData>(
 		self,
-		key_pair: &KeyPair,
+		key_pair: &K,
 		issuer: &Certificate,
 		issuer_key: &KeyPair,
 	) -> Result<Certificate, Error> {
@@ -160,7 +160,8 @@ impl CertificateParams {
 			key_pair: issuer_key,
 		};
 
-		let subject_public_key_info = key_pair.public_key_der();
+		let subject_public_key_info =
+			yasna::construct_der(|writer| key_pair.serialize_public_key_der(writer));
 		let der = self.serialize_der_with_signer(key_pair, issuer)?;
 		Ok(Certificate {
 			params: self,
--- a/rcgen/src/error.rs
+++ b/rcgen/src/error.rs
@@ -45,6 +45,9 @@ pub enum Error {
 	#[cfg(not(feature = "crypto"))]
 	/// Missing serial number
 	MissingSerialNumber,
+	/// X509 parsing error
+	#[cfg(feature = "x509-parser")]
+	X509,
 }
 
 impl fmt::Display for Error {
@@ -91,6 +94,8 @@ impl fmt::Display for Error {
 			)?,
 			#[cfg(not(feature = "crypto"))]
 			MissingSerialNumber => write!(f, "A serial number must be specified")?,
+			#[cfg(feature = "x509-parser")]
+			X509 => write!(f, "X509 error")?,
 		};
 		Ok(())
 	}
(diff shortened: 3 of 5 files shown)


rcgen/Cargo.toml
```diff
@@ -36,7 +36,7 @@ x509-parser = { workspace = true, features = ["verify"], optional = true }
 zeroize = { version = "1.2", optional = true }
 
 [features]
-default = ["crypto", "pem", "ring"]
+default = ["crypto", "pem", "ring", "x509-parser"]
```
djc, line 39: We'll want to revert this.

rcgen/src/certificate.rs
```diff
@@ -147,9 +147,9 @@ impl CertificateParams {
 	///
 	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
 	/// [`Certificate::pem`].
-	pub fn signed_by(
+	pub fn signed_by<K: PublicKeyData>(
 		self,
-		key_pair: &KeyPair,
+		key_pair: &K,
```
djc, line 152: Nit: I'd prefer to spell this as `&impl PublicKeyData`.

rcgen/src/key_pair.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Debug for KeyPairKind {
 	}
 }
 
+/// A public key
+#[cfg(feature = "x509-parser")]
```
djc, line 61: There's no need to guard the type directly.

rcgen/src/key_pair.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Debug for KeyPairKind {
...
+impl SubjectPublicKey {
+	/// Create a `SubjectPublicKey` value from a PEM-encoded SubjectPublicKeyInfo string
+	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
+		let spki_der = pem::parse(pem_str)._err()?.into_contents();
+		Self::from_der(&spki_der)
+	}
+
+	/// Create a `SubjectPublicKey` value from DER-encoded SubjectPublicKeyInfo bytes
+	pub fn from_der(spki_der: &[u8]) -> Result<Self, Error> {
```
djc, line 77: We should guard this constructor.

rcgen/src/key_pair.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Debug for KeyPairKind {
...
+#[derive(Debug)]
+pub struct SubjectPublicKey {
+	pub(crate) alg: &'static SignatureAlgorithm,
+	pub(crate) subject_public_key: Vec<u8>,
+}
+
+#[cfg(feature = "x509-parser")]
+impl SubjectPublicKey {
+	/// Create a `SubjectPublicKey` value from a PEM-encoded SubjectPublicKeyInfo string
+	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
```
djc, line 71: This needs the feature guards from `from_der()` and also another one for `pem`.

rcgen/src/key_pair.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Debug for KeyPairKind {
...
+			})
+			.ok_or(Error::UnsupportedSignatureAlgorithm)?;
+		Ok(Self {
+			alg,
+			subject_public_key: Vec::from(spki.subject_public_key.as_ref()),
+		})
+	}
+}
+
+#[cfg(feature = "x509-parser")]
```
djc, line 108: This doesn't need the guard.

rcgen/src/key_pair.rs
```diff
@@ -689,7 +773,7 @@ impl<T> ExternalError<T> for Result<T, pem::PemError> {
 	}
 }
 
-pub(crate) trait PublicKeyData {
+pub trait PublicKeyData {
```
djc, line 776: Needs some tweaks, as discussed here, please put those in a separate commit.

https://github.com/rustls/rcgen/issues/282#issuecomment-2206239673

Also needs documentation.

rcgen/src/error.rs
```diff
@@ -45,6 +45,9 @@ pub enum Error {
 	#[cfg(not(feature = "crypto"))]
 	/// Missing serial number
 	MissingSerialNumber,
+	/// X509 parsing error
+	#[cfg(feature = "x509-parser")]
+	X509,
```
djc, line 50: I think we might want to include a `String` representation of the nested error here.

djc (commented):
Initial round of feedback.

est31 (approved):
LGTM if the review comments are addressed

cpu (commented):
djc's feedback covered most bases, I just had a couple nits to consider.

cpu:
@kwantam Are you interested in completing this pull-request by addressing the feedback received?

djc:
I've taken this branch and addressed the comments so far in #291.

cpu:
Closing in favour of #291. Thanks for getting the ball rolling Kwantam

kwantam:
Thanks for the comments and @djc for getting this across the finish line! Apologies that I've been buried in other stuff...

PR #289, opened 2024-08-26
lib: use From trait for KeyUsagePurpose -> u16

cpu:
Quick follow-up from https://github.com/rustls/rcgen/pull/287 Using a `From` impl feels a bit more idiomatic in this case.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -436,23 +436,6 @@ pub enum KeyUsagePurpose {
 }
 
 impl KeyUsagePurpose {
-	/// Encode a key usage as the value of a BIT STRING as defined by RFC 5280.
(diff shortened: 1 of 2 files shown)


rcgen/src/lib.rs
```diff
@@ -470,13 +453,32 @@ impl KeyUsagePurpose {
...
+			let present = u16::from(*key_usage) & value != 0;
 			present.then_some(*key_usage)
 		})
 		.collect()
 	}
 }
 
+/// Encode a key usage as the value of a BIT STRING as defined by RFC 5280.
+/// [`u16`] is sufficient to encode the largest possible key usage value (two bytes).
+impl From<KeyUsagePurpose> for u16 {
```
djc, line 465: I dissuaded the OP from this because I thought we'd prefer to keep this interface private.

PR #290, opened 2024-09-19
Fix compilation issues of OpenSSL tests on 32-bit architectures

decathorpe:
We discovered this as a build failure on i686 when packaging this crate for Fedora Linux.

It looks like `openssl::asn1::Asn1Time::from_unix` takes a 64-bit number on 64-bit platforms only, and a 32-bit number on 32-bit platforms.

This PR adds the suggestion that's made by rustc - add `.try_into().unwrap()` to explicitly convert to an u32 (and fail if the number is out of range).

### DIFF
--- a/rcgen/tests/openssl.rs
+++ b/rcgen/tests/openssl.rs
@@ -422,11 +422,27 @@ fn test_openssl_crl_parse() {
 
 	// The properties of the CRL should match expected.
 	let openssl_issuer = X509::from_der(issuer.der()).unwrap();
-	let expected_last_update =
-		Asn1Time::from_unix(crl.params().this_update.unix_timestamp()).unwrap();
+	// Asn1Time::from_unix takes i64 or i32 (depending on CPU architecture)
+	#[allow(clippy::useless_conversion)]
+	let expected_last_update = Asn1Time::from_unix(
+		crl.params()
+			.this_update
+			.unix_timestamp()
+			.try_into()
+			.unwrap(),
+	)
+	.unwrap();
 	assert!(openssl_crl.last_update().eq(&expected_last_update));
-	let expected_next_update =
-		Asn1Time::from_unix(crl.params().next_update.unix_timestamp()).unwrap();
+	// Asn1Time::from_unix takes i64 or i32 (depending on CPU architecture)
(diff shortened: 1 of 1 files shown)


djc (approved):
Makes sense, thanks!

rcgen/tests/openssl.rs
```diff
@@ -422,11 +422,23 @@ fn test_openssl_crl_parse() {
...
+	)
+	.unwrap();
 	assert!(openssl_crl.last_update().eq(&expected_last_update));
-	let expected_next_update =
-		Asn1Time::from_unix(crl.params().next_update.unix_timestamp()).unwrap();
+	let expected_next_update = Asn1Time::from_unix(
+		crl.params()
+			.next_update
+			.unix_timestamp()
+			.try_into()
```
cpu, line 438: I think these `try_into()`'s are getting flagged on the 64bit build as an "unnecessary conversion": https://github.com/rustls/rcgen/actions/runs/10948893210/job/30400945346?pr=290

I would be OK with disabling the lint for both instead of trying to do a fancier gating based on arch.

decathorpe:
It seems like clippy needs to be told somehow that this conversion is, indeed, not useless when considering different architectures? Should I add something like `#[allow(clippy::useless_conversion)]` to the statement with a comment?

djc:
> It seems like clippy needs to be told somehow that this conversion is, indeed, not useless when considering different architectures? Should I add something like `#[allow(clippy::useless_conversion)]` to the statement with a comment?

Yup, that seems fine.

rcgen/tests/openssl.rs
decathorpe: I applied a clippy allow attribute to the statements. Sorry for missing this before hitting "submit" on my own comment :)

cpu: Np :-)

cpu (approved):
Thanks!

cpu:
> cpu enabled auto-merge now
 
 Trivial fix in test-only code, seems fair to merge without blocking for further reviews.

decathorpe:
Thanks!

PR #291, opened 2024-09-24
Enable signing without private key

### LINKED ISSUE #282: CSR with PublicKey only

djc:
Takes #288 and addresses the comments.

Fixes #282.

### DIFF
--- a/rcgen/src/error.rs
+++ b/rcgen/src/error.rs
@@ -45,6 +45,9 @@ pub enum Error {
 	#[cfg(not(feature = "crypto"))]
 	/// Missing serial number
 	MissingSerialNumber,
+	/// X509 parsing error
+	#[cfg(feature = "x509-parser")]
+	X509(String),
 }
 
 impl fmt::Display for Error {
@@ -91,6 +94,8 @@ impl fmt::Display for Error {
 			)?,
 			#[cfg(not(feature = "crypto"))]
 			MissingSerialNumber => write!(f, "A serial number must be specified")?,
+			#[cfg(feature = "x509-parser")]
+			X509(e) => write!(f, "X.509 parsing error: {e}")?,
 		};
 		Ok(())
 	}
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -690,9 +690,12 @@ impl<T> ExternalError<T> for Result<T, pem::PemError> {
 	}
 }
 
-pub(crate) trait PublicKeyData {
+/// The public key data of a key pair
+pub trait PublicKeyData {
+	/// The public key in DER format
 	fn der_bytes(&self) -> &[u8];
 
+	/// The algorithm used by the key pair
 	fn algorithm(&self) -> &SignatureAlgorithm;
 }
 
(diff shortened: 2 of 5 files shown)


rcgen/src/error.rs
```diff
@@ -91,6 +94,8 @@ impl fmt::Display for Error {
 			)?,
 			#[cfg(not(feature = "crypto"))]
 			MissingSerialNumber => write!(f, "A serial number must be specified")?,
+			#[cfg(feature = "x509-parser")]
+			X509(e) => write!(f, "X509 error: {e}")?,
```
cpu, line 98: nit: maybe `X.509 parsing error:` as the prefix?

rcgen/src/key_pair.rs
```diff
@@ -690,6 +690,69 @@ impl<T> ExternalError<T> for Result<T, pem::PemError> {
 	}
 }
 
+/// A public key
+#[derive(Debug)]
```
cpu, line 694: It feels like deriving `Eq` and `PartialEq` could be helpful here. WDYT?

rcgen/src/key_pair.rs
```diff
@@ -690,6 +690,69 @@ impl<T> ExternalError<T> for Result<T, pem::PemError> {
...
+	pub(crate) alg: &'static SignatureAlgorithm,
+	pub(crate) subject_public_key: Vec<u8>,
+}
+
+impl SubjectPublicKeyInfo {
+	/// Create a `SubjectPublicKey` value from a PEM-encoded SubjectPublicKeyInfo string
+	#[cfg(all(feature = "x509-parser", feature = "pem"))]
+	pub fn from_pem(pem_str: &str) -> Result<Self, Error> {
+		let spki_der = pem::parse(pem_str)._err()?.into_contents();
+		Self::from_der(&spki_der)
```
cpu, line 705: nit: inline `spki_der`?:

```rust
        Self::from_der(&pem::parse(pem_str)._err()?.into_contents())
```

rcgen/src/key_pair.rs
```diff
@@ -716,6 +779,27 @@ mod test {
 		signature::{EcdsaKeyPair, ECDSA_P256_SHA256_FIXED_SIGNING},
 	};
 
+	#[cfg(all(feature = "x509-parser", feature = "pem"))]
+	#[test]
+	fn test_subject_public_key_parsing() {
+		// NOTE: the other algorithms supported by this crate don't support keygen
```
cpu, line 785: Is this comment accurate? We support RSA and ECDSA P521 keygen w/ aws-lc-rs no?

cpu (approved):
Overall looks good, thanks for picking this up.

kwantam:
Thanks for doing this and apologies that I have been buried!

PR #296, opened 2024-10-30
Add PKCS#10 attributes to CSR serializer

lvkv:
Coming from https://github.com/rustls/rcgen/issues/285, this PR adds PKCS#10 attributes to rcgen's CSR serialization logic.
Of course, rcgen has already implicitly supported at most one attribute in the form of PKCS#9's extensionRequest[^1], but this allows you to add others.

I haven't split up the commits yet in anticipation of review comments, but the rough changes are (as of writing):
- Add an `Attribute` type that represents an RFC 5280[^2]/RFC 2986[^3] (take your pick) ATTRIBUTE.
- Add a `custom_csr_attributes` field to `CertificateParams`, of type `Attribute`
- Add serialization logic for `custom_csr_attributes` in `CertificateParams::serialize_request`
- Add an end-to-end/roundtrip test for CSR serialization with custom attributes
- Add to and update existing CSR serialization tests

Along the way, I've updated a few items that may or may not count as bugs. These are all subtle, so feel free to be pedantic when reviewing them:
- Don't write out DER for SANs if no SANs are specified. The current output isn't technically malformed ("hey, here are 0 SANs"), but it might be unexpected. Edit: yeah, the current behavior violates RFC 5280, see https://github.com/rustls/rcgen/pull/296#discussion_r1823012181 
- _Do_ still write out DER for certificate extensions in the edge case that someone wants an EKU but hasn't specified any standard key usages. The current behavior forgets to put that in the CSR. I believe it's RFC-compliant to do this (i.e. neither PKCS#9, PKCS#10 nor RFC 5280 say you can't), so the current behavior feels like a bug to me.
- Write CSR attributes using an implicit ASN.1 tag instead of an explicit one (see https://github.com/qnighy/yasna.rs/issues/8). This also correctly forces us to encode attributes as an implicit SET OF. I believe the current implementation encodes a single attribute correctly, but it's almost by accident, and definitely doesn't work if you want a number of attributes that isn't 0 or 1.

[^1]: https://datatracker.ietf.org/doc/html/rfc2985
[^2]: https://datatracker.ietf.org/doc/html/rfc5280
[^3]: https://datatracker.ietf.org/doc/html/rfc2986

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -534,6 +534,25 @@ impl CertificateParams {
 		&self,
 		subject_key: &KeyPair,
 	) -> Result<CertificateSigningRequest, Error> {
+		self.serialize_request_with_attributes(subject_key, Vec::new())
+	}
+
+	/// Generate and serialize a certificate signing request (CSR) with custom PKCS #10 attributes.
+	/// as defined in [RFC 2986].
+	///
+	/// The constructed CSR will contain attributes based on the certificate parameters,
+	/// and include the subject public key information from `subject_key`. Additionally,
+	/// the CSR will be self-signed using the subject key.
+	///
+	/// Note that subsequent invocations of `serialize_request_with_attributes()` will not produce the exact
+	/// same output.
+	///
+	/// [RFC 2986]: <https://datatracker.ietf.org/doc/html/rfc2986#section-4>
+	pub fn serialize_request_with_attributes(
+		&self,
+		subject_key: &KeyPair,
+		attrs: Vec<Attribute>,
+	) -> Result<CertificateSigningRequest, Error> {
 		// No .. pattern, we use this to ensure every field is used
 		#[deny(unused)]
 		let Self {
@@ -582,11 +601,9 @@ impl CertificateParams {
 		let der = subject_key.sign_der(|writer| {
 			// Write version
 			writer.next().write_u8(0);
-			// Write subject name
 			write_distinguished_name(writer.next(), distinguished_name);
-			// Write subjectPublicKeyInfo
 			serialize_public_key_der(subject_key, writer.next());
-			// Write extensions
+
 			// According to the spec in RFC 2986, even if attributes are empty we need the empty attribute tag
 			writer
 				.next()
@@ -595,6 +612,13 @@ impl CertificateParams {
 					writer.write_set_of(|writer| {
 						if write_extension_request {
 							self.write_extension_request_attribute(writer.next());
+						}
+
+						for Attribute { oid, values } in attrs {
+							writer.next().write_sequence(|writer| {
+								writer.next().write_oid(&ObjectIdentifier::from_slice(&oid));
+								writer.next().write_der(&values);
+							});
 						}
 					});
 				});
@@ -844,6 +868,25 @@ fn write_general_subtrees(writer: DERWriter, tag: u64, general_subtrees: &[Gener
 			}
 		});
 	});
+}
+
+/// A PKCS #10 CSR attribute, as defined in [RFC 5280] and constrained
+/// by [RFC 2986].
+///
+/// [RFC 5280]: <https://datatracker.ietf.org/doc/html/rfc5280#appendix-A.1>
+/// [RFC 2986]: <https://datatracker.ietf.org/doc/html/rfc2986#section-4>
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct Attribute {
+	/// `AttributeType` of the `Attribute`, defined as an `OBJECT IDENTIFIER`.
+	pub oid: &'static [u64],
+	/// DER-encoded values of the `Attribute`, defined by [RFC 2986] as:
+	///
+	/// ```text
+	/// SET SIZE(1..MAX) OF ATTRIBUTE.&Type({IOSet}{@type})
+	/// ```
+	///
+	/// [RFC 2986]: https://datatracker.ietf.org/doc/html/rfc2986#section-4
+	pub values: Vec<u8>,
 }
 
 /// A custom extension of a certificate, as specified in
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -546,6 +546,12 @@ impl CertificateParams {
 			return Err(Error::UnsupportedInCsr);
 		}
 
+		// Whether or not to write an extension request attribute
+		let write_extension_request = !key_usages.is_empty()
+			|| !subject_alt_names.is_empty()
+			|| !extended_key_usages.is_empty()
+			|| !custom_extensions.is_empty();
+
 		let der = subject_key.sign_der(|writer| {
 			// Write version
 			writer.next().write_u8(0);
@@ -556,10 +562,7 @@ impl CertificateParams {
 			// Write extensions
 			// According to the spec in RFC 2986, even if attributes are empty we need the empty attribute tag
 			writer.next().write_tagged(Tag::context(0), |writer| {
-				if !key_usages.is_empty()
-					|| !subject_alt_names.is_empty()
-					|| !custom_extensions.is_empty()
-				{
+				if write_extension_request {
 					writer.write_sequence(|writer| {
 						let oid = ObjectIdentifier::from_slice(oid::PKCS_9_AT_EXTENSION_REQUEST);
 						writer.next().write_oid(&oid);
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -427,6 +427,33 @@ impl CertificateParams {
 		Ok(result)
 	}
 
+	/// Write a CSR extension request attribute as defined in [RFC 2985].
+	///
+	/// [RFC 2985]: <https://datatracker.ietf.org/doc/html/rfc2985>
+	fn write_extension_request_attribute(&self, writer: DERWriter) {
+		writer.write_sequence(|writer| {
+			writer.next().write_oid(&ObjectIdentifier::from_slice(
+				oid::PKCS_9_AT_EXTENSION_REQUEST,
+			));
+			writer.next().write_set(|writer| {
+				writer.next().write_sequence(|writer| {
+					// Write key_usage
+					self.write_key_usage(writer.next());
+					// Write subject_alt_names
+					self.write_subject_alt_names(writer.next());
+					self.write_extended_key_usage(writer.next());
+
+					// Write custom extensions
+					for ext in &self.custom_extensions {
+						write_x509_extension(writer.next(), &ext.oid, ext.critical, |writer| {
+							writer.write_der(ext.content())
+						});
+					}
+				});
+			});
+		});
+	}
+
 	/// Write a certificate's KeyUsage as defined in RFC 5280.
 	fn write_key_usage(&self, writer: DERWriter) {
 		// RFC 5280 defines 9 key usages, which we detail in our key usage enum
@@ -567,30 +594,7 @@ impl CertificateParams {
 					// RFC 2986 specifies that attributes are a SET OF Attribute
 					writer.write_set_of(|writer| {
 						if write_extension_request {
-							writer.next().write_sequence(|writer| {
-								let oid =
-									ObjectIdentifier::from_slice(oid::PKCS_9_AT_EXTENSION_REQUEST);
-								writer.next().write_oid(&oid);
-								writer.next().write_set(|writer| {
-									writer.next().write_sequence(|writer| {
-										// Write key_usage
-										self.write_key_usage(writer.next());
-										// Write subject_alt_names
-										self.write_subject_alt_names(writer.next());
-										self.write_extended_key_usage(writer.next());
-
-										// Write custom extensions
-										for ext in custom_extensions {
-											write_x509_extension(
-												writer.next(),
-												&ext.oid,
-												ext.critical,
-												|writer| writer.write_der(ext.content()),
-											);
-										}
-									});
-								});
-							});
+							self.write_extension_request_attribute(writer.next());
 						}
 					});
 				});
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -461,6 +461,10 @@ impl CertificateParams {
 	}
 
 	fn write_subject_alt_names(&self, writer: DERWriter) {
+		if self.subject_alt_names.is_empty() {
+			return;
+		}
+
 		write_x509_extension(writer, oid::SUBJECT_ALT_NAME, false, |writer| {
 			writer.write_sequence(|writer| {
 				for san in self.subject_alt_names.iter() {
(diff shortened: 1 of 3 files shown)


rcgen/src/certificate.rs
```diff
@@ -461,6 +467,10 @@ impl CertificateParams {
 	}
 
 	fn write_subject_alt_names(&self, writer: DERWriter) {
+		if self.subject_alt_names.is_empty() {
```
est31, line 470: this is actually required by the spec, it seems:

```
If the subjectAltName extension is present, the sequence MUST contain
   at least one entry.
```

lvkv: Nice! If I'm understanding correctly, that means that the current behavior—which allows a sequence with zero entries—is incorrect, and the code addition you highlighted is correct.

cpu:
Hope to give this a first pass soon. Thanks!

cpu (commented):
I think this is going in the right direction. Thanks! 

Would you mind getting it split into smaller commits and knocking out some of the first round of feedback? I think it'll be easier for others to review when the changes are isolated and describe their purpose in the commit message :+1: That will also give me some time to double check the one bit of ASN.1 I wasn't confident about.

djc:
Thanks @cpu for doing an initial round of review. I had some time off last week so didn't get to this, but would definitely want this restructured into basically one commit per bullet point in your PR description (or more, potentially), and reviewing would be a lot easier based on that.

rcgen/src/certificate.rs
```diff
@@ -81,6 +81,11 @@ pub struct CertificateParams {
 	/// [^1]: <https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.13>
```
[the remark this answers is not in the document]
lvkv: Oh, yeah, this is probably the #1 most annoying thing about the GitHub UI

rcgen/src/certificate.rs
```diff
@@ -81,6 +81,11 @@ pub struct CertificateParams {
...
 	pub crl_distribution_points: Vec<CrlDistributionPoint>,
 	pub custom_extensions: Vec<CustomExtension>,
+	/// An optional set of custom PKCS #10 certificate signing request (CSR) attributes as
+	/// defined in [RFC 2986][1].
+	///
+	/// [1]: <https://datatracker.ietf.org/doc/html/rfc2986#section-4>
+	pub custom_csr_attributes: Vec<Attribute>,
```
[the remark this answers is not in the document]
lvkv: Done!

> Ideally that would be represented in CertificateSigningRequestParams but I think the API we have today doesn't make that workable

Agreed. I also have a few opinions about how to make this API better and I imagine we've arrived to some of the same conclusions. Maybe we can open an issue and work on it after this PR?

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
 			// Write subjectPublicKeyInfo
 			serialize_public_key_der(subject_key, writer.next());
 			// Write extensions
+
+			// Whether or not to write an extension request attribute
+			let write_extension_request = !key_usages.is_empty()
+				|| !subject_alt_names.is_empty()
+				|| !extended_key_usages.is_empty()
+				|| !custom_extensions.is_empty();
```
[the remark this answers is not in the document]
lvkv: Ditto https://github.com/rustls/rcgen/pull/296#discussion_r1835528641

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
...
-					|| !subject_alt_names.is_empty()
-					|| !custom_extensions.is_empty()
-				{
-					writer.write_sequence(|writer| {
-						let oid = ObjectIdentifier::from_slice(oid::PKCS_9_AT_EXTENSION_REQUEST);
-						writer.next().write_oid(&oid);
-						writer.next().write_set(|writer| {
+			writer
+				.next()
+				.write_tagged_implicit(Tag::context(0), |writer| {
```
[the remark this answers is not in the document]
lvkv: Pulled out into a separate commit!

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
...
+								let oid =
+									ObjectIdentifier::from_slice(oid::PKCS_9_AT_EXTENSION_REQUEST);
+								writer.next().write_oid(&oid);
+								writer.next().write_set(|writer| {
+									writer.next().write_sequence(|writer| {
+										// Write key_usage
+										self.write_key_usage(writer.next());
+										// Write subject_alt_names
+										self.write_subject_alt_names(writer.next());
+										self.write_extended_key_usage(writer.next());
```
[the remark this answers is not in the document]
lvkv: (done)

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
...
-								// Write custom extensions
-								for ext in custom_extensions {
-									write_x509_extension(
-										writer.next(),
-										&ext.oid,
-										ext.critical,
-										|writer| writer.write_der(ext.content()),
-									);
-								}
+								let oid =
```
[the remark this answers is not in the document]
lvkv: (done)

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
...
+									});
+								});
 							});
-						});
+						}
+
+						// Write any custom CSR attributes
+						for Attribute { oid, values } in custom_attributes {
+							writer.next().write_sequence(|writer| {
+								let oid = ObjectIdentifier::from_slice(oid);
```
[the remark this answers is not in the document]
lvkv: (done)

rcgen/tests/generic.rs
```diff
@@ -135,6 +135,70 @@ mod test_x509_custom_ext {
...
+			oid: challenge_pwd_oid.clone(),
+			values: challenge_pwd_values.clone(),
+		};
+
+		// Serialize a DER-encoded CSR
+		let mut params = CertificateParams::default();
+		params
+			.custom_csr_attributes
+			.push(challenge_password_attribute);
+		// params.key_usages.push(KeyUsagePurpose::DigitalSignature);
```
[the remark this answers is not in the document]
lvkv: (done)

rcgen/src/certificate.rs
```diff
@@ -550,37 +561,57 @@ impl CertificateParams {
...
-					writer.write_sequence(|writer| {
-						let oid = ObjectIdentifier::from_slice(oid::PKCS_9_AT_EXTENSION_REQUEST);
-						writer.next().write_oid(&oid);
-						writer.next().write_set(|writer| {
+			writer
+				.next()
+				.write_tagged_implicit(Tag::context(0), |writer| {
+					// RFC 2986 specifies that attributes are a SET OF Attribute
+					writer.write_set_of(|writer| {
+						// Write the extension request CSR attribute
```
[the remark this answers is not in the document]
lvkv: Sure, I added this in [dafb9bf](https://github.com/rustls/rcgen/pull/296/commits/dafb9bfca01f3607eef6d8d14fede88ce750ea31)!

rcgen/src/certificate.rs
lvkv: I refactored this into [a4f8b94](https://github.com/rustls/rcgen/pull/296/commits/a4f8b94675fb0df45ed2ab3638113bbe52e093a4)

rcgen/tests/generic.rs
```diff
@@ -395,12 +388,59 @@ mod test_csr {
...
+
+	#[test]
+	fn test_csr_with_key_usgaes_and_extended_key_usages_roundtrip() {
+		let mut params = CertificateParams::default();
+		params.key_usages = vec![KeyUsagePurpose::DigitalSignature];
+		params.extended_key_usages = vec![ExtendedKeyUsagePurpose::ClientAuth];
+		generate_and_test_parsed_csr(&params);
+	}
+
+	fn generate_and_test_parsed_csr(params: &CertificateParams) {
```
djc, line 517: I'd like to have one commit here that reorganizes the tests to build on this helper function, and a separate commit that adds all the additional validation logic.

rcgen/tests/generic.rs
```diff
@@ -358,6 +358,29 @@ mod test_parse_other_name_alt_name {
...
+	use x509_parser::prelude::{FromDer, ParsedExtension};
+
+	#[test]
+	fn dont_write_sans_extension_if_no_sans_are_present() {
+		let mut params = CertificateParams::default();
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		let key_pair = KeyPair::generate().unwrap();
+		let csr = params.serialize_request(&key_pair).unwrap();
+		let (_, parsed_csr) =
+			x509_parser::certification_request::X509CertificationRequest::from_der(csr.der())
```
djc, line 434: Please import `X509CertificationRequest` directly instead of using the long qualified path (same for other tests added here).

rcgen/src/certificate.rs
```diff
@@ -567,30 +594,7 @@ impl CertificateParams {
...
-												writer.next(),
-												&ext.oid,
-												ext.critical,
-												|writer| writer.write_der(ext.content()),
-											);
-										}
-									});
-								});
-							});
+							self.write_extension_request_attribute(writer.next());
```
djc, line 599: Nit: suggest moving the `write_extension_request` check into the new method.

rcgen/src/certificate.rs
```diff
@@ -607,6 +616,21 @@ impl CertificateParams {
...
 
+	/// Generate and serialize a certificate signing request (CSR).
+	///
+	/// The constructed CSR will contain attributes based on the certificate parameters,
+	/// and include the subject public key information from `subject_key`. Additionally,
+	/// the CSR will be self-signed using the subject key.
+	///
+	/// Note that subsequent invocations of `serialize_request()` will not produce the exact
+	/// same output.
+	pub fn serialize_request(
```
djc, line 627: Nit: `serialize_request()` should appear before `serialize_request_with_attributes()`.

rcgen/src/certificate.rs
```diff
@@ -846,6 +870,25 @@ fn write_general_subtrees(writer: DERWriter, tag: u64, general_subtrees: &[Gener
...
 
+/// A PKCS #10 CSR attribute, as defined in [RFC 5280][1] and constrained
+/// by [RFC 2986][2].
+///
+/// [1]: <https://datatracker.ietf.org/doc/html/rfc5280#appendix-A.1>
+/// [2]: <https://datatracker.ietf.org/doc/html/rfc2986#section-4>
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct Attribute {
+	/// `AttributeType` of the `Attribute`, defined as an `OBJECT IDENTIFIER`.
+	pub oid: Vec<u64>,
```
djc, line 881: How about we use `&'static [u64]` for this instead, to match the type used in the `oid` module today?

djc (approved):

rcgen/src/certificate.rs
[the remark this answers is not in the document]
cpu: > Maybe we can open an issue and work on it after this PR?

SGTM!

rcgen/src/certificate.rs
lvkv: I considered doing this, but it makes `subject_alt_names`, `key_usages`, and `custom_extensions` unused, which we deny here:

https://github.com/rustls/rcgen/blob/02d5c6096ec54b491f60581f7e51e83bdd764d84/rcgen/src/certificate.rs#L507-L522

The alternative is making `CertificateParams::write_extension_request_attribute` static, i.e. not a function of `&self`. I'm fine with the way it's currently written, so this one is your call!

lvkv:
@djc this is ready for another look!

rcgen/src/certificate.rs
lvkv: (Done)

rcgen/src/certificate.rs
lvkv: (Done)

rcgen/tests/generic.rs
lvkv: Sure—I ended up adding another commit too so that the flow makes sense. Lmk what you think!

rcgen/tests/generic.rs
lvkv: Done!

rcgen/tests/generic.rs
```diff
@@ -400,7 +400,29 @@ mod test_csr {
 		let csrp = CertificateSigningRequestParams::from_der(csr.der()).unwrap();
 
 		// Assert that our parsed parameters match our initial parameters
+		// Certificate parameters are #[non_exhaustive] and don't implement equality
```
djc, line 518: Do you have a hypothesis as to why we don't/can't implement `PartialEq`? Seems like we could and maybe should...

rcgen/src/certificate.rs
djc: Okay, I guess this is a reasonable mechanism for making sure we're taking everything into account.

djc (approved):

rcgen/tests/generic.rs
lvkv: I can't think of a reason not implement `PartialEq`. Same with `Eq`—I can't see a reason why we wouldn't want reflexive equality for certificate params. I've added a commit that derives both. I also threw in `Debug`, which we need for `assert_eq` and might be useful anyways.

lvkv: I can take any of these derives out if anyone feels strongly, of course.

djc: LGTM, thanks!

est31 (approved):
could you remove the `[1]` syntax? One can also do `[abcabc]` and then later `[abc]: https://example.com`. The issue with `[1]` syntax is that it might be resolved in a global context instead of a local one, where we then have multiple URLs pointing to `[1]`.

rcgen/tests/generic.rs
```diff
@@ -358,6 +358,27 @@ mod test_parse_other_name_alt_name {
...
+		params.key_usages.push(KeyUsagePurpose::DigitalSignature);
+		let key_pair = KeyPair::generate().unwrap();
+		let csr = params.serialize_request(&key_pair).unwrap();
+		let (_, parsed_csr) = X509CertificationRequest::from_der(csr.der()).unwrap();
+		parsed_csr
+			.requested_extensions()
+			.unwrap()
+			.for_each(|requested_ext| {
+				assert!(!(matches!(requested_ext, ParsedExtension::SubjectAlternativeName(_))));
+			});
```
cpu, line 435: nit: maybe a bit more natural to assert on `any()` like:

```rust
assert!(!parsed_csr
			.requested_extensions()
			.unwrap()
			.any(|ext| matches!(ext, ParsedExtension::SubjectAlternativeName(_))));
```

cpu (approved):
Really nice work!

rcgen/src/certificate.rs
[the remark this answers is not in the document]
cpu: Your analysis makes sense to me, and squares with my understanding of the ASN.1 modules in play. Thank you!

lvkv:
@est31 @cpu done!

cpu (approved):

Issue #300, opened 2024-11-21
Issues parsing -x509 with multiple domain components (DC)


da-kami:
I was not aware of this myself, but apparently `-x509` certificates can have multiple domain component fields set in the issuer. 

If I am not mistaken the RCF mentions this, see: https://datatracker.ietf.org/doc/html/rfc5280#section-7.3:

> Each domainComponent attribute represents a single label.

This has been reported as issue with other certificate management tools:
https://github.com/cert-manager/cert-manager/issues/4424 

I came across a client who is using a certificate that specifies multiple domain components in the issuer.
The client is already using this certificate with a firewall solution in production, I cannot easily tell them to replace the cert.
This forced me to create a fork and adapt the implementation of `DistinguishedName` because a `HashMap` won't allow storing multiple DC items.

PR #301, opened 2024-11-21
Multiple domain components in issuer

da-kami:
See: https://github.com/rustls/rcgen/issues/300

I did not invest a big amount of time into changing the interface; please let me know if this PR is welcome otherwise I'm happy to close it. It would be great if #300 can be overcome at some point.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -338,10 +338,10 @@ impl DistinguishedName {
 		self.entries.push((ty, s.into()));
 	}
 
-	/// Replaces the *fist occurrence* of a type with a new value.
+	/// Replaces the *first occurrence* of a type with a new value.
 	/// This is a convenience function to avoid duplicating values.
 	///
-	/// If there are multiple occurrences of a type there is currently no way of changing the besides iterating over the types and values of an existing instance and creating a new instance.
+	/// If there are multiple occurrences of a type there is currently no way of changing them besides iterating over the types and values of an existing instance and creating a new instance.
 	///
 	/// ```
 	/// # use rcgen::{DistinguishedName, DnType, DnValue};
@@ -586,7 +586,7 @@ fn write_dt_utc_or_generalized(writer: DERWriter, dt: OffsetDateTime) {
 	}
 }
 
-fn write_distinguished_name(writer: DERWriter, dn: DistinguishedName) {
+fn write_distinguished_name(writer: DERWriter, dn: &DistinguishedName) {
 	writer.write_sequence(|writer| {
 		for (ty, content) in dn.iter() {
 			writer.next().write_set(|writer| {
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -33,7 +33,6 @@ println!("{}", key_pair.serialize_pem());
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 #![warn(unreachable_pub)]
 
-use std::collections::HashMap;
 use std::fmt;
 use std::hash::Hash;
 use std::net::IpAddr;
@@ -299,8 +298,7 @@ See also the RFC 5280 sections on the [issuer](https://tools.ietf.org/html/rfc52
 and [subject](https://tools.ietf.org/html/rfc5280#section-4.1.2.6) fields.
 */
 pub struct DistinguishedName {
-	entries: HashMap<DnType, DnValue>,
-	order: Vec<DnType>,
+	entries: Vec<(DnType, DnValue)>,
 }
 
 impl DistinguishedName {
@@ -309,20 +307,22 @@ impl DistinguishedName {
 		Self::default()
 	}
 	/// Obtains the attribute value for the given attribute type
-	pub fn get(&self, ty: &DnType) -> Option<&DnValue> {
-		self.entries.get(ty)
+	pub fn get(&self, ty: &DnType) -> Vec<&DnValue> {
+		self.entries
+			.iter()
+			.filter_map(|(dn_type, dn_value)| if ty == dn_type { Some(dn_value) } else { None })
+			.collect()
 	}
-	/// Removes the attribute with the specified DnType
+	/// Removes all attributes with the specified DnType
 	///
 	/// Returns true when an actual removal happened, false
 	/// when no attribute with the specified DnType was
 	/// found.
 	pub fn remove(&mut self, ty: DnType) -> bool {
-		let removed = self.entries.remove(&ty).is_some();
-		if removed {
-			self.order.retain(|ty_o| &ty != ty_o);
-		}
-		removed
+		let prev_len = self.entries.len();
+		self.entries.retain(|(dn_type, _dn_val)| dn_type != &ty);
+
+		prev_len != self.entries.len()
 	}
 	/// Inserts or updates an attribute that consists of type and name
 	///
@@ -331,20 +331,41 @@ impl DistinguishedName {
 	/// let mut dn = DistinguishedName::new();
 	/// dn.push(DnType::OrganizationName, "Crab widgits SE");
 	/// dn.push(DnType::CommonName, DnValue::PrintableString("Master Cert".try_into().unwrap()));
-	/// assert_eq!(dn.get(&DnType::OrganizationName), Some(&DnValue::Utf8String("Crab widgits SE".to_string())));
-	/// assert_eq!(dn.get(&DnType::CommonName), Some(&DnValue::PrintableString("Master Cert".try_into().unwrap())));
+	/// assert_eq!(dn.get(&DnType::OrganizationName).get(0), Some(&DnValue::Utf8String("Crab widgits SE".to_string())).as_ref());
+	/// assert_eq!(dn.get(&DnType::CommonName).get(0), Some(&DnValue::PrintableString("Master Cert".try_into().unwrap())).as_ref());
 	/// ```
 	pub fn push(&mut self, ty: DnType, s: impl Into<DnValue>) {
-		if !self.entries.contains_key(&ty) {
-			self.order.push(ty.clone());
-		}
-		self.entries.insert(ty, s.into());
+		self.entries.push((ty, s.into()));
 	}
+
+	/// Replaces the *fist occurrence* of a type with a new value.
+	/// This is a convenience function to avoid duplicating values.
+	///
+	/// If there are multiple occurrences of a type there is currently no way of changing the besides iterating over the types and values of an existing instance and creating a new instance.
+	///
+	/// ```
+	/// # use rcgen::{DistinguishedName, DnType, DnValue};
+	/// let mut dn = DistinguishedName::new();
+	/// dn.push(DnType::CommonName, DnValue::PrintableString("Master Cert".try_into().unwrap()));
+	/// assert_eq!(dn.get(&DnType::CommonName).get(0), Some(&DnValue::PrintableString("Master Cert".try_into().unwrap())).as_ref());
+	/// dn.push(DnType::CommonName, DnValue::PrintableString("Other Master Cert".try_into().unwrap()));
+	/// assert_eq!(dn.get(&DnType::CommonName).get(1), Some(&DnValue::PrintableString("Other Master Cert".try_into().unwrap())).as_ref());
+	/// ```
+	pub fn replace_or_push(&mut self, ty: DnType, s: impl Into<DnValue>) {
+		for (dn_type, dn_value) in self.entries.iter_mut() {
+			if *dn_type == ty {
+				*dn_value = s.into();
+				return;
+			}
+		}
+
+		self.push(ty, s)
+	}
+
 	/// Iterate over the entries
 	pub fn iter(&self) -> DistinguishedNameIterator<'_> {
 		DistinguishedNameIterator {
-			distinguished_name: self,
-			iter: self.order.iter(),
+			iter: self.entries.iter(),
 		}
 	}
 
@@ -397,17 +418,14 @@ impl DistinguishedName {
 Iterator over [`DistinguishedName`] entries
 */
 pub struct DistinguishedNameIterator<'a> {
-	distinguished_name: &'a DistinguishedName,
-	iter: std::slice::Iter<'a, DnType>,
+	iter: std::slice::Iter<'a, (DnType, DnValue)>,
 }
 
 impl<'a> Iterator for DistinguishedNameIterator<'a> {
 	type Item = (&'a DnType, &'a DnValue);
 
 	fn next(&mut self) -> Option<Self::Item> {
-		self.iter
-			.next()
-			.and_then(|ty| self.distinguished_name.entries.get(ty).map(|v| (ty, v)))
+		self.iter.next().map(|(key, value)| (key, value))
 	}
 }
 
@@ -568,7 +586,7 @@ fn write_dt_utc_or_generalized(writer: DERWriter, dt: OffsetDateTime) {
 	}
 }
 
-fn write_distinguished_name(writer: DERWriter, dn: &DistinguishedName) {
+fn write_distinguished_name(writer: DERWriter, dn: DistinguishedName) {
 	writer.write_sequence(|writer| {
 		for (ty, content) in dn.iter() {
 			writer.next().write_set(|writer| {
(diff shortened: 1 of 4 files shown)


rcgen/src/lib.rs
```diff
@@ -331,21 +342,35 @@ impl DistinguishedName {
...
 	/// ```
 	pub fn push(&mut self, ty: DnType, s: impl Into<DnValue>) {
-		if !self.entries.contains_key(&ty) {
-			self.order.push(ty.clone());
-		}
-		self.entries.insert(ty, s.into());
+		self.entries.push_front((ty, s.into()));
 	}
-	/// Iterate over the entries
-	pub fn iter(&self) -> DistinguishedNameIterator<'_> {
```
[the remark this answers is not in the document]
da-kami: tbh I don't understand why the previous implementation would require the `DistinguishedNameIterator` as own iterator implementation. 

I'm implementing `Iterator` for `DistinguishedName` - so `iter()` is actually still available even though the explicit struct function was removed. 
I would argue that the interface is actually not breaking - `iter()` is still available after the change. 

@cpu  Is there a better way to achieve this so that semver accepts the change as non-breaking?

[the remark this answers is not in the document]
djc: @da-kami implementing `Iterator for DistinguishedName` does not make any method called `iter()` available -- `Iterator::iter()` does not exist. If you want to make it possible to represent multiple entries with the same `DnType`, I'd suggest changing the representation to hold a `Vec<(DnType, DnValue)>` and then build on top of that.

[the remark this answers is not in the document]
da-kami: @djc @cpu all done, ready for review again :)

rcgen/src/lib.rs
```diff
@@ -331,20 +331,41 @@ impl DistinguishedName {
...
+	/// assert_eq!(dn.get(&DnType::OrganizationName).get(0), Some(&DnValue::Utf8String("Crab widgits SE".to_string())).as_ref());
+	/// assert_eq!(dn.get(&DnType::CommonName).get(0), Some(&DnValue::PrintableString("Master Cert".try_into().unwrap())).as_ref());
 	/// ```
 	pub fn push(&mut self, ty: DnType, s: impl Into<DnValue>) {
-		if !self.entries.contains_key(&ty) {
-			self.order.push(ty.clone());
+		self.entries.push((ty, s.into()));
+	}
+
+	/// Replaces the *fist occurrence* of a type with a new value.
```
djc, line 341: "fist" -> "first"

rcgen/src/lib.rs
```diff
@@ -331,20 +331,41 @@ impl DistinguishedName {
...
+	/// This is a convenience function to avoid duplicating values.
+	///
+	/// If there are multiple occurrences of a type there is currently no way of changing the besides iterating over the types and values of an existing instance and creating a new instance.
```
djc, line 344: "changing the" -> "changing them"?

rcgen/src/lib.rs
```diff
@@ -397,17 +418,14 @@ impl DistinguishedName {
...
 }
 
 impl<'a> Iterator for DistinguishedNameIterator<'a> {
 	type Item = (&'a DnType, &'a DnValue);
 
 	fn next(&mut self) -> Option<Self::Item> {
-		self.iter
-			.next()
-			.and_then(|ty| self.distinguished_name.entries.get(ty).map(|v| (ty, v)))
+		self.iter.next().map(|(key, value)| (key, value))
```
djc, line 428: Looks like the `map()` call is unnecessary?

rcgen/src/lib.rs
```diff
@@ -568,7 +586,7 @@ fn write_dt_utc_or_generalized(writer: DERWriter, dt: OffsetDateTime) {
 	}
 }
 
-fn write_distinguished_name(writer: DERWriter, dn: &DistinguishedName) {
+fn write_distinguished_name(writer: DERWriter, dn: DistinguishedName) {
```
djc, line 589: Why do we need this change? It seems unrelated.

rcgen/tests/openssl.rs
```diff
@@ -1,21 +1,19 @@
...
 use openssl::ssl::{HandshakeError, SslAcceptor, SslConnector, SslMethod};
 use openssl::stack::Stack;
 use openssl::x509::store::{X509Store, X509StoreBuilder};
 use openssl::x509::{CrlStatus, X509Crl, X509Req, X509StoreContext, X509};
-
 use rcgen::{
 	BasicConstraints, Certificate, CertificateParams, DnType, DnValue, GeneralSubtree, IsCa,
 	KeyPair, NameConstraints,
 };
+use std::cell::RefCell;
```
djc, line 14: Don't change the order of the imports.

rcgen/tests/openssl.rs
```diff
@@ -540,3 +538,65 @@ fn test_openssl_pkcs1_and_sec1_keys() {
 	let pkcs8_ec_key_der = PrivateKeyDer::try_from(ec_key.private_key_to_pkcs8().unwrap()).unwrap();
 	KeyPair::try_from(&pkcs8_ec_key_der).unwrap();
 }
+
+#[test]
+#[cfg(feature = "x509-parser")]
+fn test_parse_certificate_with_multiple_domain_components() {
+	use rcgen::Ia5String;
+	use std::str::FromStr;
```
djc, line 546: Move these imports to the top of the file with the appropriate guard attributes, please.

rcgen/tests/openssl.rs
da-kami: I am not entirely sure what you mean, but my guess is that the `rcgen` imports should stay as separate block. I changed that. In case you meant something else please let me know.

rcgen/src/lib.rs
da-kami: It's necessary to map from `&(key, val)` to `(&key, &val)`:

image

If there is a more ergonomic way to do this let me know.

rcgen/tests/openssl.rs
da-kami: Done - I grouped them in a use block that is feature flagged. In case you prefer individual lines let me know.

da-kami: Done

djc (approved):
Thanks, this looks good to me!

est31 (approved):

rcgen/src/lib.rs
```diff
@@ -331,20 +331,41 @@ impl DistinguishedName {
...
 	pub fn push(&mut self, ty: DnType, s: impl Into<DnValue>) {
-		if !self.entries.contains_key(&ty) {
-			self.order.push(ty.clone());
+		self.entries.push((ty, s.into()));
+	}
+
+	/// Replaces the *first occurrence* of a type with a new value.
+	/// This is a convenience function to avoid duplicating values.
+	///
+	/// If there are multiple occurrences of a type there is currently no way of changing them besides iterating over the types and values of an existing instance and creating a new instance.
```
cpu, line 344:
```suggestion
	/// If there are multiple occurrences of a type that you wish to replace you must iterate over the types and values of an existing instance and creating a new instance.
```

rcgen/tests/openssl.rs
```diff
@@ -540,3 +544,62 @@ fn test_openssl_pkcs1_and_sec1_keys() {
 	let pkcs8_ec_key_der = PrivateKeyDer::try_from(ec_key.private_key_to_pkcs8().unwrap()).unwrap();
 	KeyPair::try_from(&pkcs8_ec_key_der).unwrap();
 }
+
+#[test]
+#[cfg(feature = "x509-parser")]
+fn test_parse_certificate_with_multiple_domain_components() {
```
cpu, line 550: I think this test should be moved to `tests/generic.rs` since it only relies on `x509-parser`. 

While the static test certificate was generated with the `openssl` command line tool, it's not using the Rust `openssl` crate dependency like the other tests in this file.

rcgen/tests/openssl.rs
```diff
@@ -540,3 +544,62 @@ fn test_openssl_pkcs1_and_sec1_keys() {
...
+p3mmtsYxi+XCHdwUwRLhbBrdu93z5Iy3AWIb7vGeTKznnnDweJzYpfHCXuWZdr/d
+z6cbmudPzN1l99Op5eH9i1JikA+DQ8BQv1OgkNBw2A==
+-----END CERTIFICATE-----
+"#;
+
+	let param = CertificateParams::from_ca_cert_pem(CERT_WITH_MULTI_DC).unwrap();
+
+	let domain_component_values = param.distinguished_name.get(&DnType::CustomDnType(vec![
+		0, 9, 2342, 19200300, 100, 1, 25,
+	]));
```
cpu, line 596: Since you have `x509-parser` for this test you can avoid the hardcoded bare OID with a feature-guarded import of `x509_parser::oid_registry::OID_DOMAIN_COMPONENT`:

```rust
	let domain_component_values = param.distinguished_name.get(&DnType::CustomDnType(
		OID_DOMAIN_COMPONENT.iter().unwrap().collect(),
	));
```

rcgen/src/lib.rs
```diff
@@ -776,6 +794,83 @@ mod tests {
 		}
 	}
 
+	#[test]
+	fn distinguished_name_remove_no_match() {
+		let mut dn = DistinguishedName::new();
+		// Domain Component (DC)
+		let dc_type = DnType::CustomDnType(vec![0, 9, 2342, 19200300, 100, 1, 25]);
```
cpu, line 801: It seems unimportant for the test added in this file that the `DnType` be a `Custom` one for DC. I think you could avoid some duplication by using a predefined `DynType`. WDYT?

rcgen/src/lib.rs
```diff
@@ -309,20 +307,22 @@ impl DistinguishedName {
 		Self::default()
 	}
 	/// Obtains the attribute value for the given attribute type
-	pub fn get(&self, ty: &DnType) -> Option<&DnValue> {
-		self.entries.get(ty)
+	pub fn get(&self, ty: &DnType) -> Vec<&DnValue> {
```
cpu, line 310: Isn't this a breaking API change? That conclusion also seems to be supported by the necessity to change the example code on `push()`.

cpu (commented):
Sorry for the delay in taking a look at this. I've wanted to do it with some care and its harder to find time for that vs simpler PRs.  I'm afraid as of now I don't feel comfortable +1ing this branch. 

Firstly, unless I'm mistaken it still seems to be a semver breaking change and I don't think the necessity of the feature justifies that.

Secondly, I think while it's true that some distinguished name types should allow multiple value I don't think it's a universal property. In particular I'm nervous that this will encourage users to try things like adding two common names to the certificate subject instead of using the subject alternative names extension. You haven't offered a strong citation to a standards document that supports the idea that all DN types should allow multiple values and so I think this wariness is justified.

I do agree that for domain components in particular it should be possible to encode multiple values, but I think the API as designed here is too broad a change. For comparsion, see how Go's x509 package's [`pkix.Name.ToRdnSequence()`](https://pkg.go.dev/crypto/x509/pkix#Name.ToRDNSequence) only treats a specific set of dn types as multi-value (and as expected, `OrganizationalUnit` is one of them).

Could we think up a more complex API that works similar to Go's that would be less likely to cause misuse? Probably. Is it worth the effort? I'm not so sure. 

@djc @est31 WDYT?

rcgen/src/lib.rs
djc: Yup, it is -- sorry I missed that.

djc:
> I do agree that for domain components in particular it should be possible to encode multiple values, but I think the API as designed here is too broad a change. For comparsion, see how Go's x509 package's [`pkix.Name.ToRdnSequence()`](https://pkg.go.dev/crypto/x509/pkix#Name.ToRDNSequence) only treats a specific set of dn types as multi-value (and as expected, `OrganizationalUnit` is one of them).
> 
> Could we think up a more complex API that works similar to Go's that would be less likely to cause misuse? Probably. Is it worth the effort? I'm not so sure.
> 
> @djc @est31 WDYT?

"worth the effort" implies a benefit/cost trade-off. I tend to feel that there is a decent amount of value in designing/maintaining careful, misuse-resistant APIs for libraries like this even if the amount of usage is fairly small, because the API itself can help more people guide towards doing the right thing. Realistically, the alternative is that someone faced with this use case will throw together their certs with a less careful API which overall will decrease the quality of certificates in the ecosystem.

(There might be a corollary to Postel's law in here, but I'm not sure I have a pithy formulation of it.)

So that brings us to the cost -- someone has to bear the cost of designing and reviewing such an API, and it's not completely obvious to me who might be interested in and capable of designing it right in this case.

cpu:
> "worth the effort" implies a benefit/cost trade-off. I tend to feel that there is a decent amount of value in designing/maintaining careful, misuse-resistant APIs for libraries like this even if the amount of usage is fairly small

I agree with that in the abstract, but in practice I think there's a fixed amount of time anyone can spend on this project and so we should direct it towards improvements that offer the most bang for the buck.

Putting that aside, is it fair to say we're in agreement the proposed updated API is not a careful, misuse-resistant API, or is there some disagreement there?

> Realistically, the alternative is that someone faced with this use case will throw together their certs with a less careful API which overall will decrease the quality of certificates in the ecosystem.

True, but there are many projects that would allow someone to do this outside of rcgen. My principle concern is not making the outcome impossible, but ensuring that any change we make to support niche use-cases doesn't encourage misuse in the more common circumstances.

> So that brings us to the cost -- someone has to bear the cost of designing and reviewing such an API, and it's not completely obvious to me who might be interested in and capable of designing it right in this case.

Indeed, and I think I'm unlikely to want to spend considerable time on this myself. The lag in review is also good evidence I can't prioritize this in practice.

Maybe we should take a step back and learn more about the use-case that is driving this feature request? In #301 the problem was described as one of extant certs with multiple-DCs not **parsing** correctly, but here we have an API to **produce** such certificates. I feel like I'm missing some details on how we went from A to B.

djc:
> > "worth the effort" implies a benefit/cost trade-off. I tend to feel that there is a decent amount of value in designing/maintaining careful, misuse-resistant APIs for libraries like this even if the amount of usage is fairly small
> 
> I agree with that in the abstract, but in practice I think there's a fixed amount of time anyone can spend on this project and so we should direct it towards improvements that offer the most bang for the buck.
> 
> Putting that aside, is it fair to say we're in agreement the proposed updated API is not a careful, misuse-resistant API, or is there some disagreement there?

Yes, I agree with that.

> Indeed, and I think I'm unlikely to want to spend considerable time on this myself. The lag in review is also good evidence I can't prioritize this in practice.
> 
> Maybe we should take a step back and learn more about the use-case that is driving this feature request? In #301 the problem was described as one of extant certs with multiple-DCs not **parsing** correctly, but here we have an API to **produce** such certificates. I feel like I'm missing some details on how we went from A to B.

That's definitely fair. @da-kami can you give us more details about what you are trying to achieve?

est31:
I'm fine with making the API more complex, especially I prefer explicit complexity over implicit complexity (which is C style: you need to use the "simple" API correctly and respect the implicit invariants). Ideally there would still be a "simple" API for the 90% of users who don't need multiple domain components.

cpu:
Since we haven't heard back from the original author and there seems to be consensus this isn't the right solution I would be in favour of closing it for now. :+1: / :-1:?

cpu:
@thhapke Are you or your organization willing to sponsor development?

thhapke:
I am checking how lasting the multiple-domain approach is going to be. Personally I am on your line to think it is rather exotic .  If there are strong reasons and if we like to use it then I am coming back. Thanks

cpu:
> I am checking how lasting the multiple-domain approach is going to be

Sorry, I'm not sure I'm parsing this question correctly.

AFAIK nobody is working on a fix for this issue. I think doing so would involve a redesign of the way RDNs are handled and I'm personally not interested in taking on that work as a volunteer at this time.

thhapke:
Yep I totally got it. I just wanted to mention that there is another use case in addition to the original one and I fully understood that this was not a commitment.

da-kami:
> Are you or your organization willing to sponsor development of an acceptable fix?

Hey @cpu , sorry for only cycling back to this after significant time.
I think it would still be great to land a change, but I would like to understand better what you have in mind.

I'm not sure how much further I can argue in favor of certificates that have multiple domain components. We have customers that do that, and it's hard to convince them otherwise. So for now I'm forced to use a fork.

If we assume that it would be a more complete implementation, would you accept a change?

On top of that, when it comes to changing this, would you rather have:

1. Funding so you can get one of the maintainers to do it?
    1. Would you have a rough estimate in mind how much (USD) would be needed?
3. Accept a contribution?
    1. If so: I would like to understand better what
    > a redesign of the way RDNs are handled
    
    could look like in your opinion.

Thanks

PR #302, opened 2024-12-19
feat: Add Certificate::from_der + Certificate::from_pem

blacktea1105:
Related to #274 Add Certificate::from_der + Certificate::from_pem.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -1419,7 +1419,7 @@ mod tests {
 			assert_eq!(origin_cert.der(), cert.der());
 		}
 	}
-	
+
 	#[cfg(feature = "pem")]
 	mod test_pem_serialization {
 		use super::*;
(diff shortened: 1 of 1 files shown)


djc:
What is your use case? I wrote in #293:

> If you still believe there is a use case that rcgen doesn't cover, I suggest you open a new issue with a clear explanation of your use case rather than commenting on this confusingly-named closed PR.

And instead you open another PR that's similar to #293 rather than doing as asked.

blacktea1105:
Sorry for the confusion earlier!

I initially thought the leaf certificate invalidity was due to the fingerprint mismatch between the root trust CA certificate in the OS (create and save PEM to file/DB) and the restore/resingned certificate from PEM (after several tests, I found this wasn't the case).

My problem has been resolved (my mistake). Perhaps I should close this PR now?

Thank you very much for your reply and assistance!

PR #303, opened 2024-12-19
rcgen: 0.13.1 -> 0.13.2

lvkv:
Coming from https://github.com/rustls/rcgen/issues/285#issuecomment-2552866476. @djc did you want me to write the changelog as well? Don't want to step on your toes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -879,7 +879,7 @@ dependencies = [
 
 [[package]]
 name = "rcgen"
-version = "0.13.1"
+version = "0.13.2"
 dependencies = [
  "aws-lc-rs",
  "botan",
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rcgen"
-version = "0.13.1"
+version = "0.13.2"
 documentation = "https://docs.rs/rcgen"
 description.workspace = true
 repository.workspace = true


djc (approved):

cpu (approved):
thanks!

est31:
A changelog would be nice, otherwise LGTM

djc:
I'll generate a changelog when I publish.

djc:
@est31 so all the other rustls projects use GitHub releases and their generated changelogs instead of a manually curated `CHANGELOG.md`. I forget how much we have discussed this before, but would you be open to using that? We would pull out manually written high-level release notes for anything of particular interest to users. I would be willing to convert the old `CHANGELOG.md` contents into GitHub releases if that helps.

est31:
Personally I'm not a fan of github releases. they are useful for attaching downloadable build artifacts and advertise the release on github, but they are github proprietary, are paginated on a very small scale, and you need network access to access them. `CHANGELOG.md` on the other hand sits in any clone of the repository and gets shipped to users via cargo vendor.

Doing both has some advantages, but if it is you who are writing the changelogs then I don't want to tell you what to do. Maybe we keep `CHANGELOG.md` and say that any newer release can be found on github releases?

djc:
> Doing both has some advantages, but if it is you who are writing the changelogs then I don't want to tell you what to do. Maybe we keep CHANGELOG.md and say that any newer release can be found on github releases?

Will do, thanks!

djc:
* Published rcgen v0.13.2 at registry `crates-io`
* [new tag]         v0.13.2 -> v0.13.2
* [Release notes](https://github.com/rustls/rcgen/releases/tag/v0.13.2)

PR #305, opened 2024-12-20
Update dependencies

djc:


cpu (approved):

est31:
Behind the "Update semver-compatible dependencies" there is a `cargo update` right?

est31 (approved):

djc:
> Behind the "Update semver-compatible dependencies" there is a `cargo update` right?

Yup, that's all it is!

PR #307, opened 2025-01-04
change signature of signed_by to accept &impl AsRef<CertificateParams> issuer

audunhalland:
I had the issue that I wanted to use an issuer that was a `CertificateParams`, not a `Certificate` (because CertificateParams can be deserialized from file, but Certificate can not). The best solution I've found is to change the signature to accept `&impl AsRef<CertificateParams>`, and then implement that `AsRef` for both `Certificate` and `CertificateParams`.

I think this is also a backwards-compatible change, but I'm not 100% sure.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -60,6 +60,12 @@ impl From<Certificate> for CertificateDer<'static> {
 	}
 }
 
+impl AsRef<CertificateParams> for Certificate {
+	fn as_ref(&self) -> &CertificateParams {
+		&self.params
+	}
+}
+
 /// Parameters used for certificate generation
 #[allow(missing_docs)]
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


djc (approved):

rcgen/src/certificate.rs
```diff
@@ -843,6 +849,12 @@ impl CertificateParams {
 	}
 }
 
+impl AsRef<CertificateParams> for CertificateParams {
```
djc, line 852: Is this necessary? I would kind of expect that there's a blanket impl for this.

djc:
This seems like a nice improvement!

rcgen/src/certificate.rs
audunhalland: Unfortunately, yes: https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity

cpu (approved):
Looks good, thanks!

est31:
The issue with `AsRef` tricks is that it isn't immediately obvious what types implement `AsRef<CertificateParams>`. It would be nice to point it out in the docs that both `Certificate` and `CertificateParams` implement it.

Issue #310, opened 2025-01-17
Subject alternative name extension must be marked critical if the "subject" field is empty


howardjohn:
Per  https://tools.ietf.org/html/rfc5280#section-4.1.2.6 :

> If subject
>    naming information is present only in the subjectAltName extension
>    (e.g., a key bound only to an email address or URI), then the subject
>    name MUST be an empty sequence and the subjectAltName extension MUST
>    be critical.

However, currently rcgen hardcodes SANs as non-critical: https://github.com/rustls/rcgen/blob/cd88a397e4e53b26b19c95bb86e7b3d00f75d412/rcgen/src/certificate.rs#L502

It would be nice to have this either automatically detect empty subject and mark it as critical, or have a way to indicate the extension as critical.

If I understand right, the only way to do this currently would be with a custom extension which seems like a lot of work.

I am willing to work on a fix for this

PR #311, opened 2025-01-17
Fix: mark SAN as critical when subject is empty

howardjohn:
Fixes https://github.com/rustls/rcgen/issues/310

### DIFF
--- a/rcgen/tests/generic.rs
+++ b/rcgen/tests/generic.rs
@@ -518,3 +518,54 @@ mod test_csr {
 		assert_eq!(*params, csrp.params);
 	}
 }
+
+#[cfg(feature = "x509-parser")]
+mod test_subject_alternative_name_criticality {
+	use x509_parser::certificate::X509Certificate;
+	use x509_parser::extensions::X509Extension;
+	use x509_parser::{oid_registry, parse_x509_certificate};
+
+	use crate::util::default_params;
+
+	#[test]
+	fn with_subject_sans_not_critical() {
+		let (params, keypair) = default_params();
+		assert!(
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Seems reasonable to me!

cpu (approved):
Thanks for fixing this. I agree it's the right approach (and [matches Go x509](https://github.com/golang/go/blob/40b3c0e58a0ae8dec4684a009bf3806769e0fc41/src/crypto/x509/x509.go#L1256-L1259)'s behaviour - always reassuring).

Would you mind adding a small test w/ x509-parser to prevent regression? Something like generating two certs, both with SANs, but only one with a subject, and then asserting the criticality of the SAN ext matches expectation would be great.

est31 (approved):

cpu:
> Would you mind adding a small test w/ x509-parser to prevent regression? 

I'll push a test commit for this in a moment. There's no use waiting on a test if it's quick to write.

rcgen/tests/generic.rs
```diff
@@ -518,3 +518,54 @@ mod test_csr {
...
+	fn with_subject_sans_not_critical() {
+		let (params, keypair) = default_params();
+		assert!(
+			!params
+				.distinguished_name
+				.iter()
+				.collect::<Vec<_>>()
+				.is_empty(),
+			"non-empty subject required for test"
+		);
```
cpu, line 540: this is a little awkward, but it didn't seem worth adding API surface for, or moving the test into the same module to get at the `entries`.

PR #321, opened 2025-03-11
Update dependencies

djc:


djc:
I've reverted the botan upgrade (sticking with 0.11.0 instead of moving to 0.11.1) and filed [an issue](https://github.com/randombit/botan-rs/issues/144). For the base64ct failure, I've chosen to remove the `rsa-irc` example that was relying on the rsa crate (which in turn relies on the base64ct crate). This also removes the need for rand, so it seems like a nice simplification.

cpu (approved):
> I've reverted the botan upgrade (sticking with 0.11.0 instead of moving to 0.11.1) and filed https://github.com/randombit/botan-rs/issues/144.

Thank you!

>  For the base64ct failure, I've chosen to remove the rsa-irc example that was relying on the rsa crate (which in turn relies on the base64ct crate). This also removes the need for rand, so it seems like a nice simplification.

I'm a :+1: on that. It feels like this example was more important when there wasn't a way to generate RSA private keys w/ rcgen (because of `ring` limitations). Since that's now handled by `aws-lc-rs` support _and_ the `rsa` crate has unresolved vulnerabilities  (_and RSA is a poor choice in 2025 generally_) it seems like an obvious choice to remove it.

est31 (approved):

Cargo.toml
```diff
@@ -17,9 +16,3 @@ readme = "README.md"
 description = "Rust X.509 certificate generator"
 repository = "https://github.com/rustls/rcgen"
 keywords = ["mkcert", "ca", "certificate"]
-
-# This greatly speeds up rsa key generation times
-# (only applies to the dev-dependency of rcgen because cargo
-# ignores profile overrides for non leaf packages)
-[profile.dev.package.num-bigint-dig]
```
est31, line 24: I suppose removing this is okay because we now have aws-lc-rs which can also do rsa keygen.

Issue #324, opened 2025-03-15
Breaking change in 0.13.3?


ousado:
I've only just seen this issue in rustls-acme after updating dependencies and getting below compilation error. 
It doesn't occur with 0.13.2. I saw the 0.13.3 release is just about an hour old, so I posted this issue before investigating. 

`cargo update -p rcgen --precise 0.13.2` fixes it for indirect dependencies.

```
error[E0277]: the trait bound `Arc<rcgen::Certificate>: AsRef<CertificateParams>` is not satisfied
   --> [...]/cargo_home/registry/src/index.crates.io-6f17d22bba15001f/rustls-acme-0.13.0/src/caches/test.rs:71:48
    |
71  |         let cert = params.signed_by(&key_pair, &self.ca_cert, &self.ca_key).unwrap();
    |                           ---------            ^^^^^^^^^^^^^ the trait `AsRef<CertificateParams>` is not implemented for `Arc<rcgen::Certificate>`
    |                           |
    |                           required by a bound introduced by this call
    |
    = help: the trait `AsRef<rcgen::Certificate>` is implemented for `Arc<rcgen::Certificate>`
    = help: for that trait implementation, expected `rcgen::Certificate`, found `CertificateParams`
note: required by a bound in `CertificateParams::signed_by`
   --> [...]/cargo_home/registry/src/index.crates.io-6f17d22bba15001f/rcgen-0.13.3/src/certificate.rs:161:17
    |
158 |     pub fn signed_by(
    |            --------- required by a bound in this associated function
...
161 |         issuer: &impl AsRef<CertificateParams>,
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `CertificateParams::signed_by`

For more information about this error, try `rustc --explain E0277`.
```

https://github.com/FlorianUekermann/rustls-acme/issues/73

junkurihara:
Hi, I just faced the exactly same error while `AsRef<CertificateParams>` is implemented for `Certificate`. The reason why the compilation error happens is that the `Certificate` is `Arc`-ed.

So my solution is just just deref the object, i.e., `&*self.ca_cert` in the code. It works perfectly at my side.

cpu:
@djc @est31 I think we should yank this release and revisit the change. Agree?

djc:
Yup, can you do it?

cpu:
I'm on it 👍

cpu:
[...]

PR #325, opened 2025-03-20
Revert impl AsRef issuer

audunhalland:
Reverts #307 and #312 because they broke semver (#324) .

I'll make a followup that re-introduces this feature using another set of methods.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -61,6 +61,12 @@ impl From<Certificate> for CertificateDer<'static> {
 	}
 }
 
+impl AsRef<CertificateParams> for Certificate {
+	fn as_ref(&self) -> &CertificateParams {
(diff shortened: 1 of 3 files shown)


cpu (approved):
Thanks for picking this up, much appreciated :-) (and again, sorry we missed it in review :sweat: )

cpu:
Since this is a revert to the original state I'm going to merge without blocking on further reviews.

djc:
I wanted to suggest that it might make sense to scope this down instead of removing all of it? The `AsRef` implementations still make sense to me, we could only change back the argument type for `signed_by()`?

cpu:
> The AsRef implementations still make sense to me, we could only change back the argument type for signed_by()?

:+1:

djc (approved):

PR #329, opened 2025-04-13
Move string types to separate module

est31:
Looking at the rustdoc, it's a bit weird to see the string types mixed up with certificate related stuff.

The rename was done because it follows the way `std` names its modules.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -47,6 +47,8 @@ use yasna::tags::{TAG_BMPSTRING, TAG_TELETEXSTRING, TAG_UNIVERSALSTRING};
 use yasna::DERWriter;
 use yasna::Tag;
 
+use crate::string::{BmpString, Ia5String, PrintableString, TeletexString, UniversalString};
+
 pub use certificate::{
 	date_time_ymd, Attribute, BasicConstraints, Certificate, CertificateParams, CidrSubnet,
 	CustomExtension, DnType, ExtendedKeyUsagePurpose, GeneralSubtree, IsCa, NameConstraints,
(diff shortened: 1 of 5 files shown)


djc (approved):
Nice!

rcgen/src/lib.rs
```diff
@@ -65,7 +67,6 @@ pub use key_pair::{KeyPair, RemoteKeyPair, SubjectPublicKeyInfo};
 use ring_like::digest;
 pub use sign_algo::algo::*;
 pub use sign_algo::SignatureAlgorithm;
-pub use string_types::*;
```
cpu, line 68: Is your intent to make this a semver incompatible change? If so we should add a version bump to [fix the semver check job](https://github.com/rustls/rcgen/actions/runs/14434832706/job/40474197283).

Otherwise, maybe the `use` on L50 should be a `pub use` to continue to allow downstream code to import the string types through the top-level crate namespace?

cpu (commented):
Nice! I like it.

I think there are a couple CI failures to sort out, one of which is a semver compat check issue I left a comment about.

rcgen/src/lib.rs
est31: Yeah my idea was to do it now because we want to break the API anyways. I can bump the version, but let's do a release later only.

cpu:
> cpu force-pushed the string_types_module branch from 85e90a9 to 64279e3 

I fixed a build failure in the version bump commit (`rustls-cert-gen` needed its path import version updated), as well as a doc test failure from the module refactoring (changing `use crate::x` to `use crate::strings::x` throughout).

It seemed easier to fix it & push than to do another back-and-forth round. Hope you don't mind.

cpu (approved):

PR #331, opened 2025-04-22
Deduplicate issuer usage

djc:
Based on:

- #328

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -44,8 +44,8 @@ use time::{OffsetDateTime, Time};
 use yasna::models::ObjectIdentifier;
 use yasna::models::{GeneralizedTime, UTCTime};
 use yasna::tags::{TAG_BMPSTRING, TAG_TELETEXSTRING, TAG_UNIVERSALSTRING};
-use yasna::DERWriter;
 use yasna::Tag;
+use yasna::{DERWriter, DERWriterSeq};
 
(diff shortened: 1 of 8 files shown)


rcgen/src/lib.rs
```diff
@@ -139,6 +139,42 @@ struct Issuer<'a, S> {
...
+		Self {
+			distinguished_name: &params.distinguished_name,
+			key_identifier_method: &params.key_identifier_method,
+			key_usages: &params.key_usages,
+			key_pair,
+		}
+	}
+}
+
+trait ToDer {
```
est31, line 153: as I've said previously, I'd prefer a better name for that trait: `SignDer` or something like it.

djc: Yup. Do you think the introduction of the trait carries its weight here? I'm not totally convinced...

est31: not really tbh. I think having consistent inherent functions is better (traits need to be imported to be used/called, and even if you add inherent functions on top, they still add complexity, which needs to be justified, and in this case i think it's not).

cpu:
> djc marked this pull request as draft 3 days ago

Did you want me to hold off reviewing this for now?

djc:
> > djc marked this pull request as draft 3 days ago
> 
> Did you want me to hold off reviewing this for now?

Yup!

PR #333, opened 2025-04-25
Extract write_extensions() method, reducing rightward drift

djc:


est31 (approved):
in such cases it's best to review with the "[hide whitespace](https://github.com/rustls/rcgen/pull/333/files?diff=split&w=1)" option disabled

cpu (approved):

PR #336, opened 2025-05-01
Distribute methods for parsing params elements from x509

djc:
This takes the processing steps for transforming an `x509_parser::Certificate` into `CertificateParams` and distributes them over the relevant element's types, in preparation for splitting `from_ca_cert_pem()` and `from_ca_cert_der()` from `CertificateParams` (to make it statically obvious that their use is only intended to build an `Issuer`).

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -207,7 +207,7 @@ impl CertificateParams {
 		let validity = x509.validity();
 		let subject_alt_names = SanType::from_x509(&x509)?;
 		let key_usages = KeyUsagePurpose::from_x509(&x509)?;
-		let extended_key_usages = Self::convert_x509_extended_key_usages(&x509)?;
+		let extended_key_usages = ExtendedKeyUsagePurpose::from_x509(&x509)?;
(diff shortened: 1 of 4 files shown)


cpu (approved):
Ver nice! Thank you.

I think the coverage CI failures are fair to ignore since they're pre-existing coverage gaps.

est31 (approved):
makes sense but can you squash?

djc:
> makes sense but can you squash?

Why? All of these commits compile and could likely pass CI on their own.

est31:
I feel it's too fine. I think it should be easy to squash the "Move ..." commits together, they are form an independent unit compared to the other commits.

djc:
> I feel it's too fine. I think it should be easy to squash the "Move ..." commits together, they are form an independent unit compared to the other commits.

What are the downsides of merging smaller commits? If there's ever a regression or you want to sift through `git blame`, smaller commits will definitely be easier to review.

est31:
On the other hand, it's harder to understand the motivation of a bigger commit compared to a small one. But whatever, it's bikeshedding. Let's just merge this one.

Issue #340, opened 2025-05-14
Consider eagerly implementing common traits


lvkv:
My immediate need/use case is `Clone` on `CertificateSigningRequest`. After skimming the codebase, however, I'm getting the impression that there are many more opportunities to implement the common traits outlined in the [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits). @est31 also echoed this over at https://github.com/rustls/rcgen/issues/17#issuecomment-503656409.

Prior related work:
- https://github.com/rustls/rcgen/issues/315 /  https://github.com/rustls/rcgen/pull/316
- https://github.com/rustls/rcgen/pull/319
- and more, probably...

If you're amenable to eagerly implementing these traits, I think it'd be easiest (from a review standpoint) to create a pull request per trait and dedicate each PR to sweeping the codebase and nitpicking the details—similar to https://github.com/rustls/rcgen/pull/316. I'm open to other approaches too.

cpu:
> After skimming the codebase, however, I'm getting the impression that there are many more opportunities to implement the common traits outlined in the [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

Agreed! I think some of these were harder to take back when some of the params types were holding private data but we're in a better spot now.

> I think it'd be easiest (from a review standpoint) to create a pull request per trait and dedicate each PR to sweeping the codebase and nitpicking the details

I'd also be open to one PR with each trait handled per-commit but don't feel strongly.

djc:
One trait per commit in a single PR sounds good to me.

lvkv:
The test run of #341 was pretty seamless, so I'll open one more PR for all the other common traits—one trait per commit.

djc:
I think we should stop after the PRs that already exist and wait for someone who has a particular impl they require.

lvkv:
> I think we should stop after the PRs that already exist

Feel free to manage/close this PR whenever, by the way. There are no further requests on my end after https://github.com/rustls/rcgen/pull/341 🙂.

PR #341, opened 2025-05-15
Eagerly derive Clone, Copy, where possible

lvkv:
Coming from https://github.com/rustls/rcgen/issues/340, this PR derives `Clone` and `Copy` to align rcgen more closely with Rust's [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits). Here's how I went about this:

1. Derive `Clone` for as many types as possible in a full codebase pass (finding types by basically grepping for `struct ` and `enum `). Keep making passes until no more types can derive it.
2. Ditto for `Copy`

I've intentionally opted to derive these traits eagerly at first so that types that we _don't_ want to derive `Clone` or `Copy` for will come up in this review. I think most of these make sense, but I could see some being not desirable—e.g., `Copy` for `SignatureAlgorithmParams`.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -968,7 +968,7 @@ impl GeneralSubtree {
 	}
 }
 
-#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[allow(missing_docs)]
 /// CIDR subnet, as per [RFC 4632](https://tools.ietf.org/html/rfc4632)
 ///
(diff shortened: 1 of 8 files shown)


cpu:
> ci / build-windows (pull_request)Started 10 minutes ago — This check has started...

FWIW I expect this to fail, it looks [broken on `main`](https://github.com/rustls/rcgen/actions/runs/15052227395/job/42309602978). Feel free to disregard for this branch.

djc (approved):
This broadly looks good to me. Nit: I'd like the new additions to roughly appear in alphabetical order.

lvkv:
> Nit: I'd like the new additions to roughly appear in alphabetical order.

Sure! I'll rework this while still keeping separate commits for `Clone` and `Copy`.

lvkv:
@djc done—feel free to nit further, of course

cpu (approved):
Nice :-) Thank you

est31 (approved):

est31:
Windows CI is broken but it doesn't seem to be fault of this patch.

PR #342, opened 2025-05-16
Updated `.gitignore` to be more specific

Rynibami:
Hi, I want to improve the `rcgen` crate and noticed that certain items were not included into the `.gitignore`. So I thought updating this file would be a great start of introducing myself and start working on the project (if you would have me). 😄


Rynibami:
Yes, I'm building on MacOS with VSCode (and have used Rust Rover in the past). So there are a few files / directories I like to mask out for git.

djc:
Okay, can you get rid of the three-line section headers and trim this to just the stuff that you need today?

.gitignore
```diff
@@ -1,4 +1,26 @@
...
+.history/
+.vscode/
+!.vscode/settings.json
+!.vscode/tasks.json
+
+# Rust and Cargo
 **/*.rs.bk
-certs/
-.idea
+
```
djc, line 21: I think `/target` should be part of "Rust and Cargo".

.gitignore
```diff
@@ -1,4 +1,26 @@
...
+# Output
+/target
+
+# Secret
+certs/
```
djc, line 26: Nit: can you add a trailing newline?

est31:
I don't think we should be adding such extensive changes. editor specific files, backup files, etc should go into the user-local gitignore, not into each project's gitignore. project gitignores should get stuff that is either specific to the language/technology (like build artifact dir) or to the individual project (say a dir for generated certificates, if the tool generates certificates).

est31:
From https://git-scm.com/docs/gitignore

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig.

Note that in the past I've also done this extend gitignore game, but eventually I realized that user-specific gitignores are better for editor specific stuff.

Rynibami:
> I don't think we should be adding such extensive changes. editor specific files, backup files, etc should go into the user-local gitignore, not into each project's gitignore. project gitignores should get stuff that is either specific to the language/technology (like build artifact dir) or to the individual project (say a dir for generated certificates, if the tool generates certificates).

@est31, can you specify what items you would like to update / remove? Following your strategy, the original file would also remove the `.idea` rule.

est31:
yeah probably we should also remove `.idea` as well, and call it done. The comments are also useful. Outside of that I don't see a need to change `.gitignore`.

Rynibami:
@est31, I'll return the `.gitignore` to what it mostly was (and keep some comments). Shall I put a small description (and reference to this PR as well) that describes / explains this choice? This will prevent future discussions around this topic.

est31 (approved):
LGTM, thanks for filing!

PR #343, opened 2025-05-16
Eagerly implemented `Debug` trait

Rynibami:
Issue #340 discussed implementing the most common traits for this crate. I implemented the `Debug` trait for all possible `stuct`s and `enum`s in this PR.

I also came across the `CertifiedKey` and `Issuer` types that didn't derive the `Clone` trait, but are able to. Should this be addressed as well, or should we keep it that way?


rcgen/src/lib.rs
```diff
@@ -87,6 +87,7 @@ pub mod string;
 pub type RcgenError = Error;
 
 /// An issued certificate, together with the subject keypair.
+#[derive(Debug)]
 pub struct CertifiedKey<S: SigningKey> {
```
cpu, line 91: We wouldn't want to derive `Debug` for `CertifiedKey` or `Issuer` on L136 as both contain private key data.

cpu:
> I also came across the CertifiedKey and Issuer types that didn't derive the Clone trait, but are able to. Should this be addressed as well, or should we keep it that way?

For the same reason I flagged these as not being appropriate for `Debug` they aren't appropriate for `Clone` and should be left as-is.

Rynibami:
I'll update them and push the next commit online :)

djc (approved):
This seems okay to me.

I'd prefer to avoid implementing `Clone` on types that include a private key.

rcgen/src/lib.rs
djc: For Issuer I don't think it matters? The responsibility for not leaking private data in a SigningKey impl is up to the implementer.

cpu: That seems reasonable :+1:

Rynibami: Okay, then I'll enable debug again on these structs.

est31: > The responsibility for not leaking private data in a SigningKey impl is up to the implementer.

it is, but the stakes are too high. At work we did leak secrets into the debug impl once and put it into the logs. it's not beautiful.

let's not implement Debug here, or if we do, via a manual impl that has `<redacted>` or something for the `key_pair`.

est31 (changes requested):
Debug impl of `CertifiedKey`

rcgen/src/lib.rs
Rynibami: All right, I'll remove it again.

rcgen/src/lib.rs
```diff
@@ -132,6 +132,7 @@ pub fn generate_simple_self_signed(
 	Ok(CertifiedKey { cert, key_pair })
 }
 
+#[derive(Debug)]
```
est31, line 135: also please remove this one

rustls-cert-gen/src/cert.rs
```diff
@@ -146,6 +147,7 @@ impl Ca {
 }
 
 /// End-entity [Certificate]
+#[derive(Debug)]
```
est31, line 150: here too

rustls-cert-gen/src/cert.rs
```diff
@@ -124,6 +124,7 @@ impl CaBuilder {
...
```
est31, line 127: this too

est31 (changes requested):
I think instead of removing the `Debug` impl, it would be better to manually implement it. that way, we don't add it back by accident. can be a followup though.

Rynibami:
> I think instead of removing the `Debug` impl, it would be better to manually implement it. that way, we don't add it back by accident. can be a followup though.

I can remove the implementations and leave it at that... but that just doesn't sit right to leave it at that. Maybe it is a great idea to create a proc macro crate that implements derive macros. Then we could derive a similar macro that actually implements the basic traits in a certain manner. For example:

```rs
#[derive(DebugLimited)]
struct Issuer<'a, S> {
    distinguished_name: &'a DistinguishedName,
    key_identifier_method: &'a KeyIdMethod,
    key_usages: &'a [KeyUsagePurpose],
    #[debug_remove] // or #[debug_redact] to show the field but not the value
    key_pair: &'a S,
}
``` 

This would (in this case) implement the `Debug` trait normally as derive trait, but would change the actual implementation where `*_remove` would remove the whole field and `*_redact` replaces the field value with `*** REDACTED ***` (or something similar). This pattern can the also be used to extend other limited derive implementations.

djc:
The costs of adding yet another derive macro are substantial. You could just do a manual Debug implementation instead, it's fine.

Rynibami:
> The costs of adding yet another derive macro are substantial. You could just do a manual Debug implementation instead, it's fine.

In that case I would like to know what fields / types you would like me to redact / remove.

est31:
Only remove `key_pair` please. It's not hard to manually implement Debug, see the [second example](https://doc.rust-lang.org/stable/std/fmt/trait.Debug.html#examples) in the libstd rustdocs. Maybe do a destructuring exhaustive pattern match of the struct, i.e. `let Issuer { field_a, field_b, field_c: _ } = ...`, so that if a field gets added they must also update that function (and add it to the debug impl). plus a comment explaining the choice.

djc (approved):

est31 (approved):
thanks!

cpu (approved):

PR #344, opened 2025-05-16
Eagerly implemented `PartialEq` and `Eq` traits

Rynibami:
Issue #340 discussed implementing the most common traits for this crate. I implemented the `PartialEq` and `Eq` traits for all possible `stuct`s and `enum`s in this PR. I included the derive for private key data as I thought that this would not leak any secret information. Please correct me if I'm wrong.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -22,7 +22,7 @@ use crate::{
 };
 
 /// An issued certificate
-#[derive(Debug, Clone)]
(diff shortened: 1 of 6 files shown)


est31 (approved):

djc:
I don't think we should implement these in rustls-cert-gen which I don't consider as having a public Rust API anyway.

djc (approved):

PR #346, opened 2025-05-18
Eagerly implemented `PartialOrd` and `Ord` traits

Rynibami:
Issue #340 discussed implementing the most common traits for this crate. I implemented the `PartialOrd` and `Ord` traits for all stucts and enums which I thought needed the implementation.


djc:
I don't think any of these have an intrinsic ordering.

PR #348, opened 2025-05-20
Minor tweaks to Debug impls and other style improvements

djc:
Follow up on #343.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -153,10 +153,10 @@ impl CertificateParams {
 	///
 	/// The returned [`Certificate`] may be serialized using [`Certificate::der`] and
 	/// [`Certificate::pem`].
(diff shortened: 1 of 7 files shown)


djc:
> I don't like the strong use of `Self` here: let's not change that: if you don't use an IDE, you can quickly find where some type is constructed or used, but if it's mentioned via `Self`, that's not possible. The other changes are fine.

Dropped the commit for this.

cpu (approved):

cpu:
>  if you don't use an IDE

This doesn't seem like a category of contributor/user worth optimizing for IMO.

est31:
> This doesn't seem like a category of contributor/user worth optimizing for IMO.

even if you do use an IDE, not every time you interact you have an IDE around, say when reviewing PRs on github for example, or when searching on github search. And IDE features are not omnipresent/omniscient. e.g. macros come in mind, or stuff like `#[cfg(test)]` code.

PR #359, opened 2025-06-30
Apply suggestions from clippy 1.88

djc:


djc:
Since this is super boring, merging with one approval.

PR #361, opened 2025-07-02
Check MSRV in CI

djc:
After upgrading some downstream packages I noticed that I introduced an `#[expect]` attribute, which substantially increased the MSRV for what I would describe as pretty minor developer experience gain.

# Proposed release notes

- Declare 1.71 `rust-version` and check MSRV in CI.


cpu (approved):
Oops :face_exhaling: Good catch.

est31 (approved):

djc:
* Published rcgen v0.14.1 at registry `crates-io`
* [new tag]         v0.14.1 -> v0.14.1
* [Release notes](https://github.com/rustls/rcgen/releases/tag/v0.14.1)

PR #362, opened 2025-07-03
Provide a non-owning constructor for `Issuer`

p-avital:
Hey `rcgen` folks!

I must say I really like what you've done with the new API changes, but I'm really missing a way to sign certificates without giving the issuer ownership of the `KeyPair` (especially frustrating since it's not `Clone`).

Maybe I missed an alternative (maybe there's a newtype around a reference that directly impls `SigningKey`), but this seems like a straight forward enough change.

Do keep in mind that if you accept this PR, you're probably locking yourself into having these `MaybeOwned` in the internal type representation (just thought I'd warn you :) ).

Note: feel free to reject this PR if you don't want that extra surface, I'm gonna newtype `&KeyPair` to get that copy-less use in my project :)

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -158,7 +158,11 @@ impl<'a, S: SigningKey> Issuer<'a, S> {
 		}
 	}
 
-	fn from_params(params: &'a CertificateParams, signing_key: &'a S) -> Self {
+	/// Create a new issuer from the given parameters and signing key references.
+	///
+	/// Use [`Issuer::new`] instead if you want to obtain an [`Issuer`] that owns
+	/// its parameters.
+	pub fn from_params(params: &'a CertificateParams, signing_key: &'a S) -> Self {
 		Self {
 			distinguished_name: Cow::Borrowed(&params.distinguished_name),
 			key_identifier_method: Cow::Borrowed(&params.key_identifier_method),


rcgen/src/lib.rs
```diff
@@ -158,6 +158,14 @@ impl<'a, S: SigningKey> Issuer<'a, S> {
 		}
 	}
 
+	/// Create a new issuer fromthe given parameters and signing key references.
+	///
+	/// Use [`Issuer::new`] instead if you want to obtain an [`Issuer`] with a static
+	/// lifetime.
+	pub fn from_refs(params: &'a CertificateParams, signing_key: &'a S) -> Self {
```
djc, line 165: I think it could make sense to expose this, but I don't think it's useful to wrap `from_params()` when we can just expose it directly? Also wondering about the best name, I don't like `from_refs()` much, I think `from_params()` is probably okay.

rcgen/src/lib.rs
```diff
@@ -158,6 +158,14 @@ impl<'a, S: SigningKey> Issuer<'a, S> {
 		}
 	}
 
+	/// Create a new issuer fromthe given parameters and signing key references.
```
djc, line 161: Nit: "fromthe".

djc:
(You might also be interested in https://github.com/rustls/rcgen/pull/363.)

p-avital:
Gah!
I have no way to do that without from GH (I didn't actually download the repo for such a simple PR) D:
Any chance you can use the `Squash & Merge` option on GH? You should have it if you unfold the merge button :)

djc:
> Gah! I have no way to do that without from GH (I didn't actually download the repo for such a simple PR) D: Any chance you can use the `Squash & Merge` option on GH? You should have it if you unfold the merge button :)

We normally use the merge queu, which doesn't make this easy, but we can bypass that if squashing is too hard for you.

p-avital:
Aaaand squashed :)

cpu (approved):
Thank you :-)

PR #366, opened 2025-07-11
Allow splitting CertifiedIssuer into its parts without cloning

davids-work:
This somewhat improves ergonomics for the case where you want to allow alternatively loading the issuer certificate from a file, or generating it when it doesn't exist.

Consider this function:
```rust
fn load_or_generate_ca_certificate(
    ca_certificate_path: &Path,
    ca_keypair_path: &Path
) -> Result<(CertificateDer<'static>, Issuer<'static, KeyPair>)> {
    if ca_certificate_path.exists() && ca_keypair_path.exists() {
        load_ca_certificate(&ca_certificate_path, &ca_keypair_path)
    } else {
        let ca_certificate: CertifiedIssuer<'static, KeyPair> = generate_ca_certificate()?;
        write_ca_certificate(ca_certificate.as_ref(), &ca_certificate_path)?;
        write_ca_keypair(ca_certificate.key(), &ca_keypair_path)?;
        // What now? You can get an Issuer via deref(), but then you have to clone it
       // to get an owned version, and you can get get the certificate via as_ref(), but same here
    }
}
```
`load_ca_certificate` can't return `CertifiedIssuer` (or `rcgen::Certificate`) since that's not possible when loading from a file.

https://github.com/davids-work/kypare/commit/305d95ef985a2c7af1971efc86a85b78edb18762 shows how the addition of `into_issuer_and_certificate` makes this possible without excessive cloning.

### DIFF
--- a/rcgen/src/lib.rs
+++ b/rcgen/src/lib.rs
@@ -177,6 +177,11 @@ impl<'a, S: SigningKey> CertifiedIssuer<'a, S> {
 	pub fn der(&self) -> &CertificateDer<'static> {
 		self.certificate.der()
 	}
+
+	/// Converts the `CertifiedIssuer` into its inner `Issuer` and `Certificate`
+	pub fn into_issuer_and_certificate(self) -> (Issuer<'a, S>, Certificate) {
+		(self.issuer, self.certificate)
+	}
 }
 
 impl<'a, S> Deref for CertifiedIssuer<'a, S> {


djc:
Why do you want to yield `CertifiedIssuer` in this case rather than just building separate `Certificate` and `Issuer` values?

davids-work:
I guess the difference isn't that great, and when you look at how it works internally this more or less the same thing as your suggestion. From a mental model perspective, I think of a "CA Certificate, with its associated private key", which are concrete things which can map to e.g. files on a disk, while an `Issuer` here represents a more abstract concept. Thus it felt more natural to have them together in one type as long as possible, "the certificate" and "the key".

On a tangent, what would really improve ergonomics for my use case would be the possibility to get an rcgen::Certificate from a PEM/DER file, but it doesn't seem that easy to implement with the current design.

davids-work:
> I think you can avoid this by using Issuer::from_ca_cert_der() instead?

Sure, but then you effectively parse PEM -> DER -> Issuer, i.e. between two serialized formats and into one parsed format.

PR #374, opened 2025-09-02
Add unstable support for ML-DSA algorithms

djc:
Would like to use this for testing in rustls.


djc:
> I think the CI failures need addressing as well.

This seems to be another annoying feature unification error... Not sure how to fix this yet.

rustls-cert-gen/src/cert.rs
```diff
@@ -288,10 +230,62 @@ impl fmt::Display for KeyPairAlgorithm {
...
+}
+
+impl FromStr for KeyPairAlgorithm {
+	type Err = anyhow::Error;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		match s {
+			"rsa" => Ok(Self::Rsa),
+			"ed25519" => Ok(Self::Ed25519),
+			"ecdsa-p256" => Ok(Self::EcdsaP256),
```
est31, line 250: is this naming based on some standard? how supported/stable is this format going to be, ie when are we going to break it, etc.

djc: This is actually the current format, "invented" by the `Bpaf` derive macro, made explicit.

rcgen/src/sign_algo.rs
```diff
@@ -209,6 +215,36 @@ pub(crate) mod algo {
 		oid_components: &[1, 3, 101, 112],
 		params: SignatureAlgorithmParams::None,
 	};
+
+	/// ML-DSA-44 signing as per <https://www.ietf.org/archive/id/draft-ietf-lamps-dilithium-certificates-07.html#name-identifiers>.
+	#[cfg(all(feature = "aws_lc_rs_unstable", not(feature = "fips")))]
+	pub static PKCS_ML_DSA_44: SignatureAlgorithm = SignatureAlgorithm {
+		oids_sign_alg: &[&[2, 16, 840, 1, 101, 3, 4, 3, 17]],
```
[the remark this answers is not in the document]
djc: Yup.

Issue #375, opened 2025-09-04
`CertifiedIssuer::from_ca_cert_{pem,der}`?


cyqsimon:
While upgrading from 0.13 to 0.14, I ran into some difficulties trying to keep the old functionality. Specifically, I would like to have access to the issuer certificate while generating a leaf certificate.

Scouring the changelogs and documentation, I did find `CertifiedIssuer` which seems to be what I need. Unfortunately I couldn't find a way to load an existing certificate like you can with `Issuer` (`Issuer::from_ca_cert_{pem,der}`). So here's the feature request.

---

Here's a slimmed-down version of my current code using rcgen-0.13:

```rust
fn main() -> Result<()> {
    let issuer = load_issuer("path/to/issuer.crt", "path/to/issuer.key")?;
    let leaf = generate(&issuer)?;
    dbg!(leaf);
}

fn load_issuer(cert_pem: &str, key_pem: &str) -> Result<CertifiedKey> {
    let key_pair = KeyPair::from_pem(key_pem)?;
    let cert = {
        let mut params = CertificateParams::from_ca_cert_pem(cert_pem)?;
        // https://github.com/rustls/rcgen/issues/274#issuecomment-2121969453
        params.self_signed(&key_pair)?
    };
    // various checks omitted here ...
    Ok(CertifiedKey { cert, key_pair })
}

pub fn generate(issuer: &CertifiedKey) -> Result<CertifiedKey> {
    let params = {
        let issuer_params = issuer.cert.params();
        let mut params = CertificateParams::new([])?;
        // set certificate expiry to match issuer
        params.not_before = issuer_params.not_before;
        params.not_after = issuer_params.not_after;
        // set other certificate params ...
        params
    };
    let key_pair = KeyPair::generate_for(&PKCS_ED25519)?;
    let cert = params.signed_by(&key_pair, &issuer.cert, &issuer.key_pair)?;
    Ok(CertifiedKey { cert, key_pair })
}
```

I would like to update `load_issuer` to return a `CertifiedIssuer` (and ideally get rid of the self-signing hack), so that I can read from the issuer certificate in `generate`. But without `CertifiedIssuer::from_ca_cert_{pem,der}` I don't think it's possible.

cyqsimon:
> I think this runs afoul of exactly the hole that the 0.14 API is trying to close. From the 0.13 documentation:
> 
[...]

Issue #354, opened 2025-06-14
How to create a certificate signed by an existed CA certificate?


xuxinhang:
I try to create a certificate signed by another CA certificate file. I suppose [`CertificateParams.signed_by`](https://docs.rs/rcgen/0.13.2/rcgen/struct.CertificateParams.html#method.signed_by) looks like what I need:

```rust
pub fn signed_by(
  self,
  public_key: &impl PublicKeyData,
  issuer: &Certificate,
  issuer_key: &KeyPair,
) -> Result<Certificate, Error>
```

But, I'm confused on how to get `issuer` from an existed CA pem file on my disk.

Or, is there any other way to create a certificate signed by existed CA certificate?

Thanks.

Issue #268, opened 2024-04-17
Signing leaf certificates with parsed intermediate


pfzetto:
Hello,
I'm currently building a small PoC using this Library and stumbled upon the following error:

I have generated three Certificates (Root, Intermediate,Leave) and key pairs.
The blockage reveals itself when I try to load the Intermediate Certificate.
From the public documentation I learned that I can load a certificate by deserializing the Certificate using `CertificateParams::from_ca_cert_pem` and signing it using `CertificateParams::self_signed` or `CertificateParams::signed_by`.
In my case self signing the certificate would break the chain and I don't have access to the root private key to sign the intermediary on every restart of the server.
In the Issues I've learned that there was a `Certificate::from_pem` that should satisfy my case but got it removed.

How can I load an existing CA Certificate without 'resigning' it?

Thanks
Paul

pfzetto:
Thanks, the "hack" works. I've only needed to manually modify the pem of the leaf certificate to include the original pem of the intermediary to complete the chain.

PR #378, opened 2025-09-10
examples: add writing and reading ca files

iamjpotts:
Adds an example based on `0.14` that generates a certificate chain, and stores each step as PEM files, and reads from the PEM files for the next step, rather than doing all of the steps in memory.

Related:
* https://github.com/rustls/rcgen/issues/268
* https://github.com/rustls/rcgen/issues/274
* https://github.com/rustls/rcgen/issues/354
* https://github.com/rustls/rcgen/issues/375
* https://github.com/rustls/rcgen/pull/99

@oscartbeaumont said he "wanted to know if there is a way to load an existing CA certificate to sign some more certificates without triggering signing."

@acul009 mentioned the api obscurity as discouraging an upgrade from `0.12` to `0.13` in https://github.com/rustls/rcgen/issues/274#issuecomment-2475186136.

This new example does not resign any certificates, instead loading the already signed certificate as an `Issuer`.

### DIFF
--- a/rcgen/Cargo.toml
+++ b/rcgen/Cargo.toml
@@ -32,6 +32,10 @@ zeroize = { workspace = true, optional = true }
 openssl = { workspace = true }
 
 [[example]]
+name = "files"
+required-features = ["pem", "x509-parser"]
+
+[[example]]
 name = "rsa-irc-openssl"
 required-features = ["pem"]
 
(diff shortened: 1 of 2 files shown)


iamjpotts:
@djc see https://github.com/rustls/rcgen/pull/379 as an alternative example to this example.

PR #381, opened 2025-09-26
Group imports

iamjpotts:
Enable nightly formatting option `group_imports = "StdExternalCrate"`.

https://github.com/rustls/rcgen/pull/379#discussion_r2371377127


iamjpotts:
It is not clear to me why the github ui is expecting a `Format & clippy` job. That job was split into two - a `Clippy` job and a `Format (unstable)` job.

.rustfmt.toml
```diff
@@ -1,2 +1,5 @@
 hard_tabs = true
 match_block_trailing_comma = true
+
+unstable_features = true
```
djc, line 4: I don't think we should do it this way -- instead, refer to the rustls workflow I linked to in your other PR which has two separate rustfmt configs and makes sure that the default config can still run with stable rustfmt.

iamjpotts: Done.

I've also added `imports_granularity = "Module"` since that was in the linked workflow.

> It is not clear to me why the github ui is expecting a Format & clippy job.

This also appears resolved (note that I also renamed the workflow job from `fmt` to `fmt-unstable`)

djc (approved):

PR #385, opened 2025-09-30
fix(features): enable x509-parser/verify-aws if aws_lc_rs is selected

jalil-salame:
Otherwise we pull in `ring` and `aws-lc-rs`.

This is broken until a new release of x509-parser with this fix: https://github.com/rusticata/x509-parser/commit/b7dcc9397b596cf9fa3df65115c3f405f1748b2a


cpu:
I suspect this fell off everyone's radar. Let's continue with https://github.com/rustls/rcgen/pull/398 which builds clean.

PR #386, opened 2025-10-05
Expand rustdoc for CertificateSigningRequestParams::from_der

dwhjames:
- document that verification of the signature within the CSR occurs
- fix documentation around what is supported within the CSR


cpu (approved):
Thanks, looks good to me :+1:

djc (approved):

PR #388, opened 2025-10-09
Revise errors for CertificateSigningRequestParams::from_der

dwhjames:
- Introduce specific error for CSR signature verification
- Make error name more specific for unsupported CSR extensions

---

Following up from
- https://github.com/rustls/rcgen/pull/386


djc:
> It sounds like you agree on 1? Do you also agree on 2? Are you open to a more specific name for the rejection of an extension in a CSR?

I'm on the fence because I think the documentation for the variant and error string (`Display` representation) are pretty clear that it's about the CSR -- therefore I'm not sure changing this to a different name (in a semver-compatible release, deprecating the old variant) is worth it. @cpu thoughts?

PR #394, opened 2025-11-17
Add support for multiple values of the same name type

### LINKED ISSUE #81: Support multiple DnValue per key in DistinguishedName
I need to generate a certificate with a subject line containing multiple OU key-value pairs:

```
Subject: OU=certtype:instance, OU=compartment:compartment, OU=instance:instance
```

The problem is subsequent calls to `DistinguishedName::push` overwrite the previous value:

```
params.distinguished_name.push(DnType::OrganizationalUnitName, format!("certtype:{}", certtype));
params.distinguished_name.push(DnType::OrganizationalUnitName, format!("compartment:{}", compartment));
params.distinguished_name.push(DnType::OrganizationalUnitName, format!("instance:{}", instance));
```

This leaves me with a subject containing only the last value:

```
Subject: OU=instance:instance
```

djc:
Fixes #81. This is a semver-incompatible change, so will need a version bump.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -1493,4 +1493,43 @@ PITGdT9dgN88nHPCle0B1+OY+OZ5
 			);
 		}
 	}
+
+	#[cfg(feature = "x509-parser")]
+	#[test]
+	fn test_certificate_with_multiple_domain_components_roundtrip() {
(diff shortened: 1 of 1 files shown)


djc:
@cpu want to take a look?

cpu:
> @cpu want to take a look?

Yes, thanks for the ping. I will look in the next day or so.

PR #403, opened 2025-12-24
Implement `From<KeyPair> for PrivateKeyDer<'static>`

LebedevRI:
I've stumbled into this when trying to write a simple rustls-driven server with self-signed cert.

The "obvious" and safe way to go from `KeyPair` to `PrivateKeyDer` seems to round-tripping via PEM:
```
rustls::pki_types::PrivateKeyDer::from_pem_slice(
            signing_key.serialize_pem().as_bytes(),
        )
```

But `rcgen::key_pair::KeyPair::serialized_der()` is
```
/// Returns a reference to the serialized key pair (including the private key)
/// in PKCS#8 format in DER
```
and `PrivateKeyDer` can be constructed from such a input.
This avoids memory allocation and PEM roundtripping,
both of which are optional features anyways,
and this conversion is non-failing.

Perhaps it makes sense to provide this QOL interface?

### DIFF
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -595,6 +595,20 @@ impl TryFrom<&PrivateKeyDer<'_>> for KeyPair {
 	}
 }
 
+#[cfg(feature = "crypto")]
+impl From<KeyPair> for PrivatePkcs8KeyDer<'static> {
+	fn from(val: KeyPair) -> Self {
+		val.serialize_der().into()
+	}
+}
+
+#[cfg(feature = "crypto")]
+impl From<KeyPair> for PrivateKeyDer<'static> {
+	fn from(val: KeyPair) -> Self {
+		Self::from(PrivatePkcs8KeyDer::from(val))
+	}
+}
+
 /// The key size used for RSA key generation
 #[cfg(all(feature = "crypto", feature = "aws_lc_rs"))]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]


rcgen/src/key_pair.rs
```diff
@@ -595,6 +595,21 @@ impl TryFrom<&PrivateKeyDer<'_>> for KeyPair {
...
+	fn from(val: KeyPair) -> Self {
+		val.serialize_der().into()
+	}
+}
+
+#[cfg(feature = "crypto")]
+impl From<KeyPair> for PrivateKeyDer<'static> {
+	fn from(val: KeyPair) -> Self {
+		let der: PrivatePkcs8KeyDer = val.into();
+		der.into()
```
djc: This is a bit too much indirection for my taste. Does `Self::from(val.into())` work? Otherwise `Self::from(PrivatePkcs8KeyDer::from(val))`?

LebedevRI: Let me clarify.
1. We start with `KeyPair`.
2. `PrivateKeyDer` can be created from `PrivatePkcs8KeyDer`.
3. There are two `From` here to allow conversion to either one of the `Private*KeyDer`
4. The conversion to `PrivateKeyDer` defers to conversion to `PrivatePkcs8KeyDer` in order to avoid code duplication, and to ensure that both of them are covered by the single test.

Do you want me to just inline `From<KeyPair> for PrivatePkcs8KeyDer<'static>`
into `From<KeyPair> for PrivateKeyDer<'static>`,
or drop `From<KeyPair> for PrivatePkcs8KeyDer<'static>` altogether?

LebedevRI:
@djc thank you for taking a look!

rcgen/src/key_pair.rs
djc: I understand what it is doing. I just stylistically dislike type annotations, so I'd prefer to avoid the `into()` that needs a type annotation.

cpu (approved):

PR #405, opened 2026-01-13
Enable certificate signatures using ECDSA with the secp256k1 curve

mkmks:
This PR adds ECSDA with secp256k1 to the list of allowed certificate signature algorithms. The use case for it is explained in https://github.com/rustls/pki-types/pull/96.

While it was pointed out in the discussion under the related https://github.com/rustls/webpki/pull/427 PR that the secp256k1 curve isn't mentioned in any standardised TLS version, `rcgen` is meant to be a certificate generation library agnostic to the use of certificates generated with it. X.509 certificates aren't only used in TLS, so, perhaps, the same argument won't hold here.

For comparison, OpenSSL and its forks do support signing certificates with ECDSA using the secp256k1 curve. It'd be nice to be able to do the same with `rcgen`. 

This PR also happens to fix https://github.com/rustls/rcgen/issues/400 (since it requires the `aws_lc_rs` feature), or so it seems to its author.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,24 +4,24 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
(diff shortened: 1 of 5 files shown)


djc:
Without this change, is it still possible to generate secp256k1 certificates with rcgen? If so, how much boilerplate is there?

PR #409, opened 2026-01-26
API cleanup: non-exhaustive types, `SanType` -> `GeneralName`

djc:
For when we do our next semver-incompatible release.


djc:
We're open for semver-incompatible changes now -- @ctz do you have time to rebase this?

PR #411, opened 2026-02-10
Take semver-compatible dependency updates

djc:
In particular, the time update for https://rustsec.org/advisories/RUSTSEC-2026-0009.html.

Not super happy about the MSRV bump here, but given that we also ship a CLI crate, this seems like the prudent thing to do maybe?


cpu:
@est31 are you interested/available to review this?

PR #413, opened 2026-02-26
Rename `SignatureAlgorithm` values a bit

ctz:
Just noticed some of these bits, not fixing a specific issue.

I also noticed that there's a lot of RSA-PSS-related things that are commented out. I've left that alone but it probably deserves some attention for someone who knows the background?


djc (approved):
Good stuff, thanks!

CI indicates something went not quite right with RSA_PSS?

est31 (approved):

cpu:
ctz: FWIW I think this is blocked on the CI failures being investigated.

djc:
@ctz would you have a chance to rebase this/clean this up?

PR #417, opened 2026-03-12
Added support for RSASSA-PSS padding algorithms

MasonCitywide:
Added algorithms to `sign_algo.rs` `PKCS_RSA_PSS_SHA256`, `PKCS_RSA_PSS_SHA384`, and `PKCS_RSA_PSS_SHA512`. 

A half implemented version of `PKCS_RSA_PSS_SHA256` already existed with a comment saying this doesn't work because `ring` hasn't implemented PSS padding ([here](https://github.com/briansmith/ring/issues/1353)). It seems that since then it has ([here](https://github.com/briansmith/ring/blame/main/src/rsa/padding/pss.rs)), and that comment was made before the release of `aws-lc-rs`.

There was also an issue in the pre-existing `PKCS_RSA_PSS_SHA256` function in which the salt length was set to the default 20 instead of the recommended value of the number of octets of the hash algorithm ([RFC 4055](https://datatracker.ietf.org/doc/html/rfc4055#section-3.1), pg. 9).

This is an important change as, if I am reading it correctly, non-PSS padding has been deprecated since [RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.3) (pg. 70), with security concerns like [ROBOT](https://robotattack.org/).

I was able to successfully create CSRs using all three of these algorithms using the `aws-lc-rs` backend. However, I'm not familiar with the unit testing of this library and I am new to contributions, so I would appreciate an independant review of these additions before they are merged.

Thank you for your time,
MC

### DIFF
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -259,12 +259,18 @@ impl KeyPair {
 		} else if alg == &PKCS_RSA_SHA512 {
 			let rsakp = RsaKeyPair::from_pkcs8(&serialized_der)._err()?;
 			KeyPairKind::Rsa(rsakp, &signature::RSA_PKCS1_SHA512)
-		} else if alg == &PKCS_RSA_PSS_SHA256 {
-			let rsakp = RsaKeyPair::from_pkcs8(&serialized_der)._err()?;
-			KeyPairKind::Rsa(rsakp, &signature::RSA_PSS_SHA256)
 		} else {
 			#[cfg(feature = "aws_lc_rs")]
-			if alg == &PKCS_ECDSA_P521_SHA256 {
+			if alg == &PKCS_RSA_PSS_SHA256 {
+				let rsakp = RsaKeyPair::from_pkcs8(&serialized_der)._err()?;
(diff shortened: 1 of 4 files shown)


djc (approved):
The changes here look okay so far, but CI doesn't pass and we'll want to make sure there some tests exercising this against a different backend (maybe OpenSSL)?

cpu:
> There was also an issue in the pre-existing PKCS_RSA_PSS_SHA256 function in which the salt length was set to the default 20 instead of the recommended value of the number of octets of the hash algorithm ([RFC 4055](https://datatracker.ietf.org/doc/html/rfc4055#section-3.1), pg. 9).

Just wanted to note I bumped into this in https://github.com/rustls/rcgen/pull/445 where I noticed not only did we set the salt length to the wrong value, but we also misencoded it by writing the default that should have been elided for proper DER encoding. I removed the half implemented support in https://github.com/rustls/rcgen/pull/445/changes/f8ef6ee505f3ece136d816435bd0ab8f37d49398 and if anyone revives this work we should take care to not re-introduce that bug.

PR #426, opened 2026-04-28
Support multi-valued distinguished names

dwhjames:
As `DistinguishedName` has a private internal structure, it is possible to make a backwards compatible change to support multi-valued relative distinguished names, such as for `OU`, which is the common case.

Existing code that was written with the expectation of strict set semantics will continue to have the same behavior. New functions `push_multi`, `get_all` allow the extended behavior.

Parsing via `from_name` previously returned an error on multi-attribute RDNs (a SET containing more than one attribute/value pair). It now accepts them by flattening each attribute into a separate entry; on round-trip they will be emitted as individual single-attribute RDNs.

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -279,7 +279,7 @@ impl CertificateParams {
 
 		// Per https://tools.ietf.org/html/rfc5280#section-4.1.2.6, SAN must be marked
 		// as critical if subject is empty.
-		let critical = self.distinguished_name.entries.is_empty();
+		let critical = self.distinguished_name.is_empty();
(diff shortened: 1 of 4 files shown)


cpu:
Thanks for the PR.

I don't think adding new disparate API surface is the best path forward for this. I personally prefer https://github.com/rustls/rcgen/pull/394 and handling this in https://github.com/rustls/rcgen/issues/418 without consideration for backwards compatibility within 0.14.x.

dwhjames:
@cpu thanks for the quick review!
Sorry to you and @djc, I clearly didn’t do a very good job searching the open Issues and PRs 🙃 .

cpu:
No big deal :-) Thanks for tacking a crack at improving it either way.

PR #427, opened 2026-04-28
update key_pair to signing_key in README.md

fakelozic:
https://docs.rs/rcgen/latest/rcgen/ used signing_key so updated the README.md


cpu (approved):
Thanks!

PR #433, opened 2026-05-26
doc: amend description to point to CertificateParams for additional explanation

mcr:
amend CertificateSigningRequest documentation to more clearly point at CertificateParams as the way to generate a new CSR with attributes.
[In the future, building one from a CSR attributes (RFC9908) object will be contributed]

### DIFF
--- a/rcgen/src/csr.rs
+++ b/rcgen/src/csr.rs
@@ -37,6 +37,11 @@ impl PublicKeyData for PublicKey {
 }
 
 /// A certificate signing request (CSR) that can be encoded to PEM or DER.
+///
+/// A new certificate signing request is created by filling out a [`CertificateParams`] object
+/// and then calling [`CertificateParams::serialize_request()`] or
+/// [`CertificateParams::serialize_request_with_attributes()`] with a
+/// [`crate::KeyPair`] (or other [`SigningKey`] capable trait).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CertificateSigningRequest {
 	pub(crate) der: CertificateSigningRequestDer<'static>,


rcgen/src/csr.rs
```diff
@@ -37,6 +37,11 @@ impl PublicKeyData for PublicKey {
 }
 
 /// A certificate signing request (CSR) that can be encoded to PEM or DER.
+///
+/// A new certificate signing request is created by filling out a CertificateParams object
+/// and then calling serialize_request() or serialize_request_with_attributes() with a
+/// KeyPair (or other SigningKey capable trait).
+///
```
cpu: Nit: the trailing `///` without content doesn't seem necessary.

rcgen/src/csr.rs
```diff
@@ -37,6 +37,11 @@ impl PublicKeyData for PublicKey {
 }
 
 /// A certificate signing request (CSR) that can be encoded to PEM or DER.
+///
+/// A new certificate signing request is created by filling out a CertificateParams object
```
cpu: I think it would be nicer to use intra-item rustdoc links for `CertificateParams`, `KeyPair`, `SigningKey`, and the ref'd methods.

cpu (commented):
Thanks for the PR. I think it's a good idea but had nits to consider for the impl.

rcgen/src/csr.rs
cpu: Resolved.

rcgen/src/csr.rs
cpu: Resolved.

cpu:
> Thanks for the PR. I think it's a good idea but had nits to consider for the impl.

I pushed the requested changes myself. Hope you don't mind. The nits were tiny and it seemed silly to leave this sitting around for another 2 weeks for minor adjustments.

djc (approved):

cpu:
> github-merge-queue Bot removed this pull request from the merge queue due to failed status check

Usual GitHub flakyness :roll_eyes: 

> Internal server error. Correlation ID: cadaf68a-825d-4db5-8b3d-5e1ba2ffac8d

PR #440, opened 2026-06-29
Fix non-compliant PKCS#8 v2 Ed25519 test key encoding

justsmth:
The `ED25519_TEST_KEY_PAIR_PEM_V2` test key encoded the `publicKey` field with explicit tagging (`a1 23 03 21 00 …`), which is the old ring 0.16.x form.

* Per [RFC 8410 §10.3](https://www.rfc-editor.org/rfc/rfc8410#section-10.3) (worked example: *"Ed25519 private key encoded with an attribute and the public key"*) and the normative `OneAsymmetricKey` definition in [§7](https://www.rfc-editor.org/rfc/rfc8410#section-7), where `publicKey [1]` is `IMPLICIT` -- the field must be an implicitly tagged primitive `BIT STRING` (`81 21 00 …`). This change re-tags it to that form.
* The key material is unchanged. It's the same seed and public key, only the tagging is corrected.
* This touches a test fixture only; no library code changes.

The current key fails strict parsers. Specifically, `aws-lc-rs` backed by AWS-LC >= 5.0.0 fails `test_openssl_25519_v2_given`, b/c AWS-LC now rejects the unconsumed trailing field as of aws/aws-lc#3242. (More lenient parsers silently ignore trailing data.)

I verified the `verify-tests` `openssl` and `webpki` suites pass with the new key against AWS-LC 5.1.0.


ctz (approved):
FWIW Graviola fails to parse the prior encoding (with `Asn1Error(UnexpectedTrailingData)`) and successfully parses the fixed encoding, so this change seems good to me.

verify-tests/src/lib.rs
```diff
@@ -6,14 +6,13 @@ use rcgen::{
 };
 use time::{Duration, OffsetDateTime};
 
-// Generated by adding `println!("{}", cert.serialize_private_key_pem());`
```
djc, line 9: So is the `cert.serialize_private_key_pem()` code that this was apparently generated from correct these days? Or is there something broken there still?

djc (approved):

verify-tests/src/lib.rs
justsmth: Fixed here? https://github.com/briansmith/ring/pull/1680

djc: Makes sense, thanks!

cpu: It looks like the generated version that was in-tree was quite old, circa 2019 (https://github.com/rustls/rcgen/commit/07c8824ecc713e3d9762384b654ab560f652b36e)

PR #445, opened 2026-08-10
Spec compliance bug fixes

cpu:
I let claude loose to do a review of bugs similar to https://github.com/rustls/rcgen/pull/444 and it came up with several findings I confirmed myself. As an experiment I let it write the unit tests and fixes and then reviewed them myself to satisfaction. In summary:

* We were using the wrong time for CRL InvalidityDate
* We allowed an empty CRL distribution points URI list that would produce a misencoding
* Cert. params with only key_usages or crl_distribution_points wouldn't produce a cert w/ those extensions
* There was a way to write a CRL entry with reasonCode==0, which is invalid
* PKCS_RSA_PSS_SHA256 salt encoding was wrong at least two ways

The last commit drops some crate-internal RSASSA-PSS bits rather than try to fix the salt length bugs. I think that's the right call and we can revive it (correctly) in https://github.com/rustls/rcgen/pull/417 when someone has the time/energy.

~~I opted for a "write an ignored test that catches the bug and fails without the fix, then write the fix and unignore the test" approach here because I wanted to keep the LLM on the rails. If folks prefer we can squash each of those paired commits into just one. LMK.~~

### DIFF
--- a/rcgen/src/certificate.rs
+++ b/rcgen/src/certificate.rs
@@ -493,8 +493,10 @@ impl CertificateParams {
 			// write extensions
 			let should_write_exts = self.use_authority_key_identifier_extension
 				|| !self.subject_alt_names.is_empty()
+				|| !self.key_usages.is_empty()
 				|| !self.extended_key_usages.is_empty()
 				|| self.name_constraints.iter().any(|c| !c.is_empty())
+				|| !self.crl_distribution_points.is_empty()
 				|| matches!(self.is_ca, IsCa::ExplicitNoCa)
 				|| matches!(self.is_ca, IsCa::Ca(_))
 				|| !self.custom_extensions.is_empty();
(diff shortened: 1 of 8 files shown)


rcgen/src/certificate.rs
```diff
@@ -482,8 +493,10 @@ impl CertificateParams {
 			// write extensions
 			let should_write_exts = self.use_authority_key_identifier_extension
 				|| !self.subject_alt_names.is_empty()
+				|| !self.key_usages.is_empty()
```
cpu, line 496: The way we do this is super brittle and makes me want to revive the extension rework branch I started ~2 years ago that reworked it as a systematic fix. 

Will try to find time to do that soon so we don't bump into this again.

djc (approved):
Nice, LGTM!

I'd be in favor of squashing both commits for each fix into one.

rcgen/src/certificate.rs
djc: Maybe add a note in

- https://github.com/rustls/rcgen/issues/418 

I think we could ship a 0.15 pretty quickly with some good improvements.

cpu: Filed https://github.com/rustls/rcgen/issues/446 with some quick notes from my old branch and added it to the 0.15 release planning. Check in with me if I haven't pushed anything for this by end of next week.

cpu:
> I'd be in favor of squashing both commits for each fix into one.

Done.

PR #449, opened 2026-08-29
Rework X.509 cert/csr/crl extension representation

### LINKED ISSUE #122: X509v3 extensions from certificate not transferred to CSR

### LINKED ISSUE #150: Allow CSR parsing to handle custom extensions
Hi,

Currently when using a user defined [custom_extensions](https://docs.rs/rcgen/latest/rcgen/struct.CertificateParams.html#structfield.custom_extensions), this causes the [CSR parsing](https://docs.rs/rcgen/latest/rcgen/struct.CertificateSigningRequest.html#method.from_der) to error out with [UnsupportedExtension](https://docs.rs/rcgen/latest/rcgen/enum.RcgenError.html#variant.UnsupportedExtension).

Would it be possible to allow CSR parsing of custom extensions, such as by providing a closure to handle unknown extensions?

### LINKED ISSUE #155: Enforce extension uniqueness
As a breaking API change `rcgen` should refuse to generate certificates with duplicate extensions. 

[RFC 5280 §4.2](https://www.rfc-editor.org/rfc/rfc5280#section-4.2) says:
> A certificate MUST NOT include more than one instance of a particular extension.

This can occur with the current API in two ways:

1. Specifying duplicate extensions within `CertificateParams.custom_extensions`
2. Specifying an extension in `CertificateParams.custom_extensions` that has the same OID as an extension `rcgen` emits natively (e.g. `id-ce-subjectAltName`, `id-ce-authorityKeyIdentifier `, etc).

### LINKED ISSUE #446: Rework X509 extension code
The extension-writing code currently duplicates each extension's presence condition away from the code that encodes it (e.g. the `should_write_exts` gate vs. `write_extensions`), which has already produced bugs where requested extensions were silently dropped. 

I think we should rework this to represent extensions with an internal `Extension` trait implemented once per extension type, with per-type `from_params` constructors so presence logic lives beside encoding, plus an `Extensions` collection that rejects duplicate OIDs (fixing https://github.com/rustls/rcgen/issues/155) and only emits the outer extensions wrapper when the built collection is non-empty. This would make emptiness observed rather than predicted, and eliminating the gate-drift bug class.

I had started on some of this in [a very stale WIP branch](https://github.com/rustls/rcgen/compare/main...cpu:rcgen:cpu-rework-ext-handling) and we should revive that before a next major release.

cpu:
This starts the extension rework from #446. Each extension is now its own type in a new ext module, with the presence decision living next to the encoding. Breaking changes are lifted to the follow-up https://github.com/rustls/rcgen/pull/454


rcgen/src/ext.rs
```diff
@@ -0,0 +1,249 @@
...
+	fn from(critical: bool) -> Self {
+		match critical {
+			true => Self::Critical,
+			false => Self::NonCritical,
+		}
+	}
+}
+
+/// Serializes an X.509v3 extension according to RFC 5280.
+pub(crate) fn write_extension(writer: DERWriter, extension: &dyn Extension) {
```
djc, line 81: I think this could be `Extension::write(&self, writer)`?

rcgen/src/ext.rs
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	fn criticality(&self) -> Criticality {
+		T::CRITICALITY
+	}
+
+	fn write_value(&self, writer: DERWriter) {
+		// Calling with fully qualified syntax to disambiguate.
+		StaticExtension::write_value(self, writer)
+	}
```
djc, line 37: Nits: I think the order of these methods should be reversed.

rcgen/src/ext.rs
```diff
@@ -0,0 +1,249 @@
...
+/// instead.
+pub(crate) trait StaticExtension: Debug {
+	/// The OID components of the extension.
+	const OID: &'static [u64];
+
+	/// The criticality of the extension.
+	const CRITICALITY: Criticality;
+
+	/// Write the extension's value (the content of the extnValue OCTET STRING).
+	fn write_value(&self, writer: DERWriter);
```
djc, line 22: Nits: I think the order of these items within the `trait` definition should be reversed.

rcgen/src/ext.rs
```diff
@@ -1,9 +1,10 @@
 use std::fmt::Debug;
+use std::net::IpAddr;
 
 use yasna::models::ObjectIdentifier;
 use yasna::{DERWriter, Tag};
 
-use crate::{oid, Issuer, KeyIdMethod, SigningKey};
+use crate::{oid, CertificateParams, Issuer, KeyIdMethod, SanType, SigningKey};
```
djc, line 8: I think the order of items in this module is bottom up right now and should be more like:

- `AuthorityKeyIdentifier`
- `Extension`
- `StaticExtension`
- `Criticality`

rcgen/src/ext.rs
```diff
@@ -273,6 +276,45 @@ impl StaticExtension for KeyUsage<'_> {
 	}
 }
 
+/// An X.509v3 extended key usage extension according to [RFC 5280 §4.2.1.12].
+///
+/// [RFC 5280 §4.2.1.12]: <https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.12>
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ExtendedKeyUsage<'params>(&'params [ExtendedKeyUsagePurpose]);
```
djc, line 283: Nit: IMO we should also move `ExtendedKeyUsagePurpose` (and other such types that are mainly/only used for these extensions) into this module.

djc (commented):
28 commits is a bit much. 😄 

Suggest we split off a separate PR for the first bunch of commits, prior to the version bump? I have reviewed those and will be able to approve that quickly.

rcgen/src/ext.rs
cpu: Good idea. Done.

rcgen/src/ext.rs
cpu: Done.

rcgen/src/ext.rs
cpu: Done.

rcgen/src/ext.rs
cpu: Done.

rcgen/src/ext.rs
cpu: Done.

cpu:
> Suggest we split off a separate PR for the first bunch of commits, prior to the version bump? I have reviewed those and will be able to approve that quickly.

Done. I lifted everything after the version bump to a stacked follow-up PR https://github.com/rustls/rcgen/pull/454

djc (approved):

PR #447, opened 2026-08-11
Take advantage of stable ML-DSA in aws-lc-rs

djc:

Issue #414, opened 2026-03-05
Adding SymCrypt as a crypto provider


juliusl:
Hello, I have a fork where I've added support for [SymCrypt](https://github.com/microsoft/rust-symcrypt) as a crypto provider - so far it seems to work w/ the tests in the repo as well as another project I'm working with as a drop-in replacement.

Since it would be adding a new crypto provider, I wanted to file an issue to see if this something this project would be open to before putting up a PR.

juliusl:
> I haven't given this deep thought yet, but my initial instinct is that we should consider following the rustls approach of making it possible for code external to the library to provide its own cryptography provider of choice and not bundling more options as built-in features.

That would be helpful, and then I would be able to just use the `rustls-symcrypt` provider

PR #450, opened 2026-09-02
add pluggable crypto provider

### LINKED ISSUE #228: Support `CryptoProvider` like plugin mechanism
https://github.com/rustls/rustls/issues/1540 in rustls provides a mechanism to plug in arbitrary crypto providers. While the in-tree ones only have `ring` and `aws_lc_rs`, mirroring rcgen, there are out of tree providers that can be used. However, if an application depends on `rcgen`, these benefits are negated (assuming a project requires all crypto to be done by a specific implementation).

It would be nice to have a pluggable model.

jgreeer:
Adds a pluggable cryptography provider API to rcgen, allowing applications to use custom backends without depending on Ring or AWS-LC. This design is inspired by rustls and jsonwebtoken.

This PR is a follow up on https://github.com/rustls/rcgen/issues/414. The goal is to support our own external SymCrypt provider crate.

Ring and AWS-LC remain available through built-in provider implementations.

- Added a `CryptoProvider` trait providing:
  - Hashing
  - Key generation
  - Private-key loading
  - Signature verification
- Updated APIs for key generation/loading, certificate signing, CRL signing, CSR parsing, and convenience helpers to take a provider explicitly.
- Removed process-wide provider state and automatic provider selection.
- Moved Ring and AWS-LC operations into independent provider modules.
- Made `KeyPair` backend-neutral by storing a `SigningKey` trait object instead of concrete Ring/AWS-LC key types.
- Routed certificate serial-number and key-identifier hashing through `CryptoProvider::hash`.
- Routed CSR self-signature verification through `CryptoProvider::verify` instead of `x509-parser` crypto features.
- Made P-521 algorithm metadata available to custom providers.
- Brought over ML-DSA support in the AWS-LC provider (https://github.com/rustls/rcgen/pull/447).
- Added provider documentation and end-to-end custom-provider tests.

Built-in providers must now be enabled and selected explicitly:

- rcgen does not enable Ring or AWS-LC by default.
- The `ring` feature exposes the built-in Ring provider.
- The `aws_lc_rs` feature exposes the built-in AWS-LC provider.
- If both are enabled, the application still chooses which provider to pass.

Providers are passed directly to APIs:

```rust
let provider = rcgen::crypto::ring::default_provider();

let key = KeyPair::generate(provider)?;
let cert = params.self_signed(&key, provider)?;
```

Resolves https://github.com/rustls/rcgen/issues/228


cpu:
Thanks for picking this up.

> Custom providers can either be installed once:
> ...
> Or passed directly to explicit APIs:
> ...
> The explicit form is intended for libraries and avoids process-global state.

I think we should probably model this like rustls `main` and not like rustls 0.23. E.g. splitting the providers we offer first-party into their own crates in the workspace, and requiring all the library code consumers to instantiate the provider explicitly, not doing any kind of default in rcgen's lib code itself based on feature flags.

@djc Are we on the same page there?

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+	/// Provider for signature verification.
+	pub signature_verification_provider: &'static dyn SignatureVerificationProvider,
+}
+
+impl CryptoProvider {
+	/// Set this provider as the default for the current process.
+	///
+	/// This can succeed at most once during a process execution. Call it before invoking any
+	/// convenience API that uses the process-wide provider.
+	pub fn install_default(self) -> Result<(), Arc<Self>> {
```
djc: I don't think we're going to need/want a default crypto provider setup.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum HashAlgorithm {
+	/// SHA-256.
+	Sha256,
+	/// SHA-384.
+	Sha384,
+	/// SHA-512.
+	Sha512,
+}
```
djc: We'll want items to be ordered top-down, so `HashAlgorithm` should be at the bottom.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+		match self {
+			Self::Sha256 => 32,
+			Self::Sha384 => 48,
+			Self::Sha512 => 64,
+		}
+	}
+}
+
+/// Hash operations supplied by a [`CryptoProvider`].
+pub trait DigestProvider: Debug + Send + Sync {
```
djc: Looks like this diverged substantially from the rustls `Hash` setup? Should mimic that more closely.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+	/// Providers that do not support selectable RSA key sizes may retain the default
+	/// implementation.
+	fn generate_rsa(
+		&self,
+		algorithm: &'static SignatureAlgorithm,
+		key_size: RsaKeySize,
+	) -> Result<KeyPair, Error> {
+		let _ = (algorithm, key_size);
+		Err(Error::KeyGenerationUnavailable)
+	}
```
djc: I don't think this needs to be a separate method?

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+
+	/// Decode and validate an exportable private key.
+	///
+	/// If `algorithm` is `Some`, the key must be loaded for exactly that signature algorithm.
+	/// If it is `None`, the provider detects a supported algorithm from the key.
+	fn load_private_key(
+		&self,
+		key_der: PrivateKeyDer<'static>,
+		algorithm: Option<&'static SignatureAlgorithm>,
```
djc, line 43: Why do we need `algorithm` here?

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+/// features are enabled. With no backend or the `custom-provider` feature, applications must
+/// install a provider explicitly.
+#[derive(Clone, Debug)]
+pub struct CryptoProvider {
+	/// Provider for generating and loading private key pairs.
+	pub key_pair_provider: &'static dyn KeyPairProvider,
+	/// Provider for SHA-256, SHA-384, and SHA-512 hashing.
+	pub digest_provider: &'static dyn DigestProvider,
+	/// Provider for signature verification.
+	pub signature_verification_provider: &'static dyn SignatureVerificationProvider,
```
djc: Looks like it might be simpler here to just have a `trait CryptoProvider` with `hash()`, `load_private_key()` and `verify()` methods.

rcgen/src/crypto/aws_lc_rs.rs
```diff
@@ -0,0 +1,492 @@
...
+use ::aws_lc_rs::rsa::KeySize;
+use ::aws_lc_rs::signature::{
+	self, EcdsaKeyPair, Ed25519KeyPair, KeyPair as _, RsaEncoding, RsaKeyPair,
+	VerificationAlgorithm,
+};
+#[cfg(feature = "aws_lc_rs")]
+use ::aws_lc_rs::signature::{
+	PqdsaKeyPair, PqdsaSigningAlgorithm, ML_DSA_44, ML_DSA_44_SIGNING, ML_DSA_65,
+	ML_DSA_65_SIGNING, ML_DSA_87, ML_DSA_87_SIGNING,
+};
```
djc: I don't think we need the `::` prefix on these imports?

djc (commented):
Thanks! This looks directionally right to me.

jgreeer:
> Thanks for picking this up.
> 
> > Custom providers can either be installed once:
> > ...
> > Or passed directly to explicit APIs:
> > ...
> > The explicit form is intended for libraries and avoids process-global state.
> 
> I think we should probably model this like rustls `main` and not like rustls 0.23. E.g. splitting the providers we offer first-party into their own crates in the workspace, and requiring all the library code consumers to instantiate the provider explicitly, not doing any kind of default in rcgen's lib code itself based on feature flags.
> 
> @djc Are we on the same page there?

hey! thanks for the quick review. just before i change anything, are you in agreement with this as well @djc?

djc:
I agree that we should not have a default provider or enable any provider by default, and API that needs it should take one explicitly. I don't think it's worth splitting the providers out into separate crates at this point? We're not close to API stability and I feel like rcgen is much less likely than rustls to end up deeper in dependency graphs.

cpu:
>  I don't think it's worth splitting the providers out into separate crates at this point? 

That's fine with me :+1:

rcgen/src/certificate.rs
```diff
@@ -249,21 +298,23 @@ impl CertificateParams {
...
+		subject_key_identifier: Option<&[u8]>,
+	) {
 		let is_ca = match &self.is_ca {
 			IsCa::Ca(bc) => Some(bc),
 			IsCa::ExplicitNoCa => None,
 			IsCa::NoCa => return,
 		};
 
-		if let Some(pub_key_spki) = pub_key_spki {
+		if let Some(subject_key_identifier) = subject_key_identifier {
```
djc: Don't make unrelated changes.

rcgen/src/crypto/mod.rs
jgreeer: i think this is for compatibility with `from_der_and_sign_algo`. the old `KeyPair` was already bound to `SignatureAlgorithim`. this could be changed to have the signing algorithim desicion happen at the time of actual signing though.

rcgen/src/crypto/mod.rs
jgreeer: makes sense, i only made it a struct to match rustls and jsonwebtoken but the abstraction fitting the project the best probably matters more here.

rcgen/src/certificate.rs
jgreeer: this function actually is different. because derive can fall now (through a provider), i moved that function call to write_extensions. so even though it looks like they are being passed the same thing, it's actually the much smaller SKI being passed instead of the SPKI. it might be better to make this function failable and retain the original arguments though.

rcgen/src/crypto/mod.rs
jgreeer: changed to be closer to rustls

rcgen/src/certificate.rs
jgreeer: hash no longer fails, reverted this back to the original

.github/workflows/ci.yml
```diff
@@ -48,6 +48,12 @@ jobs:
             - run: cargo clippy --features ring,pem,x509-parser --all-targets
             # rustls-cert-gen require either aws_lc_rs or ring feature
             - run: cargo clippy -p rcgen --no-default-features --all-targets
+            - run: cargo clippy -p rcgen --no-default-features --features crypto,pem,x509-parser --all-targets
+            - name: Ensure custom-provider builds have no built-in crypto dependencies
+              run: |
+                  if cargo tree -p rcgen --no-default-features --features crypto --edges normal,build | grep -E 'ring v|aws-lc'; then
```
cpu: The regexp here seems a little odd (matching `" v"` for ring but not aws-lc, and only using the prefix of the aws-lc-rs dep).

Should it be `ring|aws-lc-rs` or `\sring\s|\saws-lc-rs\s` instead?

rcgen/src/certificate.rs
```diff
@@ -139,33 +139,61 @@ impl CertificateParams {
 	/// [`Certificate::pem`].
 	pub fn signed_by(
 		&self,
-		public_key: &impl PublicKeyData,
+		public_key: &(impl PublicKeyData + ?Sized),
 		issuer: &Issuer<'_, impl SigningKey>,
+		#[cfg(feature = "crypto")] provider: &dyn CryptoProvider,
```
cpu: I think we've kind of made a mess of this already with the `crypto` feature but I'd like to avoid making more pub API surface that has feature-gated params/fields. Unification makes it unpleasant for consumers who can have working code fail to compile after adding a dep that activates the feature under them.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+	) -> Result<(), Error>;
+}
+
+/// `ring`-based cryptography provider.
+#[cfg(feature = "ring")]
+pub mod ring;
+
+/// AWS-LC-based cryptography provider.
+#[cfg(feature = "aws_lc_rs")]
+pub mod aws_lc_rs;
```
djc: Nit: module declarations before any other items (but after imports).

rcgen/src/certificate.rs
djc: Agreed. Also given the existence of an `impl SigningKey` there must be a `CryptoProvider` in play somewhere?

Feels to me like we should just get rid of the `crypto` feature in this PR because the crypto dependencies are instead abstracted via traits, and the `aws-lc-rs` and `ring` features should in principle only be used for their respective provider implementations?

.github/workflows/ci.yml
```diff
@@ -44,15 +44,23 @@ jobs:
...
             - run: cargo clippy -p rcgen --no-default-features --all-targets
+            - run: cargo clippy -p rcgen --no-default-features --features pem,x509-parser --all-targets
+            - name: Ensure backend-free builds have no built-in crypto dependencies
+              run: "! cargo tree -p rcgen --no-default-features --edges normal,build | grep -E 'ring v|aws-lc'"
+            - name: Ensure FIPS requires an explicit provider
+              run: |
+                  if output=$(cargo check -p rcgen --no-default-features --features fips 2>&1); then
+                    exit 1
+                  fi
+                  grep -Fq "the 'fips' feature currently requires the 'aws_lc_rs' feature" <<<"$output"
```
djc: I think we should split the features into `aws-lc-rs` and `aws-lc-rs-fips`, instead of having a separate `fips` feature; but I also think this PR is already large and I think we can scope this out into a separate PR.

.github/workflows/ci.yml
```diff
@@ -44,15 +44,23 @@ jobs:
...
             - run: cargo clippy --no-default-features --features ring --all-targets
             - run: cargo clippy --no-default-features --features aws_lc_rs,pem,x509-parser --all-targets
             - run: cargo clippy --no-default-features --features aws_lc_rs_unstable,pem,x509-parser --all-targets
             - run: cargo clippy --no-default-features --features aws_lc_rs --all-targets
-            - run: cargo clippy --no-default-features --features fips,pem,x509-parser --all-targets
+            - run: cargo clippy --no-default-features --features aws_lc_rs,fips,pem,x509-parser --all-targets
```
djc: Why is this necessary?

rcgen/examples/rsa-irc-openssl.rs
```diff
@@ -5,21 +5,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+	let key_pair = rcgen::KeyPair::from_pem(&key_pair_pem, provider)?;
 
-	let cert = params.self_signed(&key_pair)?;
+	let cert = params.self_signed(&key_pair, provider)?;
 	let pem_serialized = cert.pem();
 	let pem = pem::parse(&pem_serialized)?;
 	let der_serialized = pem.contents();
-	let hash = ring::digest::digest(&ring::digest::SHA512, der_serialized);
-	let hash_hex = hash.as_ref().iter().fold(String::new(), |mut output, b| {
+	let hash = openssl::sha::sha512(der_serialized);
```
djc: Why use openssl here instead of ring?

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+	/// STRING contents, matching [`PublicKeyData::der_bytes`](crate::PublicKeyData::der_bytes).
+	fn verify(
+		&self,
+		algorithm: &'static SignatureAlgorithm,
+		public_key: &[u8],
+		message: &[u8],
+		signature: &[u8],
+	) -> Result<(), Error>;
+}
+/// A hash algorithm required by rcgen.
```
djc, line 60: Nit: keep an empty line between these.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+	) -> Result<KeyPair, Error>;
+
+	/// Verify `signature` over `message` using `public_key` and `algorithm`.
+	///
+	/// rcgen uses this operation to verify the self-signature on a parsed PKCS#10 certificate
+	/// signing request. `public_key` contains the SubjectPublicKeyInfo `subjectPublicKey` BIT
+	/// STRING contents, matching [`PublicKeyData::der_bytes`](crate::PublicKeyData::der_bytes).
+	fn verify(
+		&self,
+		algorithm: &'static SignatureAlgorithm,
```
djc: Argument order should be

- `message`
- `signature`
- `public_key`
- `algorithm`

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+	/// Hash `input` with `algorithm`.
+	fn hash(&self, algorithm: HashAlgorithm, input: &[u8]) -> HashOutput;
+
+	/// Generate an exportable key pair for `algorithm`.
+	///
+	/// `key_size` selects an explicit RSA key size. It must be `None` for non-RSA algorithms.
+	fn generate(
+		&self,
+		algorithm: &'static SignatureAlgorithm,
+		key_size: Option<RsaKeySize>,
```
djc: Instead of passing in a separate key size, suggest we first (in a separate PR) change it so there are separate `SignatureAlgorithm` values for different RSA key sizes.

rcgen/src/certificate.rs
```diff
@@ -139,33 +135,46 @@ impl CertificateParams {
 	/// [`Certificate::pem`].
 	pub fn signed_by(
 		&self,
-		public_key: &impl PublicKeyData,
+		public_key: &(impl PublicKeyData + ?Sized),
```
djc: Why do we need the `?Sized` bound here?

rcgen/src/certificate.rs
```diff
@@ -139,33 +135,46 @@ impl CertificateParams {
...
 		self.key_identifier_method
-			.derive(key.subject_public_key_info())
+			.derive(provider, key.subject_public_key_info())
 	}
 
-	#[cfg(all(test, feature = "x509-parser"))]
+	#[cfg(all(
+		test,
+		feature = "x509-parser",
+		any(feature = "ring", feature = "aws_lc_rs")
```
djc: I don't think we need the `any` here?

rcgen/src/certificate.rs
```diff
@@ -249,21 +258,28 @@ impl CertificateParams {
 	}
 
 	/// Write a certificate's BasicConstraints as defined in RFC 5280.
-	fn write_ca_extensions(&self, writer: &mut DERWriterSeq, pub_key_spki: Option<&[u8]>) {
+	fn write_ca_extensions(
+		&self,
+		writer: &mut DERWriterSeq,
+		pub_key_spki: Option<&[u8]>,
+		provider: Option<&dyn CryptoProvider>,
```
djc: The separate `Option`s for `pub_key_spki` and `provider` don't make sense.

rcgen/src/crl.rs
```diff
@@ -187,7 +178,16 @@ impl CertificateRevocationListParams {
...
+		provider: &dyn CryptoProvider,
 	) -> Result<CertificateRevocationList, Error> {
+		self.validate(issuer)?;
+
+		Ok(CertificateRevocationList {
+			der: self.serialize_der(issuer, provider)?.into(),
+		})
+	}
+
+	fn validate(&self, issuer: &Issuer<'_, impl SigningKey>) -> Result<(), Error> {
```
djc: Why move this into a separate method?

If you're only going to get your LLM to iterate on this PR we might as well stop reviewing this here, or your employer can pay me for my time.

.github/workflows/ci.yml
jgreeer: noted, happy to make a PR for that once this merges.

.github/workflows/ci.yml
jgreeer: i coupled aws_lc_rs + fips together as flags in this PR, but it's ultimately irrelevant to the actual provider abstration. since we're going to rewrite that in the future anyways, should i just revert to previous behaviour?

rcgen/examples/rsa-irc-openssl.rs
jgreeer: shouldn't have, we already bring the ring provider in so we can use it here for the digest. changed

rcgen/src/crypto/mod.rs
jgreeer: added

rcgen/src/crypto/mod.rs
jgreeer: changed

rcgen/src/certificate.rs
jgreeer: not needed, i've reverted this back to the original

rcgen/src/crl.rs
jgreeer: Yeah, that change doesn't really make sense, will revert. I try to review the commits, but clearly some nonsensical changes slipped in. After reviewing the rest of your comments, I'll go through and make sure there aren't any more unrelated additions here, sorry about that. Thanks so much for the reviews.

djc: It is really not great that you're getting volunteers to review something that you haven't even reviewed yourself.

rcgen/src/certificate.rs
jgreeer: yeah, there are some other lines where this was added too. took those out as well

jgreeer: the test module in `certificate.rs` pulls in a provider through `crate::test_provider()`, so we actually do have to keep this.

rcgen/src/crypto/mod.rs
jgreeer: Sure, happy to make a PR for this

jgreeer:
Reverted many of the unnecessary changes, remaining edits should be only crypto provider related

Changes:

- Removed the changing of the `fips` flag logic, it now behaves exactly like it used to. This does mean that the testing functions in `certificate.rs` require the flag in the `any` section of the config now.
- Removed other changes to README, the only README change is the new Cryptography providers section, which simply explains that you need to enable a provider now and pass it through the functions & the provider in the example
- cargo.toml now has the original feature flags, except for crypto which we agreed to remove
- Derive the SKI inline instead of before calling write_x509_extension in `certificate.rs`, same change in `serialize_der` from `crl.rs`
- Switch back to retrieving the private key via `private_key.contents()` in `key_pair.rs`, the provider calls `clone_key` later for ownership.
- Restore original shape of `serialize_pem` in `key_pair.rs` where we assign `serialize_der` to `contents
- Returned the SPKI parsing flow in `from_der` in `key_pair.rs` to its original form
- Removed boxed blanket implementations for PublicKeyData and SigningKey in `key_pair.rs`
- Added back inline certificate parsing in `self_signed` in `lib.rs`
- Use `to_vec()` and `0..20` in the derive function in `KeyIdMethod`
- Revert changes entirely to `string.rs`, nothing should be touched there
- In aws-lc-rs.rs
	- Variables have been renamed to match their original names in key_pair.rs
	- ECSDA keys use from `EcdsaKeyPair::from_pkcs8` instead of `load_with_algorithm` 
	- Wrapped match in `Ok` in the implementation of `sign` instead of around each signature to match original function
- In ring.rs
	- Restored original variable names
	- Use ecdsa_from_pkcs8 instead of load_with_algorithim for ECDSA keys
	- Return a Ok(KeyPair) out in each if block like the `generate`
	- Changed SigningKey implementation to be like original `KeyPair.rs`
- Drop unneeded test in custom_provider.rs

rcgen/src/certificate.rs
jgreeer: do you mean combining them into one parameter? the only thing with that is that it's different from the shape of the other functions

.github/workflows/ci.yml
jgreeer: I went ahead and removed this for now, I'm not sure if the way the flag works right now is how it's intended to?

rcgen/src/key_pair.rs
```diff
@@ -1,225 +1,137 @@
...
-		Self::try_from(private_key.contents())
+		let private_key: &[_] = private_key.contents();
+		Self::from_der(
+			&PrivateKeyDer::try_from(private_key).map_err(|_| Error::CouldNotParseKeyPair)?,
+			provider,
+		)
 	}
 
-	/// Obtains the key pair from a DER formatted key
+	/// Obtains the key pair from a PEM formatted key
```
jgreeer, line 108: this change isn't relevant, but i do think this comment is wrong? (copy pasted from from_pkcs8_der_and_sign_algo maybe). but could break this change off into another PR

.github/workflows/ci.yml
```diff
@@ -48,6 +48,9 @@ jobs:
             - run: cargo clippy --features ring,pem,x509-parser --all-targets
             # rustls-cert-gen require either aws_lc_rs or ring feature
             - run: cargo clippy -p rcgen --no-default-features --all-targets
+            - run: cargo clippy -p rcgen --no-default-features --features pem,x509-parser --all-targets
+            - name: Ensure backend-free builds have no built-in crypto dependencies
+              run: "! cargo tree -p rcgen --no-default-features --edges normal,build | grep -E 'ring v|aws-lc'"
```
djc: I don't like this `cargo tree` trickery much, and it doesn't feel like it's adding much value.

rcgen/src/crypto/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+#[cfg(feature = "ring")]
+pub mod ring;
+
+/// AWS-LC-based cryptography provider.
+#[cfg(any(feature = "aws_lc_rs", feature = "fips"))]
+pub mod aws_lc_rs;
+
+/// Cryptographic operations used by rcgen.
+pub trait CryptoProvider: std::fmt::Debug + Send + Sync {
+	/// Hash `input` with `algorithm`.
```
djc: Let's have the methods in this order:

- `verify()`
- `load_private_key()`
- `generate()`
- `hash()`

Both here and in both impls.

rcgen/src/crypto/mod.rs
djc: I think we want to the signing algorithm to be committed rather than being a per-signature decision. I think we should ditch `from_der_and_sign_algo()` and `from_pem_and_sign_algo()`.

In fact, with this PR, not sure the `KeyPair` type still has value? Seems to me that `Box<dyn SigningKey>` should basically replace it.

(Maybe that could be a separate PR?)

.github/workflows/ci.yml
jgreeer: yeah, the test right before this already does backend free anyways

rcgen/src/crypto/mod.rs
jgreeer: > I think we want to the signing algorithm to be committed rather than being a per-signature decision. I think we should ditch from_der_and_sign_algo() and from_pem_and_sign_algo().

I'm good with this, do we keep the `from_pkcs8` variants though? So we'd have `from_pem` , `from_der`, `from_pkcs8_pem`, and `from_pkcs8_der`

> In fact, with this PR, not sure the KeyPair type still has value? Seems to me that Box<dyn SigningKey> should basically replace it.

It's almost just a wrapper for CryptoProvider, but one use is that we have a struct to keep the private key DER. Maybe there's a better abstraction for it though.

Issue #448, opened 2026-08-20
CertificateSigningRequestParams::from_der can parse the wrong key type


GermanCoding:
**Version:** rcgen 0.14.9 (`default-features = false, features = ["aws_lc_rs", "x509-parser"]`;
## Summary

When parsing a CSR, rcgen keeps only the raw public key bits and derives the key's
`AlgorithmIdentifier` from the CSR's **signature** algorithm. For ECDSA that algorithm names a
curve, so a CSR whose signature hash does not match its key's curve comes back out as a *different
key*. Signing it produces a certificate whose `subjectPublicKeyInfo` claims one curve over a point
from another - resulting in an invalid public key.

## Reproducer

`Cargo.toml`:

```toml
[dependencies]
rcgen = { version = "0.14.9", default-features = false, features = ["aws_lc_rs", "x509-parser"] }
```

The CSR below is a plain OpenSSL request — a P-256 key, self-signed with `ecdsa-with-SHA384`. Such a CSR is not recommended, but also not forbidden as per RFC 5480 §4.

```sh
openssl ecparam -name prime256v1 -genkey -noout -out p256.key
openssl req -new -key p256.key -sha384 -subj /CN=example.com -outform DER -out p256.csr
```

```rust
use rcgen::{CertificateSigningRequestParams, PublicKeyData};

const CSR_HEX: &str = "\
3081fa3081a102010030163114301206035504030c0b6578616d706c652e636f6d3059301306\
072a8648ce3d020106082a8648ce3d03010703420004e96cb902a24cd0a748b038003facde9f\
d674918c5e10795106cdf03c3535d6d5542d5816ad97094e8bf459f1c3a647caef2f5efa4b36\
4c74e49770536c40008fa029302706092a864886f70d01090e311a301830160603551d11040f\
300d820b6578616d706c652e636f6d300a06082a8648ce3d0403030348003045022036d3561c\
ed3827f8fc4224d21d3bb2e36042316ba2fc6deb4cbbf9be130d1f82022100baccc313da1c38\
46a030f517bdae54b6ab2565bb7b7912b5d1bb4f6b4b1110ac";

/// `1.2.840.10045.3.1.7` (prime256v1), as encoded inside an AlgorithmIdentifier.
const PRIME256V1: &[u8] = &[0x06, 0x08, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07];
/// `1.3.132.0.34` (secp384r1).
[...]

PR #452, opened 2026-09-09
add RSA key size to SignatureAlgorithm

jgreeer:
## Summary

This PR adds support for having separate `SignatureAlgorithm` values for different key sizes. It adds a field `rsa_key_size` to the `SignatureAlgorithm` struct. We create a static SignatureAlgorithm struct for each combination of hash (SHA256, SHA384, SHA512) + key size (2048, 3072, 4096) adding 6 new algorithms. As a side effect, we have to pass `rsa_key_size: None` to the other algorithm structs. 

The reason for this change is in `key_pair.rs`, you can now use the `generate_for` method to generate RSA key pairs of different key sizes by giving different `SignatureAlgorithm` arguments. This removes the need for `generate_rsa_for`, so we deprecate it in the same way `RcgenError` was. Since `RsaKeySize` is always used now, we removed the `aws-lc-rs` and `crypto` gating.

The `rsa_key_size` field is not used it the identity of the key, because it would mess up loading the key, and we don't want to mess with the OID. At load time, like in `from_der_and_sign_algo`, the user can give an RSA SigAlg value of any key size since the key size is later retrieved using `rsa_key_pair_public_modulus_len`. 

## Tests

We add a test in `lib.rs` to verify that RSA algs of different key sizes are equal to each other.

### DIFF
--- a/rcgen/src/key_pair.rs
+++ b/rcgen/src/key_pair.rs
@@ -131,7 +131,15 @@ impl KeyPair {
 				})
 			},
 			#[cfg(feature = "aws_lc_rs")]
-			SignAlgo::Rsa(sign_alg) => Self::generate_rsa_inner(alg, sign_alg, KeySize::Rsa2048),
+			SignAlgo::Rsa(sign_alg) => Self::generate_rsa_inner(
+				alg,
+				sign_alg,
+				match alg.rsa_key_size {
+					Some(RsaKeySize::_3072) => KeySize::Rsa3072,
+					Some(RsaKeySize::_4096) => KeySize::Rsa4096,
+					_ => KeySize::Rsa2048,
+				},
+			),
 			// Ring doesn't have RSA key generation yet:
 			// https://github.com/briansmith/ring/issues/219
 			// https://github.com/briansmith/ring/pull/733
@@ -145,6 +153,9 @@ impl KeyPair {
 	/// If passed a signature algorithm that is not RSA, it will return
 	/// [`Error::KeyGenerationUnavailable`].
 	#[cfg(all(feature = "crypto", feature = "aws_lc_rs"))]
+	#[deprecated(
+		note = "pass a key-size-specific algorithm such as `PKCS_RSA_SHA256_4096` to `generate_for` instead"
+	)]
 	pub fn generate_rsa_for(
 		alg: &'static SignatureAlgorithm,
 		key_size: RsaKeySize,
@@ -616,7 +627,6 @@ impl From<KeyPair> for PrivateKeyDer<'static> {
 }
 
 /// The key size used for RSA key generation
-#[cfg(all(feature = "crypto", feature = "aws_lc_rs"))]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 #[non_exhaustive]
 pub enum RsaKeySize {
--- a/rcgen/src/sign_algo.rs
+++ b/rcgen/src/sign_algo.rs
@@ -8,6 +8,7 @@ use aws_lc_rs::signature::{
 use yasna::models::ObjectIdentifier;
 use yasna::DERWriter;
 
+use crate::key_pair::RsaKeySize;
 #[cfg(feature = "crypto")]
 use crate::ring_like::signature::{self, EcdsaSigningAlgorithm, EdDSAParameters, RsaEncoding};
 use crate::Error;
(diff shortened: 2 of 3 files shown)


rcgen/src/key_pair.rs
```diff
@@ -131,7 +131,14 @@ impl KeyPair {
...
 			},
 			#[cfg(feature = "aws_lc_rs")]
-			SignAlgo::Rsa(sign_alg) => Self::generate_rsa_inner(alg, sign_alg, KeySize::Rsa2048),
+			SignAlgo::Rsa(sign_alg) => {
+				let key_size = match alg.rsa_key_size {
+					Some(RsaKeySize::_3072) => KeySize::Rsa3072,
+					Some(RsaKeySize::_4096) => KeySize::Rsa4096,
+					_ => KeySize::Rsa2048,
+				};
+				Self::generate_rsa_inner(alg, sign_alg, key_size)
```
djc: Nit: suggest inlining the `match` in the function call.

rcgen/src/sign_algo.rs
```diff
@@ -38,6 +39,9 @@ pub struct SignatureAlgorithm {
 	pub(crate) sign_alg: SignAlgo,
 	oid_components: &'static [u64],
 	params: SignatureAlgorithmParams,
+	// Only read under `aws_lc_rs`. Not part of identity.
+	#[allow(dead_code)]
```
djc, line 43: How about we just make this conditional on `#[cfg(feature = "aws_lc_rs")]`? Seems straightforward?

djc (commented):
I think this makes sense with some tweaks?

rcgen/src/sign_algo.rs
jgreeer: we'd have to revert this back when we add the pluggable crypto provider right? not a big deal but it's extensible in it's current form already.

djc: Yeah, okay.

djc (approved):
Please squash the commits into a single one.

jgreeer:
@djc i think this is good to merge now, i've integrated the changes from https://github.com/rustls/rcgen/pull/453 and https://github.com/rustls/rcgen/commit/e6c421b40a7203897225dfc3df2b4d8a6924968c

cpu:
I'm a bit hesitant about this branch. I don't think we should continue to conflate signature identity with key generation policy, I think they should be modeled separately. Have you looked at what that might involve as a potential alternative? I was giving similar feedback in https://github.com/rustls/rcgen/issues/448#issuecomment-5457091013

jgreeer:
> I'm a bit hesitant about this branch. I don't think we should continue to conflate signature identity with key generation policy, I think they should be modeled separately. Have you looked at what that might involve as a potential alternative? I was giving similar feedback in [#448 (comment)](https://github.com/rustls/rcgen/issues/448#issuecomment-5457091013)

that's fair. this PR seems to go in the opposite direction of the feedback you gave. we're ultimately just trying to get rid of `generate_rsa_for` and instead have the functionality in `generate_for`. i think splitting `SignatureAlgorithm` into two is a good idea. maybe as an alternative we have another enum, something like `KeyPairAlgorithim`, that is only used as the argument to the generate function. i think this would achieve the clearest seperation of which enum is supposed to do what. or i guess it would probably have another name since we're getting rid of `KeyPair`.

djc:
> that's fair. this PR seems to go in the opposite direction of the feedback you gave. we're ultimately just trying to get rid of `generate_rsa_for` and instead have the functionality in `generate_for`. i think splitting `SignatureAlgorithm` into two is a good idea. maybe as an alternative we have another enum, something like `KeyPairAlgorithim`, that is only used as the argument to the generate function. i think this would achieve the clearest seperation of which enum is supposed to do what. or i guess it would probably have another name since we're getting rid of `KeyPair`.

Yup -- sorry for leading you astray. Do you have a chance to implement what @cpu was proposing in #448?

jgreeer:
> > that's fair. this PR seems to go in the opposite direction of the feedback you gave. we're ultimately just trying to get rid of `generate_rsa_for` and instead have the functionality in `generate_for`. i think splitting `SignatureAlgorithm` into two is a good idea. maybe as an alternative we have another enum, something like `KeyPairAlgorithim`, that is only used as the argument to the generate function. i think this would achieve the clearest seperation of which enum is supposed to do what. or i guess it would probably have another name since we're getting rid of `KeyPair`.
> 
> Yup -- sorry for leading you astray. Do you have a chance to implement what @cpu was proposing in #448?

sure!

PR #453, opened 2026-09-11
rcgen: stop storing key DER in KeyPair

djc:
First semver-incompatible PR on `main` for 0.15.0.

Preparation for

- #450

cc @jgreeer


rcgen/src/key_pair.rs
```diff
@@ -770,6 +737,20 @@ pub trait PublicKeyData {
 	fn algorithm(&self) -> &'static SignatureAlgorithm;
 }
 
+/// Serialize private key to PEM format
+#[cfg(feature = "pem")]
+pub fn serialize_private_key_pem(key: &PrivateKeyDer<'_>) -> Result<String, Error> {
+	let tag = match key {
+		PrivateKeyDer::Pkcs8(_) => "PRIVATE KEY",
+		PrivateKeyDer::Pkcs1(_) => "EC PRIVATE KEY",
+		PrivateKeyDer::Sec1(_) => "RSA PRIVATE KEY",
```
jgreeer, line 746:
```suggestion
		PrivateKeyDer::Pkcs1(_) => "RSA PRIVATE KEY",
		PrivateKeyDer::Sec1(_) => "EC PRIVATE KEY",
```
PKCS 1 is RSA and SEC1 is EC?

rcgen/src/key_pair.rs
jgreeer, line 77: not in the struct anymore

rcgen/src/key_pair.rs
djc: Oops, fixed!

rcgen/src/key_pair.rs
djc: Fixed this.

cpu:
> Preparation for https://github.com/rustls/rcgen/pull/450

Can you say more about the motivation? The diff doesn't seem completely unreasonable, but it's also making a lot of places throw away a return value from the key generation. It feels mildly like the older approach was a little bit more ergonomic and I'm not sure I'm 100% on what we gain by trading some of that away to not have the KeyPair hold the DER.

djc:
> > Preparation for #450
> 
> Can you say more about the motivation? The diff doesn't seem completely unreasonable, but it's also making a lot of places throw away a return value from the key generation. It feels mildly like the older approach was a little bit more ergonomic and I'm not sure I'm 100% on what we gain by trading some of that away to not have the KeyPair hold the DER.

In my mind the next step here is getting rid of `KeyPair` completely in favor of only having `Box<dyn SigningKey>`, which seems like a mostly unnecessary layer at this point (which would need more complexity with crypto providers). In terms of defense in depth, since rcgen does not need the type capable of signing to also contain a more readable/leakable copy of the private key contents we should probably not retain/hold that by default.

rcgen/src/lib.rs
```diff
@@ -120,15 +122,15 @@ let CertifiedKey { cert, signing_key } = generate_simple_self_signed(subject_alt
 
 // The certificate is now valid for localhost and the domain "hello.world.example"
 println!("{}", cert.pem());
-println!("{}", signing_key.serialize_pem());
+println!("{}", signing_key.public_key_pem());
```
cpu: I think this docs change is papering over the fact that `generate_simple_self_signed()` isn't a useful helper anymore.
The user here won't be interested in the public key PEM, they'll need the private key PEM to make `cert.pem()` operational.

I don't think you can get to that anymore now that internally the helper is swallowing the private key DER from `KeyPair::generate()`.

I think probably `generate_simple_self_signed()` needs to change to return `Result<(CertifiedKey<KeyPair>, PrivateKeyDer<'static>), ...>`

rcgen/src/key_pair.rs
cpu: Did this part not get pushed? In the latest commit I'm still seeing Pkcs1 -> EC and Sec1 -> RSA:
https://github.com/rustls/rcgen/pull/453/changes/20077dc2a054a6e817f0b670e38c39e268aaf8d9#diff-186f93a466f8a7ddc0352913f9c19d66c677d47b83432f6f6756641e59789bcfR745-R746

rcgen/src/key_pair.rs
cpu: I think this change wasn't pushed as well. It still has the field.

rustls-cert-gen/Cargo.toml
```diff
@@ -23,7 +23,7 @@ aws-lc-rs = { workspace = true, optional = true }
 bpaf = { workspace = true }
 pem = { workspace = true }
 pki-types = { workspace = true }
-rcgen = { version = "0.14.2", path = "../rcgen", default-features = false, features = ["pem"] }
```
cpu, line 26: I think we can drop the `zeroize` dep from ../Cargo.toml now too.

rcgen/src/key_pair.rs
djc: Huh, weird. Cleaned it up.

rcgen/src/key_pair.rs
djc: Fixed it hopefully for real this time?

rcgen/src/lib.rs
djc: Ah, good call -- changed it as suggested.

README.md
```diff
@@ -14,7 +14,7 @@ let subject_alt_names = vec!["hello.world.example".to_string(),
 
 let CertifiedKey { cert, signing_key } = generate_simple_self_signed(subject_alt_names).unwrap();
 println!("{}", cert.pem());
-println!("{}", signing_key.serialize_pem());
```
cpu, line 17: I think this README example needs the same catch-up for the `generate_simple_self_signed()` adjustment.

rustls-cert-gen/Cargo.toml
djc: Nice, done.

cpu (approved):

cpu (approved):

PR #454, opened 2026-09-18
Continued extension representation rework

### LINKED ISSUE #122: X509v3 extensions from certificate not transferred to CSR

### LINKED ISSUE #150: Allow CSR parsing to handle custom extensions

### LINKED ISSUE #155: Enforce extension uniqueness

### LINKED ISSUE #446: Rework X509 extension code

cpu:
This revives and finishes the extension rework from #446 and continues where #449 left off by starting on breaking changes. Each extension is now its own type in a new ext module, with the presence decision living next to the encoding.

Serialization builds an `Extensions` collection first and only writes what was actually built, so the bug class that produced silently dropped extensions is addressed more structurally.

Throughout duplicate OIDs are rejected instead of silently emitted, on both the write and parse sides (#155).

CSR parsing also got friendlier. Requested extensions rcgen doesn't handle natively used to fail the whole parse with `UnsupportedExtension`. Now they're preserved as custom extensions, so a CSR round-trips through `from_der` without losing anything (#150). Same deal for extended key usages with purpose OIDs rcgen doesn't know, those now parse into  `ExtendedKeyUsagePurpose::Other` instead of erroring.

Along the way this grew some related changes:
* extensions borrow from params instead of cloning
* the path length enum was renamed `PathLenConstraint` to free up the `BasicConstraints` name for the extension type and because it's more precise.
* SKI is now emitted for end entity certs too (I think this is the right default), and the `CustomExtension`/ACME identifier API got a rework.

For most users this will all be fairly uneventful and require no code changes since the parameters+issuance APIs remain mostly the same, but some changes are semver-breaking so this bumps to 0.15.0 so we can use `main` to prepare the next major release (https://github.com/rustls/rcgen/issues/418).

Resolves #150
Resolves #155
Resolves #446
Resolves #122


cpu:
> cpu marked this pull request as ready for review now

Rebased on `main` and carried forward the tweaks from https://github.com/rustls/rcgen/pull/456 throughout.

rcgen/src/certificate.rs
```diff
@@ -496,11 +492,11 @@ pub struct Attribute {
 /// [RFC 5280](https://tools.ietf.org/html/rfc5280#section-4.2)
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub struct CustomExtension {
-	oid: Vec<u64>,
-	critical: bool,
+	pub(crate) oid: Vec<u64>,
```
djc, line 495: Suggest inserting a commit before this one that moves `CustomExtension` into `extension`? Definitely want to keep the type together with its `Extension` implementation, too.

(I did find this commit later on, but would be better to do it here IMO.)

rcgen/src/crl.rs
```diff
@@ -259,50 +250,6 @@ impl CertificateRevocationListParams {
 	}
 }
 
-/// A certificate revocation list (CRL) issuing distribution point, to be included in a CRL's
```
djc, line 262: Hmm, on `main` the `crl` module is 480 lines and `extension` is already 1276 lines. Might be better to keep CRL-only extensions in `crl`?

(Also better to keep moves and refactoring separate in terms of commits.)

rcgen/src/extension.rs
```diff
@@ -1299,6 +1297,60 @@ impl Extension for &CustomExtension {
...
+pub(crate) struct Extensions<'params> {
+	exts: Vec<Box<dyn Extension + 'params>>,
+}
+
+impl<'params> Extensions<'params> {
+	/// Add an extension to the collection.
+	///
+	/// Returns [`Error::DuplicateExtension`] if the extension's OID is already present
+	/// in the collection.
+	pub(crate) fn add_extension(
```
djc, line 1315: Nit: suggest just calling this `push()` to mimic `Vec::push()`?

rcgen/src/extension.rs
```diff
@@ -1299,6 +1297,60 @@ impl Extension for &CustomExtension {
...
+		Ok(())
+	}
+
+	/// Write the certificate's optional extensions field.
+	///
+	/// Nothing is written when the collection is empty: presence is decided by the
+	/// built collection, not predicted from the params, so an empty extensions
+	/// field is never emitted and requested extensions can never be silently
+	/// dropped.
+	pub(crate) fn write_exts_der(&self, writer: DERWriter) {
```
djc, line 1338: Nit: the `exts_` infix feels superfluous here?

djc (commented):
Suggest we split this once more, before the `AcmeIdentifier` inclusion?

Two concerns remaining (somewhat from the previous PR):

- It feels like in a number of cases, this is wrapping a local type in another local type, when we could arguably implement `Extension` for the existing type rather than wrapping it in another layer.
- The `from_params(params: ..) -> Option<Self>` pattern feels somewhat overwrought? Especially to the extent we can reuse existing types, can we just have `write()` yield `bool` instead of having this indirection between something that exists, then is written?

djc:
> * The `from_params(params: ..) -> Option<Self>` pattern feels somewhat overwrought? Especially to the extent we can reuse existing types, can we just have `write()` yield `bool` instead of having this indirection between something that exists, then is written?

Oh, I guess this is necessary because we only want to write the extensions wrapper when we know there are 1+ extensions to write? Still wondering if there isn't a more light-weight pattern at least for some of the extensions.

cpu:
> Still wondering if there isn't a more light-weight pattern at least for some of the extensions.

I will noodle on this when I split it up & address your other feedback. This part of the solution is more or less as-is from 2023 (time flies!) and I wouldn't be surprised if in 2026 I have a better idea :laughing:

PR #456, opened 2026-09-21
Tweak new extensions module

djc:
Some followup from

- #449 

For some of these, it feels like the `from_params()` pattern is a little overwrought? Like, if there's only one params field, can't we do something simpler? Or does some of the rework in

- #454 

depend on the uniformity here?


rcgen/src/crl.rs
```diff
@@ -4,7 +4,7 @@ use pki_types::CertificateRevocationListDer;
 use time::OffsetDateTime;
 use yasna::{DERWriter, Tag};
 
-use crate::ext::{
+use crate::extensions::{
```
ctz: This is the narrowest of tiny nits, but I think `extension` might be nicer? So someone can write `extension::NameConstraints`?

ctz (approved):

rcgen/src/crl.rs
djc: The module is private though? It's more of an internal dumping ground for all our extension-related types, IMO.

Let's see what @cpu thinks.

cpu: I have no strong attachment either way, but it does seem like the other modules are singular and so `extension` would fit in better 🤷

cpu (approved):
Thanks! 

> For some of these, it feels like the from_params() pattern is a little overwrought? Like, if there's only one params field, can't we do something simpler? Or does some of the rework in 454 depend on the uniformity here?

I don't think 454 necessarily depends on the uniformity, I just thought it was a nice pattern for keeping things consistent & easy to review. I'm open to changing on a case-by-case basis if you prefer an alternative.

rcgen/src/crl.rs
djc: Renamed it to `extension`.

